Issue #41, opened 2020-08-18
Installed PWA doesn't seem to load if server is unavailable


bwindels:
Request seems to make it to service worker, which does a failing request, but we don't seem to fallback to cache?

PR #120, opened 2020-10-01
Improve caching and fix working offline

### LINKED ISSUE #94: Can't load session while offline
it just says `Something went wrong: undefined.`, logging an error on the olm script element, so this is olm related files missing from the service worker cache. We should also cache the fonts and images, etc... but perhaps in a way where we keep the cached version if it hasn't changed. E.g. we should only cache `index.html` in a versioned cache, and cache the others by their filename (including content hash). Once the service worker is activated, we see which files are in cache that are not in the FILES array anymore generated by the build and remove then from the cache.

bwindels:
- [x] cache all files (e.g. olm) needed to work offline (fixes #94, make sure it also fixes https://github.com/vector-im/hydrogen-web/issues/41)
 - [x] don't pre-cache everything, some resources are mutually exclusive (and heavy) like woff vs woff2 files. Instead, cache them when requested in the service worker by their filename only (not app version, so we don't have to re-download them with every update). Also remove previous files by iterating over all cache keys and see if there is one the same but with a different hash. 
 - [x] cache unhashed files (index.html) by git commit hash in the service worker, rather than version
 - [x] cache thumbnails below a given size (px, kb) from the homeserver media repository

### DIFF
--- a/index.html
+++ b/index.html
@@ -15,6 +15,7 @@
 	<body class="hydrogen">
         <script id="version" type="disabled">
             window.HYDROGEN_VERSION = "%%VERSION%%";
+            window.HYDROGEN_GLOBAL_HASH = "%%GLOBAL_HASH%%";
         </script>
 		<script id="main" type="module">
 			import {main} from "./src/main.js";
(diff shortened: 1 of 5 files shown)


bwindels:
Works in chrome, but not yet in firefox (seems to fetch all files on a new version still). Also test in edge, safari ...

bwindels:
FF got stuck on an old version of the sw for 0.1.1 somehow :grimacing: not sure how that happened. After unregistering it manually, it updated and all worked fine :+1:

PR #148, opened 2020-10-09
Add Dockerfile

ynerant:
While the installation is already super easy, using a Dockerfile and pushing Hydrogen to Docker hub could be even better :whale:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM node:alpine3.12
+RUN apk add --no-cache git
+COPY . /code
+WORKDIR /code
+RUN yarn install
+EXPOSE 3000
+ENTRYPOINT ["yarn", "start"]


bwindels:
Thanks for this. I'm fine with adding this file (we'll need DCO sign-off), but if I understand correctly, this would add a step to the release process of having to push to docker hub every time? I don't think I would want to do that at this point.

ynerant:
Pushing to docker hub is not needed at this step of the project (but will be a good feature later when Hydrogen will be more stable ;-))
A Dockerfile only provides the steps to build a Docker image from the sources. This is useful if we want to have independant containers, without installing the application in our machine.

To build the image, simply run `docker build -t hydrogen .` (if . is the directory where the project is cloned), or better `docker build -t hydrogen https://github.com/vector-im/hydrogen-web.git`. Then, `docker run -p 3000:3000 hydrogen` launches the container and exposes the port 3000. We never installed yarn in our computer.

bwindels:
Cool, can you just [sign off](https://github.com/matrix-org/matrix-js-sdk/blob/master/CONTRIBUTING.rst#sign-off) your commit?

ynerant:
Better?

PR #171, opened 2020-10-20
Dark Mode

mcnesium:
This adds a dark mode to Hydrogen. It overrides all the light colors set in `theme.css` with darker ones on `prefers-color-scheme: dark`, so it should only be seen as a hacky first step.

Screenshot from 2020-10-20 17-30-24

This also turns scrollbars of the room list and the timeline dark, but [only in Firefox](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars).

### DIFF
--- a/src/ui/web/css/themes/element/theme.css
+++ b/src/ui/web/css/themes/element/theme.css
@@ -635,15 +635,19 @@ button.link {
 
 @media (prefers-color-scheme: dark) {
     .hydrogen {
-        background-color: #000;
+        background-color: #15191e;
         color: #fff;
     }
-    .AnnouncementView > div,
+    .AnnouncementView > div {
+        background-color: #20262b;
+    }
     .middle-header,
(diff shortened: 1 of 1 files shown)


bwindels:
Hi, thank you both for your PR! As I mentioned in the room though, I don't think I want to commit to maintaining a dark mode at this point, as I think this CSS is very likely to break in the future. The main problem is that it will automatically activate for people with dark mode on, and if half of the selectors are broken due to future changes, they might get a mixed result that is worse than no dark mode at all.

Perhaps this is better suited for https://userstyles.org/ for the time being?

mcnesium:
I understand. Just did not want to keep my fiddle to myself. Maybe someone who wants to selfhost this wants to merge it. Feel free to close the PR.

bwindels:
Thanks for understanding :(

PR #182, opened 2020-11-02
Use submit event for login instead of click

Newbytee:
This allows the user to press enter to log in instead of having to click the log in button. It could potentially also allow for other ways of confirming your login if a browser implements others, which may be good for accessibility. Currently a failed login results in the sign in UI going into a frozen state permanently however which is why this is a draft. Rest assured I will continue working on this.

### DIFF
--- a/src/platform/web/ui/general/html.js
+++ b/src/platform/web/ui/general/html.js
@@ -94,7 +94,7 @@ export const TAG_NAMES = {
     [HTML_NS]: [
         "br", "a", "ol", "ul", "li", "div", "h1", "h2", "h3", "h4", "h5", "h6",
         "p", "strong", "em", "span", "img", "section", "main", "article", "aside",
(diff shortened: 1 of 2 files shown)


Newbytee:
Seems like the behaviour where the UI "freezes" happens regardless of whether this patch is applied or not, so I'll mark this as ready for review.

bwindels:
This looks great, thank you! I'll have a deeper look later on before I merge it, but I don't see any blockers. The freezing behaviour you're seeing sounds like #152

bwindels:
Ah, could you also sign off your PR or commits @Newbytee? See https://github.com/matrix-org/matrix-js-sdk/blob/master/CONTRIBUTING.rst#sign-off

Newbytee:
> Ah, could you also sign off your PR or commits @Newbytee? See https://github.com/matrix-org/matrix-js-sdk/blob/master/CONTRIBUTING.rst#sign-off

Done

PR #191, opened 2020-11-06
Share megolm session once you start typing

bwindels:


bwindels:
make ensure... reentrant with encrypt fn

bwindels:
check how slow it is on IE11 when typing

bwindels:
see why query is being called again when sending

Issue #247, opened 2021-02-25
Decrypt device messages before attempting megolm decryption during sync


bwindels:
Currently, we write encrypted device messages to the session store as "pending decryption" during sync. Once the sync txn is written, we attempt to decrypt them in the `afterSyncCompleted` phase. If the device message decrypts to a room key, it is stored and passed to `room.notifyRoomKey` which attempts to decrypt events encrypted with that key, which were likely in the current sync response. Because of #248 , this retrying of decryption can actually be quite slow, which needs to be fixed as part of that bug. However, here we can avoid that path and store new room keys before we try megolm decryption as part of sync prepare stage, which makes sense as room keys and their messages will often be in the same sync response. No need to attempt to decrypt them twice if that is the case.

Thus, we need to:
 1. move olm decryption to prepare step
 2. pass the room keys to megolm somehow. Either by storing them already but then the prepare txn needs to become readwrite, or passing them in as "extra" keys to megolm.
 3. pass the olm decryption changes from prepare to write stage.
 4. write the olm decryption changes during `room.writeSync`.
 5. remove the decryptPending code in `DeviceMessageHandler`.

bwindels:
so how do we know whether to call notifyRoomKey because we have earlier events? I guess if no (message?) event decrypts with it?

bwindels:
We need to at least look for events encrypted with the session id in the open timeline, if any.

bwindels:
> We need to at least look for events encrypted with the session id in the open timeline, if any.

done

Issue #248, opened 2021-02-25
Reattempting decryption on new room key can be slow


bwindels:
In logging, it looks like `room.notifyRoomKey` called from `DeviceMessageHandler` can be quite slow. There is only indirect evidence (the "session" log saying up to 35s with the decryption only taking 1s as we log the keys), after we log the keys though, the only other thing we do is commit the txn, which should not take too long, so the time is very likely spent in `room.notifyRoomKey`. The average time spent per key ranges from 5ms to 500ms on a slow mobile device. It does add up for large catchup syncs of at least a few days old though.

One explanation could be that the `lastDecryptedEventKey` stopped updating, likely after a limited sync. On my desktop instance, I had a look in the store and this was the case for 54/88 encrypted rooms at least didn't have the same event index, with many having a different fragment id than the last live key in the sync writer (see `diff-summary-last-decrypted-with-last-live-key.json`), so does not seem uncommon.

Given that the `lastDecryptedEventKey` depends on comparing fragment ids, which still seems to have some issues with it (see #43), and the whole thing seems kind of fragile, maybe it's better to abandon this technique and go for something better to know what events to decrypt when a room key arrives.

Maybe it worth reiterating first why we need to retry decryption when a room key arrives, even if the timeline is not open. We only want to set the `isUnread` flag and `lastMessageTimestamp` on the room summary when we receive a message. This drives the sorting and appearance in the room list. We can only know if an event was a message after decryption, so that's why we need to retry decryption with the new key independent of whether we are looking at the timeline. `lastDecryptedEventKey` was added to limit the scan of events that could be encrypted with the room key.

The new proposed solution to this problem is the following:
 1. When encountering an encrypted event **during sync** (not while backfilling as there missing sessions is expected and this is not important for the room summary) for which we do not have the room key, store an empty session in the `inboundGroupSessions` store if none exists yet, and add the event id to it.
 2. When the room key arrives, we'll notice there is a record for it in `inboundGroupSessions` with all the event ids we know about to retry decryption with
 3. When calling `room.clearUnread` remove all the missing sessions with event ids in `inboundGroupSessions`, as we don't care about it anymore the set the `isUnread` flag. Not setting `lastMessageTimestamp` when the key arrives is probably also fine as we've looked at the room already.
[...]

PR #250, opened 2021-02-26
Provide new room keys in prepare decryption phase of sync

bwindels:
Implements #247 

Should be there but needs testing

### DIFF
--- a/src/matrix/DeviceMessageHandler.js
+++ b/src/matrix/DeviceMessageHandler.js
@@ -14,11 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {OLM_ALGORITHM, MEGOLM_ALGORITHM} from "./e2ee/common.js";
-import {countBy} from "../utils/groupBy.js";
-
-// key to store in session store
-const PENDING_ENCRYPTED_EVENTS = "pendingEncryptedDeviceEvents";
+import {OLM_ALGORITHM} from "./e2ee/common.js";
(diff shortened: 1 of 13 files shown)


src/matrix/DeviceMessageHandler.js
```diff
@@ -32,90 +29,52 @@ export class DeviceMessageHandler {
...
-        let roomKeys;
-        log.set("eventsCount", countBy(olmResults, r => r.event.type));
-        log.set("roomKeys", megOlmRoomKeysResults.length);
-        if (megOlmRoomKeysResults.length) {
-            roomKeys = await this._megolmDecryption.addRoomKeys(megOlmRoomKeysResults, txn, log);
-        }
-        log.set("newRoomKeys", roomKeys.length);
-        return {roomKeys};
+    obtainSyncLock(toDeviceEvents) {
+        return this._olmDecryption?.obtainDecryptionLock(toDeviceEvents);
```
bwindels, line 33: Sync awaits this, but awaiting undefined is fine I guess.

bwindels: I think it's fine because in `prepareSync` we first check if we have olm decryption before using the lock.

bwindels:
Needs a final review but should be good to go. Should wait to merge until #248 is ready though otherwise the room list won't update anymore for e2ee rooms.

bwindels:
Tested together with work on PR #252

PR #252, opened 2021-03-02
Track event ids of missing megolm sessions

### LINKED ISSUE #248: Reattempting decryption on new room key can be slow

bwindels:
Fixes #248, based of PR #250


bwindels:
This PR does not address the following limitations that were already present:
 - The room will be marked as unread when the reattempt to decrypt succeeds, independent of when the event was sent. This will be solved when implementing read receipts properly, at least for the syncing user.
 - Session backup will be requested when loading undecryptable events in the timeline, this has been split off into #254

PR #261, opened 2021-03-05
Don't consider lazy-load member events for room membership changes

### LINKED ISSUE #192: Room keys are re-shared for every LL member event from server
member events in the state section of the sync room response currently trigger a room key share if their membership is join because [a missing prev_content.membership is considered a leave](https://github.com/vector-im/hydrogen-web/blob/master/src/matrix/room/members/RoomMember.js#L143). This was unintentional, but seems [spec-compliant](https://matrix.org/docs/spec/client_server/latest#m-room-member), so not sure what the best solution is.

In practice, synapse seems to send member events in the state section with a `prev_content` if they represent a membership change in a sync gap. If the event is solely included for LL purposes, it seems to lack a `prev_content`. So I guess we should only use leave as default prev membership if the event was synced in the timeline, and for the state section, don't consider events that don't have prev_content as a membership transition.

bwindels:
Fixes #192

### DIFF
--- a/src/matrix/room/timeline/persistence/MemberWriter.js
+++ b/src/matrix/room/timeline/persistence/MemberWriter.js
@@ -56,10 +56,20 @@ export class MemberWriter {
         if (!existingMember || !existingMember.equals(member)) {
             txn.roomMembers.set(member.serialize());
             this._cache.set(member);
-
-            if (!isLazyLoadingMember) {
-                return new MemberChange(member, existingMember?.membership);
+            // we also return a member change for lazy loading members if something changed,
+            // so when the dupe timeline event comes and it doesn't see a diff
+            // with the cache, we already returned the event here.
+            // 
+            // it's just important that we don't consider the first LL event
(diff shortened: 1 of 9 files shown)


src/matrix/room/timeline/persistence/MemberWriter.js
```diff
@@ -0,0 +1,203 @@
...
+        let existingMember = this._cache.get(userId);
+        if (!existingMember) {
+            const memberData = await txn.roomMembers.get(this._roomId, userId);
+            if (memberData) {
+                existingMember = new RoomMember(memberData);
+            }
+        }
+
+        // either never heard of the member, or something changed
+        if (!existingMember || !existingMember.equals(member)) {
```
bwindels, line 56: if there is a display name change, and both state and timeline contain this event (which synapse does), and the room response is not limited, this will fail to emit a member change when the timeline event is processed because the state event already updated the cache/storage so the equals check passes ...

PR #279, opened 2021-03-16
Manually adapt UI when keyboard shows or hides on mobile Safari

### LINKED ISSUE #181: Viewport does not resize on iOS when showing keyboard
So the room header becomes hidden

Johennes:
Mobile Safari seems to be the only browser that does *not* resize the viewport when the keyboard shows and hides. Instead the window is moved to make room for the keyboard which moves content at the top off screen.

This uses the VisualViewport API to manually resize the `SessionView` in response to keyboard display events. Additionally, the scroll position of the `Timeline` view is retained (if the timeline is currently shown).

Below is a short demo video. This isn't quite perfect but at least it's more usable than before.

https://user-images.githubusercontent.com/1137962/111364120-21c6f900-8691-11eb-8801-f1625f0c3675.MP4

Note that the VisualViewport API was only introduced with iOS 13. According to [statista.com], versions below 13 made up for 19% of all iOS users in summer 2020, with the share continuing to fall off. As a result, this seems like an acceptable workaround.

Things that I wasn't sure about:

1. Is `main.js` the best place for this logic?
2. Should we utilize the user agent to limit this to Safari only? I wasn't able to get Hydrogen to load in mobile Chrome and Firefox so I couldn't test this on other handheld browsers.

Fixes: #181

[statista.com]: https://www.statista.com/statistics/565270/apple-devices-ios-version-share-worldwide/

### DIFF
--- a/src/platform/web/Platform.js
+++ b/src/platform/web/Platform.js
@@ -35,6 +35,7 @@ import {WorkerPool} from "./dom/WorkerPool.js";
 import {BlobHandle} from "./dom/BlobHandle.js";
 import {hasReadPixelPermission, ImageHandle, VideoHandle} from "./dom/ImageHandle.js";
 import {downloadInIframe} from "./dom/download.js";
+import {Disposables} from "../../utils/Disposables.js";
 
 function addScript(src) {
     return new Promise(function (resolve, reject) {
@@ -83,6 +84,44 @@ async function loadOlmWorker(config) {
     return olmWorker;
 }
 
+// needed for mobile Safari which shifts the layout viewport up without resizing it
+// when the keyboard shows (see https://bugs.webkit.org/show_bug.cgi?id=141832)
+function adaptUIOnVisualViewportResize(container) {
+    if (!window.visualViewport) {
+        return;
+    }
+    const handler = () => {
+        const sessionView = container.querySelector('.SessionView');
+        if (!sessionView) {
+            return;
+        }
+
+        const scrollable = container.querySelector('.bottom-aligned-scroll');
+        let scrollTopBefore, heightBefore, heightAfter;
+
+        if (scrollable) {
+            scrollTopBefore = scrollable.scrollTop;
+            heightBefore = scrollable.offsetHeight;
+        }
+
+        // Ideally we'd use window.visualViewport.offsetTop but that seems to occasionally lag
+        // behind (last tested on iOS 14.4 simulator) so we have to compute the offset manually
+        const offsetTop = sessionView.offsetTop + sessionView.offsetHeight - window.visualViewport.height;
+
+        container.style.setProperty('--ios-viewport-height', window.visualViewport.height.toString() + 'px');
+        container.style.setProperty('--ios-viewport-top', offsetTop.toString() + 'px');
+
+        if (scrollable) {
+            heightAfter = scrollable.offsetHeight;
+            scrollable.scrollTop = scrollTopBefore + heightBefore - heightAfter;
+        }
+    };
+    window.visualViewport.addEventListener('resize', handler);
+    return () => {
+        window.visualViewport.removeEventListener('resize', handler);
+    };
+}
+
 export class Platform {
     constructor(container, config, cryptoExtras = null, options = null) {
         this._config = config;
@@ -115,6 +154,10 @@ export class Platform {
         }
         const isIE11 = !!window.MSInputMethodContext && !!document.documentMode;
         this.isIE11 = isIE11;
+        // From https://stackoverflow.com/questions/9038625/detect-if-device-is-ios/9039885
+        const isIOS = /iPad|iPhone|iPod/.test(navigator.platform) || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1) && !window.MSStream;
+        this.isIOS = isIOS;
+        this._disposables = new Disposables();
     }
 
     get updateService() {
@@ -135,6 +178,13 @@ export class Platform {
         if (this.isIE11) {
             this._container.className += " legacy";
         }
+        if (this.isIOS) {
+            this._container.className += " ios";
+            const disposable = adaptUIOnVisualViewportResize(this._container);
+            if (disposable) {
+                this._disposables.track(disposable);
+            }
+        }
         window.__hydrogenViewModel = vm;
         const view = new RootView(vm);
         this._container.appendChild(view.mount());
@@ -152,7 +202,7 @@ export class Platform {
         if (navigator.msSaveBlob) {
             navigator.msSaveBlob(blobHandle.nativeBlob, filename);
         } else {
-            downloadInIframe(this._container, this._config.downloadSandbox, blobHandle, filename);
+            downloadInIframe(this._container, this._config.downloadSandbox, blobHandle, filename, this.isIOS);
         }
     }
 
@@ -200,5 +250,9 @@ export class Platform {
 
     get version() {
         return window.HYDROGEN_VERSION;
+    }
+
+    dispose() {
+        this._disposables.dispose();
     }
 }
--- a/src/platform/web/ui/css/layout.css
+++ b/src/platform/web/ui/css/layout.css
@@ -54,6 +54,13 @@ main {
     min-width: 0;
 }
 
+/* resize and reposition session view to account for mobile Safari which shifts
+the layout viewport up without resizing it when the keyboard shows */
+.hydrogen.ios .SessionView {
+    height: var(--ios-viewport-height, 100%);
+    top: var(--ios-viewport-top, 0);
+}
+
 /* hide back button in middle section by default */
 .middle .close-middle { display: none; }
 /* mobile layout */
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -351,7 +351,11 @@ async function buildCssLegacy(entryPath, urlMapper = null) {
     const preCss = await fs.readFile(entryPath, "utf8");
     const options = [
         postcssImport,
-        cssvariables(),
+        cssvariables({
+            preserve: (declaration) => {
+                return declaration.value.indexOf("var(--ios-") == 0;
+            }
+        }),
         autoprefixer({overrideBrowserslist: ["IE 11"], grid: "no-autoplace"}),
         flexbugsFixes()
     ];
--- a/src/platform/web/dom/download.js
+++ b/src/platform/web/dom/download.js
@@ -14,10 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// From https://stackoverflow.com/questions/9038625/detect-if-device-is-ios/9039885
-const isIOS = /iPad|iPhone|iPod/.test(navigator.platform) || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1) && !window.MSStream;
-
-export async function downloadInIframe(container, iframeSrc, blobHandle, filename) {
+export async function downloadInIframe(container, iframeSrc, blobHandle, filename, isIOS) {
     let iframe = container.querySelector("iframe.downloadSandbox");
     if (!iframe) {
         iframe = document.createElement("iframe");
--- a/src/platform/web/ui/session/room/TimelineList.js
+++ b/src/platform/web/ui/session/room/TimelineList.js
@@ -40,7 +40,7 @@ function viewClassForEntry(entry) {
 export class TimelineList extends ListView {
     constructor(viewModel) {
         const options = {
-            className: "Timeline",
+            className: "Timeline bottom-aligned-scroll",
             list: viewModel.tiles,
         }
         super(options, entry => {


src/main.js
```diff
@@ -52,6 +52,33 @@ export async function main(platform) {
         });
         await vm.load();
         platform.createAndMountRootView(vm);
+
+        // needed for mobile Safari which does *not* resize the viewport when the keyboard
+        // shows or hides (see https://bugs.webkit.org/show_bug.cgi?id=141832)
+        if (window.visualViewport) {
```
bwindels: This file shouldn't contain any DOM related code, so I'd move this to `platform.createAndMountRootView`. In there you can see we already add a "legacy" class to the root DOM element if IE11 is detected, I'd do the same for iOS (there's already a check for iOS in `platform/web/DOM/download.js` that you can move to `Platform.js` perhaps and pass that into `downloadInFrame`.

I'd also prefer to only run this code for iOS, so in `createAndMountRootView` you check the iOS flag and conditionally set up this workaround.

src/main.js
```diff
@@ -52,6 +52,33 @@ export async function main(platform) {
...
+                const timeline = document.getElementsByClassName('Timeline')[0];
+                let scrollTopBefore, heightBefore, heightAfter;
+
+                if (timeline) {
+                    scrollTopBefore = timeline.scrollTop;
+                    heightBefore = timeline.getBoundingClientRect().height;
+                }
+
+                sessionView.style.height = window.visualViewport.height.toString() + 'px';
+                sessionView.style.top = window.visualViewport.offsetTop.toString() + 'px';
```
bwindels: Won't this be `0`? Also, `top` won't have any effect AIUI as session view is statically positioned?

src/main.js
```diff
@@ -52,6 +52,33 @@ export async function main(platform) {
...
+
+                const timeline = document.getElementsByClassName('Timeline')[0];
+                let scrollTopBefore, heightBefore, heightAfter;
+
+                if (timeline) {
+                    scrollTopBefore = timeline.scrollTop;
+                    heightBefore = timeline.getBoundingClientRect().height;
+                }
+
+                sessionView.style.height = window.visualViewport.height.toString() + 'px';
```
bwindels: I'd like to make this workaround obvious in `layout.css` where the [height is set up](https://github.com/vector-im/hydrogen-web/blob/master/src/platform/web/ui/css/layout.css#L46), rather than being an invisible monkey patch in the background. One possibility is to use a css custom property (`--ios-viewport-height`) (support for custom properties overlaps with `visualViewport` support on iOS) to pass the viewport height from JS to CSS, to keep the specifics of which element is resized in CSS. In CSS, you can only apply this custom variable to the height if the root element has the `ios` class set, like we already do for `.legacy` in `spinner.css`. E.g. something like:

```css
.hydrogen.ios .SessionView {
  height: var(--ios-viewport-height, 100%);
}

File: src/main.js [hunk 4]
```diff
@@ -52,6 +52,33 @@ export async function main(platform) {
...
+                sessionView.style.top = window.visualViewport.offsetTop.toString() + 'px';
+
+                if (timeline) {
+                    heightAfter = timeline.getBoundingClientRect().height;
+                    timeline.scrollTop = scrollTopBefore + heightBefore - heightAfter;
```
bwindels: Hmm, wonder if we could easily make this not specific to the timeline ... like do it for all scrollable elements or something :thinking:

bwindels (changes requested):
Thank you for your research and this PR! `visualViewport` seems like the way to go indeed on iOS. There's a bit of refactoring needed before we can merge this, and I've left a few comments with pointers how to make this fit with how things are organized in hydrogen. If you're up for addressing these, great! If not, no worries and I can do the necessary when I get to it.

src/main.js
bwindels: Ah, it's `fixed` indeed :) Still, won't `top` be `0`?

Johennes:
> I've left a few comments with pointers how to make this fit with how things are organized in hydrogen. If you're up for addressing these, great! If not, no worries and I can do the necessary when I get to it.

Great feedback! These look like they might be manageable, even for someone like me. :sweat_smile: 

I'll give it a go and get back to you later this week.

src/platform/web/Platform.js
```diff
@@ -130,9 +163,12 @@ export class Platform {
     }
 
     createAndMountRootView(vm) {
-        if (this.isIE11) {
+        if (this.isIE11 || this.isIOS) {
```
bwindels: I think we probably need a separate `ios` class for iOS rather than reusing `legacy`.

src/main.js
Johennes: Yeah, it's a little confusing with these two viewports. `offsetTop` is the distance from the top edge of the layout viewport to the top edge of the visual viewport. So when the keyboard shows, `offsetTop` is >0 and we're moving the session view down by the same amount so that its top aligns with the visual viewport. When the keyboard hides, `offsetTop` returns to 0 and we move the session view back up to the top of the layout viewport.

I'm not sure if this poorly crafted ASCII graphic helps but essentially when the keyboard shows, we're going from this

```
|-- Device screen --| |- Layout viewport -| |- Visual viewport -|
|...................| |...................| |...................|
|...................| |...................| |...................|
|...................| |...................| |...................|
|...................| |...................| |...................|
|...................| |...................| |...................|
|-------------------| |-------------------| |-------------------|
```

to this

```
                      |- Layout viewport -|                       \
                      |...................|                       | offsetTop
                      |...................|                       /
|-- Device screen --| |...................| |- Visual viewport -|
|...................| |...................| |...................|
|...................| |...................| |...................|
|...................| |-------------------| |-------------------|
|...................|                       |----- Keyboard ----|
|...................|                       |...................|
|-------------------|                       |-------------------|
```

src/platform/web/Platform.js
Johennes: Yes, you're right. :facepalm: 

Sorry, I'm going through the comments one by one and hadn't gotten to the one where the class is used yet.

src/main.js
Johennes: That's pretty ingenious, I hadn't thought of this. Much cleaner than changing the style itself in JS.

Johennes:
I addressed most comments but still need to look into generalizing the scrolling. I'll let you know once I'm done with that.

src/platform/web/Platform.js
```diff
@@ -82,6 +82,36 @@ async function loadOlmWorker(paths) {
...
 }
 
+// needed for mobile Safari which does *not* resize the viewport when the keyboard
+// shows or hides (see https://bugs.webkit.org/show_bug.cgi?id=141832)
+function adaptUIOnVisualViewportResize() {
+    if (!window.visualViewport) {
+        return;
+    }
+    window.visualViewport.addEventListener('resize', () => {
+        const sessionView = document.getElementsByClassName('SessionView')[0];
```
bwindels: You can just pass in `this._container` from `Platform` in here and set the custom css property on that as they are inherited, so you don't need the hard-coded selector.

src/platform/web/Platform.js
```diff
@@ -82,6 +82,36 @@ async function loadOlmWorker(paths) {
...
+        if (!sessionView) {
+            return;
+        }
+
+        const timeline = document.getElementsByClassName('Timeline')[0];
```
bwindels: If there's no way make this generic for all scrollable areas, better scope the selector to the container rather than the document, as the idea is that hydrogen won't touch anything outside of that root dom element (for embedding). You could also replace it by `.querySelector(".Timeline")`.

src/platform/web/ui/css/layout.css
```diff
@@ -54,6 +54,11 @@ main {
     min-width: 0;
 }
 
+.hydrogen.ios .SessionView {
```
bwindels, line 59: Perhaps add a comment here why ios is special-cased here.

bwindels (changes requested):
Thanks for addressing the comment! We're close, just a few more small things.

src/main.js
bwindels: Gotcha, thanks!

bwindels: Oh, perhaps just mention in a comment that we need to offset dom element because the document contents is just shifted upwards and the top part is just clipped by the visual viewport.

bwindels:
> I addressed most comments but still need to look into generalizing the scrolling. I'll let you know once I'm done with that.

We actually wouldn't need to do it for all scrollable elements, just for the ones that are bottom aligned, like the timeline, so adding a class (e.g. `bottom-aligned-scroll`) to the Timeline view to look for might be the most straightforward way to do this actually.

src/main.js
Johennes: I reified the comment above `func adaptUIOnVisualViewportResize` a bit. Please let me know in case that's not specific enough.

Johennes:
> > I addressed most comments but still need to look into generalizing the scrolling. I'll let you know once I'm done with that.
> 
> We actually wouldn't need to do it for all scrollable elements, just for the ones that are bottom aligned, like the timeline, so adding a class (e.g. `bottom-aligned-scroll`) to the Timeline view to look for might be the most straightforward way to do this actually.

Yeah, good call. I implemented it like this now.

The only other option I could think of was to traverse all elements in the container and look for scrollable children. However, we might have ended up applying the scroll position magic where it's not needed / wanted. Your proposal makes things more explicit which will probably keep things easier to maintain. :+1:

src/platform/web/Platform.js
```diff
@@ -82,6 +82,31 @@ async function loadOlmWorker(paths) {
     return olmWorker;
 }
 
+// needed for mobile Safari which shifts the layout viewport up without resizing it
+// when the keyboard shows (see https://bugs.webkit.org/show_bug.cgi?id=141832)
+function adaptUIOnVisualViewportResize(container) {
+    if (!window.visualViewport) {
+        return;
+    }
+    window.visualViewport.addEventListener('resize', () => {
```
bwindels: Just realized that for the embedded use-case, it might be good to be able to clean up this event handler, so that in case hydrogen is (dis)mounted multiple times during the lifetime of the page, we don't end up accumulating event handlers.

The easiest way to do this would be to:
 - return a no-argument closure that removes the event handler from this function
 - add an instance of `Disposables` to `Platform`, and add a `dispose` method that forwards it to `this._disposables`.
 - when calling this function, add the returning closure to disposables.

As platform is part of the public API for integrators, they can then call dispose if they need to.

src/platform/web/Platform.js
```diff
@@ -82,6 +82,31 @@ async function loadOlmWorker(paths) {
...
+        const scrollable = container.querySelector('.bottom-aligned-scroll');
+        let scrollTopBefore, heightBefore, heightAfter;
+
+        if (scrollable) {
+            scrollTopBefore = scrollable.scrollTop;
+            heightBefore = scrollable.getBoundingClientRect().height;
```
bwindels:
```suggestion
            heightBefore = scrollable.offsetHeight;
```
Is a bit more concise and _might_ be faster as the whole bounding box doesn't need to be calculated. Same below.

src/platform/web/Platform.js
```diff
@@ -82,6 +82,31 @@ async function loadOlmWorker(paths) {
...
+        }
+
+        container.style.setProperty('--ios-viewport-height', window.visualViewport.height.toString() + 'px');
+        container.style.setProperty('--ios-viewport-top', window.visualViewport.offsetTop.toString() + 'px');
+
+        if (scrollable) {
+            heightAfter = scrollable.getBoundingClientRect().height;
```
bwindels:
```suggestion
            heightAfter = scrollable.offsetHeight;
```

bwindels (changes requested):
This is looking great! Just realized a few more improvements to make, sorry. I appreciate this is a lot of back-and-forth for a small change.

Johennes:
> This is looking great! Just realized a few more improvements to make, sorry. I appreciate this is a lot of back-and-forth for a small change.

No worries really. If anything _I_ should be sorry because explaining things to me probably took a lot longer than just hammering it out yourself. :sweat_smile:

bwindels:
Hmm, just tested it in the simulator. I made a local build (`yarn build`) and then opened `http://localhost:3000/target/` and the build script inlines the css custom properties (it does this for IE11), so it just says `height: 100%; top: 0;` ... argh.

Johennes:
:flushed: 

I had only tested on a device so far and with `yarn start` / `192.168...:3000`. Interestingly it works fine there. I didn't know the `/target` build differs from that. :facepalm:

bwindels:
Yeah, I didn't think about this, sorry. I think for now it would be best to not use css custom properties. I was just looking at how to fix it in the simulator and looks like iOS doesn't like the `position: fixed` fix we have for Android. So, probably best to:
 - the `position: fixed` is only done in `.hydrogen:not(.ios) .SessionView`
 - we set `container.querySelector(".SessionView").style.height/marginTop` from the viewport script for ios, rather than custom properties

Could be wrong though about the `position: fixed`, so any way to make it work would be good.

Johennes:
I'm not sure if this is a better hack or not but the postcss-css-variables plugin that's used in `build.mjs` supports fairly fine grained control of custom properties. We could, for instance, use something like this to preserve all `--ios...` custom properties without impacting any of the others.

```
diff --git a/scripts/build.mjs b/scripts/build.mjs
index dbbeabc..d32ca5d 100644
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -315,7 +315,11 @@ async function buildCssLegacy(entryPath, urlMapper = null) {
     const preCss = await fs.readFile(entryPath, "utf8");
     const options = [
         postcssImport,
-        cssvariables(),
+        cssvariables({
+            preserve: (declaration) => {
+                return declaration.value.indexOf("var(--ios") == 0;
+            }
+        }),
         autoprefixer({overrideBrowserslist: ["IE 11"], grid: "no-autoplace"}),
         flexbugsFixes()
     ];
```

In the generated CSS this ends up as

```
.hydrogen.ios .SessionView {
    height: 100%;
    height: var(--ios-viewport-height, 100%);
    top: 0;
    top: var(--ios-viewport-top, 0);
}
```

We could also trim it down further and require a prefix of `--ios-viewport` since it's (hopefully) unlikely that this exemption is needed elsewhere.

Alternatively, I can revert back to setting the `height` and `top` CSS properties in JS. I'm not sure how `position: fixed` plays into this though. What issues were you seeing with it on the iOS simulator?

bwindels:
Oh, yes, I like preserving css variables starting with `--ios-`! Wrt the position: fixed, I think the workaround still didn't work and I didn't see the header bar. Will test again once the css variables are preserved. Thanks!

Johennes:
Ok, I pushed the change to preserve `--ios-` properties.

The position issue is super odd. I works fine on my device but I tested on the simulator today and it only works fine half of the time there. Occasionally I end up with something like this where it seems as if the `top` offset didn't take effect.

Simulator Screen Shot - iPhone 12 Pro Max - 2021-03-25 at 11 27 56

Will have to do some more digging to figure out what's wrong here. 🤔

Johennes:
Debugging this in the simulator from desktop Safari, I found that `window.visualViewport.offsetTop` is occasionally reported wrong. 🤯

Sometimes when the keyboard shows and the `resize` event fires, `window.visualViewport.offsetTop` is still 0 even though it should be >0 (because the layout viewport was moved up and the visual viewport shrunk accordingly).

Not sure if that is a bug in Safari. One way we could address this is by computing the top offset based on the current offset and the height difference between `SessionView` and the visual viewport. Just quickly hacked this for demonstration purposes and it appears to work fine for me:

```
diff --git a/src/platform/web/Platform.js b/src/platform/web/Platform.js
index 553f322..7d3811a 100644
--- a/src/platform/web/Platform.js
+++ b/src/platform/web/Platform.js
@@ -98,8 +98,11 @@ function adaptUIOnVisualViewportResize(container) {
             heightBefore = scrollable.offsetHeight;
         }

+        const top = container.querySelector('.SessionView').offsetTop;
+        const heightChange = container.querySelector('.SessionView').offsetHeight - window.visualViewport.height;
         container.style.setProperty('--ios-viewport-height', window.visualViewport.height.toString() + 'px');
-        container.style.setProperty('--ios-viewport-top', window.visualViewport.offsetTop.toString() + 'px');
+        // container.style.setProperty('--ios-viewport-top', window.visualViewport.offsetTop.toString() + 'px');
+        container.style.setProperty('--ios-viewport-top', (top + heightChange).toString() + 'px');

         if (scrollable) {
             heightAfter = scrollable.offsetHeight;
```

Johennes:
I pushed a cleaned-up version of the fix shown above and tested on both simulator and device as well as inside and outside of the `/target` build. Seems to work correctly now for me. 🤞

bwindels (approved):
Sorry it took me so long to get back to this. Thank you for your changes and testing. It all looks good to me. I just double checked and tested it on the simulator and seemed to work great, so this is good to go!

Issue #93, opened 2020-09-14
SSO Login


ptman:
SAML / OpenID Connect login support

bwindels:
Hey, thanks for filing an issue, but I would rather not have a feature request issue in here for every missing feature that is not on the immediate roadmap, so closing this for now.

bwindels:
Probably best to raise and discuss feature requests in #hydrogen:matrix.org come to think of it.

bwindels:
Also see https://matrix.org/docs/guides/sso-for-client-developers

tawfiek:
Assign it to me @bwindels

tawfiek:
As discussed with @bwindels 

He suggest : 

> Adding a method `startLogin(server)` to `SessionContainer` that returns a Login object to which the response from the server is passed. Depending on the flows, it could have an sso property and password property with objects that have methods for those specific login mechanisms

And for UX and Component structure, we came out with two suggestions : 
 * 1st
   Make the current  `LoginView`  is the root of the application which takes the `homeserver` URL Then load buttons of supported login methods that linked with individual view  for every  method.

```
LoginView
  homeserver field
  if supported: PasswordLoginView
  if suppoted: SSOLoginView
```
 * 2nd 
 Make another parent component for the current `LoginView` called for example `ClientConfig` that takes `homeserver`  as a user input gets the supported methods and render `LoginView` shows all supported login options just like in Element web. 
 
```
ClientConfig
    homeserver field
    set state
    LoginView
```

bwindels:
Was a bit overly optimistic of me to mark this as a good-first-issue, the changes are quite involved and there isn't a lot of documentation to work with.

bwindels:
I have made some notes on how I expect the feature should work on https://github.com/vector-im/hydrogen-web/blob/master/doc/impl-thoughts/SSO.md

PR #282, opened 2021-03-18
Add SSO functionality  

tawfiek:


tawfiek:
#

tawfiek:
#93

src/platform/web/ui/login/LoginView.js
```diff
@@ -37,39 +37,90 @@ export class LoginView extends TemplateView {
...
-                        }, [vm.i18n`Go Back`]),
-                        t.button({
-                            className: "button-action primary",
-                            type: "submit"
-                        }, vm.i18n`Log In`),
-                    ]),
-                ]),
+                t.if(
+                    (vm) => vm.error,
+                    t.createTemplate((t) =>
```
bwindels: heads up I've removed `createTemplate` in #280, you can now just do `t.if(vm => vm.condition, t => t.div())`

src/domain/LoginViewModel.js
```diff
@@ -93,4 +93,47 @@ export class LoginViewModel extends ViewModel {
...
+        }));
+        this._loadViewModel.start();
+        this.emitChange("loadViewModel");
+        this._loadViewModelSubscription = this._trackLoadViewModelChange();
+    }
+
+    _mapLoginView(loginFlows) {
+        // TODO: map view based on the loginFlows.
+    }
+
```
tawfiek, line 138: Next step : is to use this function with loaded `loginFlows`  to set new state in into `LoginViewModel` that describe the state of  `loginView` I think there is no need to make a new view for every login flow. 
what do you think ?! @bwindels

src/domain/LoginViewModel.js
```diff
@@ -81,4 +93,47 @@ export class LoginViewModel extends ViewModel {
...
+        return this.track(this._loadViewModel.disposableOn("change", _onchangeHandler));
+    }
+
+    /**
+     * This function requests the login flows from _sessionContainer using SessionLoadViewModel
+     * @param {String} homeServer - the homeserver url
+     */
+    _requestLoginFlows (homeServer) {
+        this._loadViewModel  = this.track(new SessionLoadViewModel({
```
bwindels, line 118: Loading the login flows should not set the load view model. The load view model is for once you are logging in, it show "fetching your conversations from the server", etc ... It would be nice to have a spinner though, so you may want to set a `queryingHomeserver` flag or something.

PR #283, opened 2021-03-18
Basic push notification support

bwindels:
- [x] enable push notifications
 - [x] disable push notifications
 - [x] show notification on push message
 - [x] open app & room when clicking notification
 - [x] provide config for main sygnal instance, integrate with build system.
 - [x] tune how push messages create notifications


sw.js
```diff
@@ -0,0 +1 @@
+src/platform/web/service-worker.js
```
bwindels, line 1: Hope this won't prevent a checkout on windows, but https://github.com/git-for-windows/git/wiki/Symbolic-Links suggests it will just be ignored.

Issue #290, opened 2021-03-24
Display verbose message on invalid credentials


Sanskar95:
**Describe the bug**

**To Reproduce**
1. Go to login
2. Type incorrect combination of username and password
3. Scroll down to: 'Same card itself, the invalid prompt is not proper
"Something went wrong: Invalid password on POST https://matrix.org/_matrix/client/r0/login?." appears. I feel thats kinda too much info for the user

**Expected behavior**
I would say something like "Invalid credentials" in red 

**Screenshots**
![image](https://user-images.githubusercontent.com/8667334/112292384-54748100-8c91-11eb-9d89-cccd08b52554.png)

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Ubuntu 20.04
 - Browser [e.g. chrome, safari] Chrome

**Smartphone (please complete the following information):**

**Additional context**

PR #291, opened 2021-03-24
Changed if logic to return unknown error

Sanskar95:


Sanskar95:
This is for fixing https://github.com/vector-im/hydrogen-web/issues/290

bwindels (changes requested):
I'm not sure what this is supposed to do. From a quick glance, you are checking if the load status is both Error and not in the enum? Don't those two conditions exclude each other?

bwindels:
Closing this due to inactivity

PR #300, opened 2021-03-26
Update readme

Prathap-P:
Windows users have to be careful before trying to run locally .Invoking the command prompt as administrator is encouraged. The different behaviour of symlinks in windows is the root cause. Updated the readme file with appropriate instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,6 @@ If you find this interesting, come and discuss on [`#hydrogen:matrix.org`](https
 Hydrogen is deployed to [hydrogen.element.io](https://hydrogen.element.io). You can run it locally `yarn install` (only the first time) and `yarn start` in the terminal, and point your browser to `http://localhost:3000`. If you prefer, you can also [use docker](doc/docker.md).
 
 Hydrogen uses symbolic links in the codebase, so if you are on Windows, have a look at [making git & symlinks work](https://github.com/git-for-windows/git/wiki/Symbolic-Links) there.
-## For Windows Users
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,6 +16,10 @@ If you find this interesting, come and discuss on [`#hydrogen:matrix.org`](https
 
 Hydrogen is deployed to [hydrogen.element.io](https://hydrogen.element.io). You can run it locally `yarn install` (only the first time) and `yarn start` in the terminal, and point your browser to `http://localhost:3000`. If you prefer, you can also [use docker](doc/docker.md).
```
bwindels, line 18: Can we make this more subtle? E.g. just adding this rather than its own section:

```suggestion

Hydrogen uses symbolic links in the codebase, so if you are on Windows, have a look at [making git & symlinks work](https://github.com/git-for-windows/git/wiki/Symbolic-Links) there.
```

bwindels (commented):
Thanks for this! Would prefer to be a bit less subtle in the readme, see suggestion.

README.md
Prathap-P: Yeah that's nice, will commit the suggestion.

Prathap-P: I think the suggestion got appended rather than replacing.

bwindels: Yeah, just remove what you previously wrote and we can merge :+1:

Prathap-P: Will do it.

Prathap-P: Added another commit. I think, now it should be fine.

bwindels (approved):
Thanks! Normally we ask contributors to [sign off](https://github.com/matrix-org/matrix-js-sdk/blob/master/CONTRIBUTING.rst#sign-off) their contribution, but I think in this case it is fine without.

Prathap-P:
Yeah, just now noticed that. Anyway, thanks for guiding.

PR #301, opened 2021-03-27
Add license comments to make GNU LibreJS accept Hydrogen's JS

### LINKED ISSUE #209: [Request] Add licenses for JavaScript according to GNU recommendation
The code in this project is licensed under the Apache License, a free license. However, the license is not embedded into the code in the way [GNU recommended](https://www.gnu.org/software/librejs/free-your-javascript.html). This causes browsers with LibreJS to treat it as nonfree JS, thus blocking it. Please add these license in such format.

LibreJS has detected the following pieces of code and blocked them:

- [Nontrivial script in hydrogen.element.io on line 19](view-source:https://hydrogen.element.io/#line19)
- [External script with no license](https://hydrogen.element.io/hydrogen-1718798642.js)

Johennes:
This adds proper license comments to all _internal_ JS scripts in the `/target` build. This way the code self-identifies as Apache-2.0 which prevents LibreJS from blocking it.

Given the [supported options](https://www.gnu.org/software/librejs/free-your-javascript.html), dynamically adding `// @license ...` / `// @license-end` comments via `build.mjs` seemed like the most practical solution. This only makes the `/target` build work with LibreJS not the development build but achieving the latter would have required annotation _all_ JS source files.

Note that this fix doesn't do anything about _external_ scripts, namely olm. It would be possible to declare olm's license on the Hydrogen level, too. However, I believe this should rather be addressed upstream so that any client that uses olm can benefit from it being LibreJS-compliant (see matrix-org/olm#52).

To test:

- Install the [LibreJS add-on for Firefox](https://addons.mozilla.org/en-US/firefox/addon/librejs/)
- `yarn build && yarn start`
- Load `localhost:3000/target` and notice how all scripts except for olm have been allowed by LibreJS

Screenshot 2021-03-27 at 14 28 13

To test the whole app, wrap the content of `target/olm-....js` with

```
// @license magnet:?xt=urn:btih:8e4f440f4c65981c5bf93c76d35135ba5064d8b7&dn=apache-2.0.txt Apache-2.0
// @source: https://gitlab.matrix.org/matrix-org/olm/-/tree/3.2.2
```

and

```
// @license-end
```

after building.

Closes: #209

Signed-off-by: Johannes Marbach <n0-0ne+github@mailbox.org>


bwindels (commented):
Thanks for this. I'm surprised to not see the license header being added to the generated bundles, only the inline script tags?

Johennes:
You mean the `hydrogen-....js` / `hydrogen-legacy-....js` files? My change should wrap these in the license comments as well. That happens through using `wrapWithLicenseComments` in `buildJs` and `buildJsLegacy`. Just merged in `master` and tested again and I still see it working.

In my screenshot above, you can also see that the license was recognized:

Screenshot 2021-04-28 at 20 26 26

Or did you mean other files? Apologies in case I'm misunderstanding.

bwindels:
> You mean the hydrogen-....js / hydrogen-legacy-....js

Yes, indeed. I guess `the hydrogen.js` bundle is covered because it is imported from a script tag that has the header, but that's not the case for the legacy bundle. I don't think this is necessarily a problem though, as older browsers are probably not supported by the librejs extension anyway.

bwindels:
I'm somewhat hesitant to merge it without having a fix for olm though, as it will leave the app working but have e2ee subtily broken with librejs. Perhaps being completely broken is better, as it's more obvious?

Is there no way to add attributes to the olm script tag to mark it as acceptable for librejs?

Johennes:
> Yes, indeed. I guess the hydrogen.js bundle is covered because it is imported from a script tag that has the header, but that's not the case for the legacy bundle.

No, what I meant is both of those files have the comments applied _inside_ them. The files are generated via the `buildJs` and `buildJsLegacy` functions, respectively, and I'm wrapping the contents with the license comments before returning.

> I'm somewhat hesitant to merge it without having a fix for olm though, as it will leave the app working but have e2ee subtily broken with librejs. Perhaps being completely broken is better, as it's more obvious?

Olm is actually already fixed. @uhoreg helped me add the license comments in https://github.com/matrix-org/olm/issues/52. I think they haven't tagged a new release that includes these changes yet though.

> Is there no way to add attributes to the olm script tag to mark it as acceptable for librejs?

There's a way to declare licenses for external scripts but it involves adding an HTML table. So since the next version of Olm will be compliant, I figured that might not be worth the hassle.

IIRC LibreJS will still break hydrogen entirely even when only Olm is identified as proprietary. Merging the fixes for the Hydrogen code could still be nice though because then people would just have to manually whitelist one script (olm) instead of multiple ones.

Screenshot 2021-04-30 at 16 44 27

That being said, I'd also understand if you'd first want to wait for the next olm release.

bwindels:
> No, what I meant is both of those files have the comments applied _inside_ them. The files are generated via the `buildJs` and `buildJsLegacy` functions, respectively, and I'm wrapping the contents with the license comments before returning.

Indeed so, I missed that :man_facepalming: 

So yeah, this looks good to go then!

PR #307, opened 2021-03-31
Allow specifying a custom defaultHomeServer

Half-Shot:
I think this works :)


bwindels:
Lgtm, can you just add this option as well in index.html?

Half-Shot:
done.

bwindels:
Thanks!

PR #311, opened 2021-04-01
Anticipate firefox throwing a SecurityError from sw.register()

bwindels:
and just proceed as if they were not supported


bwindels:
This won't work, we need to await to get the exception, but our bootstrapping logic is sync atm.

Issue #246, opened 2021-02-25
Clickable links


ptman:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I have to try to copy a link from Hydrogen to my browser.

**Describe the solution you'd like**
I would like to be able to click on links to open them in a new browser tab.

**Describe alternatives you've considered**
Using another client.

**Additional context**
![2021-02-25_13-53](https://user-images.githubusercontent.com/24669/109150041-ff684c80-7770-11eb-9ca7-7670c8038040.png)

ptman:
Thank you!

ptman:
@MidhunSureshR This was a really big improvement, excellent work!

PR #314, opened 2021-04-04
Links in chats are now clickable

Ash110:
Created a separate function for rendering chat messages that parses links before rendering to make them clickable. 

Solves [Issue#246](https://github.com/vector-im/hydrogen-web/issues/246)

### DIFF
--- a/src/platform/web/ui/general/html.js
+++ b/src/platform/web/ui/general/html.js
@@ -87,6 +87,26 @@ export function text(str) {
     return document.createTextNode(str);
 }
 
+export function chatText(str) {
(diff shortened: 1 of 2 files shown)


MidhunSureshR:
Hello :wave: ,
I don't think that this makes all valid URLs clickable. You can find a complete list of valid TLDs [here](https://www.icann.org/resources/pages/tlds-2012-02-25-en). 
As an example, try intel.tech.

I'm sure you can rewrite this to not execute the same regex twice? Perhaps using [RegExp.prototype.exec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec) which gives you index and lastIndex.

bwindels (commented):
Thank you for your PR. The approach taken here is not really what I'm looking for, for this feature. I'd like to go with making this part of the domain layer, rather than making this a pure view concern, and the link detection to be somewhat simpler (e.g. http(s)://*).

bwindels:
@MidhunSureshR is picking up this work, so I will be closing this PR. Thank you for your effort nonetheless.

PR #342, opened 2021-05-05
Archive room on leave

### LINKED ISSUE #147: Rooms are not left
We can start by keeping the left room around and just updating the membership in the room summary, and filtering those out
If we want to remove a room, it needs to be removed from:
 - timeline events
 - timeline fragments
 - room summary
 - room state
 - userIdentities
 - outgoing group session
 - incoming group sessions?
 - ...

bwindels:
Fixes #147


bwindels:
(Own) code review looks ok

PR #343, opened 2021-05-06
Dont assume there is a rooms section in the sync

bwindels:
Fixes `Sync failed because of TypeError: Cannot read property 'invite' of undefined`


bwindels:
This was due to https://github.com/matrix-org/synapse/pull/9919 fwiw

PR #345, opened 2021-05-08
Render URLs as clickable links in timeline

MidhunSureshR:
Implements feature request #246.


src/domain/MessageObjectFormat.js
```diff
@@ -0,0 +1,24 @@
+export class MessageObjectFormat {
```
bwindels: I like the concept of this class, but the name is a bit implementation-detail-y. What about `MessageBody` or `MessageBodyBuilder`?

Also, can we put this and the linkifier in `src/domain/session/room/timeline` as it is specific to the timeline?

src/domain/Linkifier.js
```diff
@@ -0,0 +1,173 @@
...
+        this._text = text;
+        this._curr = 0;
+        this._message = new MessageObjectFormat();
+    }
+
+    /**
+     * Separate string into text, newlines and add them into message object.
+     * @param {String} text 
+     */
+    _addTextToMessage(text) {
```
bwindels: This doesn't seem related to linkifying, I wonder if we should move the newline handling to `MessageObjectFormat`. So for every line `MessageObjectFormat` detects, it calls the linkifier with that line to split it up. Perhaps the `Linkifier` class can even be a `linkify` function given it has so little state, where apart from the text you pass in a `callback(isLink, text)` which `MessageObjectFormat` then translates into `insertText` (if `isLink` is `false`) or `insertLink` calls internally. Wdyt?

src/domain/session/room/timeline/tiles/TextTile.js
```diff
@@ -15,15 +15,16 @@ limitations under the License.
...
 
 import {MessageTile} from "./MessageTile.js";
+import { Linkifier } from "../../../../Linkifier.js";
 
 export class TextTile extends MessageTile {
-    get text() {
+    get messageFormat() {
         const content = this._getContent();
-        const body = content && content.body;
+        let body = content && content.body;
```
bwindels: I think the linkifier won't work if body is undefined here, so perhaps we can fall back to an empty string
```suggestion
        let body = content?.body || "";
```

src/domain/session/room/timeline/tiles/TextTile.js
```diff
@@ -15,15 +15,16 @@ limitations under the License.
 */
 
 import {MessageTile} from "./MessageTile.js";
+import { Linkifier } from "../../../../Linkifier.js";
 
 export class TextTile extends MessageTile {
-    get text() {
+    get messageFormat() {
```
bwindels: I'd expect a format to be a format identifier like `html` or something ... not the actual message content. What about just calling this `body`?

src/platform/web/ui/session/room/timeline/TextMessageView.js
```diff
@@ -21,28 +21,30 @@ import {renderMessage} from "./common.js";
...
-            if (elements.length) {
-                elements.push(t.br());
-            }
-            if (line.length) {
-                elements.push(t.span(line));
-            }
+        const children = [];
+        for (const m of value) {
+            const f = this._formatFunction[m.type];
+            const element = f({ obj: m, t: t });
```
bwindels: perhaps don't create an extra object here, but just pass `m` and `t` as separate arguments? Otherwise looks great!

src/platform/web/ui/session/room/timeline/TextMessageView.js
```diff
@@ -21,28 +21,30 @@ import {renderMessage} from "./common.js";
...
         return renderMessage(t, vm,
             [t.p([bodyView, t.time({className: {hidden: !vm.date}}, vm.date + " " + vm.time)])]
         );
     }
 }
 
 class BodyView extends StaticView {
+
+    get _formatFunction() {
+        return {
```
bwindels: this object can probably just exist as a variable at the module level so it doesn't get recreated every time `_formatFunction` is called.

bwindels (changes requested):
This looks great, thank you!

A few comments about naming, and one about keeping the linkifier just about links, rather than also handling newlines. Other than that, looking great :+1:

src/domain/session/room/timeline/MessageBodyBuilder.js
```diff
@@ -0,0 +1,161 @@
...
+export class MessageBodyBuilder {
+
+    constructor(message = []) {
+        this._root = message;
+    }
+
+    fromText(text) {
+        const components = text.split("\n");
+        components.flatMap(e => ["\n", e]).slice(1).forEach(e => {
+            if (e === "\n")
```
bwindels: please use brackets for all block statement (this should really be in a style guide, sorry)

src/domain/session/room/timeline/MessageBodyBuilder.js
```diff
@@ -0,0 +1,161 @@
...
+                { type: "newline" },
+                { type: "text", text: "This is another sentence." },
+                { type: "newline" }
+            ];
+            test(assert, input, output);
+        },
+
+        // Tests for links
+        "Link with host": assert => {
+            testLink(assert, "https://matrix.org");
```
bwindels: can we split up the tests for links and put them in `linkify.js`?

src/domain/session/room/timeline/linkify.js
```diff
@@ -0,0 +1,12 @@
+export function linkify(text, callback) {
+    const regex = /(?:https|http|ftp):\/\/[a-zA-Z0-9:.\[\]#-]+(?:\/[^\s]*[^\s.,?!]|[^\s\u{80}-\u{10ffff}.,?!])/gui
```
bwindels: define this at module scope so its doesn't get recreated on every invocation?

bwindels: Can we have some comments explaining the different parts of the regex?
I'm also somewhat confused by:
 - the `#` in `[a-zA-Z0-9:.\[\]#-]`, hosts can't have a `#` in them AFAIK?
 - `(?:\/[^\s]*[^\s.,?!]|[^\s\u{80}-\u{10ffff}.,?!])`, `\/[^\s]` selects /* until space, `[^\s.,?!]` selects anything but whitespace and punctuation, and why do we need `[^\s\u{80}-\u{10ffff}.,?!]` again (something related to unicode but forgot why "not whitespace" doesn't cut it)? Also, what is the operator precedence of `|` here? E.g. is the left operand `[^\s]*[^\s.,?!]` or `[^\s.,?!]`?

bwindels (changes requested):
Thank you for the changes!
A few more bits, we're close now :)

src/domain/session/room/timeline/linkify.js
MidhunSureshR: > why do we need `[^\s\u{80}-\u{10ffff}.,?!]`

If path/fragment is absent, URL must end with ASCII character.
i.e https://matrix.org🔥 -> 🔥 is not part of URL, https://matrix.org/🔥 -> 🔥 is part of URL.
AFAIK `80-10ffff` is the non-ASCII unicode range.
>Also, what is the operator precedence of | here? E.g. is the left operand `[^\s]*[^\s.,?!]` or `[^\s.,?!]`?

The left operand is  `[^\s]*[^\s.,?!]`.

src/domain/session/room/timeline/linkify/regex.js
```diff
@@ -0,0 +1,36 @@
...
+
+/*
+Things to keep in mind:
+1.  URL must not contain non-ascii characters in host but may contain
+    them in path or fragment components.
+    https://matrix.org/<smiley> - valid
+    https://matrix.org<smiley> - invalid
+2. Do not treat punctuation at the end as a part of the URL (.,?!)
+*/
+const urlRegex = `${scheme}${host}+(?:${additional}|${endASCII})`;
```
bwindels: endASCII is basically checking the last character of the host part, right? It should have the same constraints as `host`, and additionally not accept any punctuation. This is true whether there is a path/fragment or not. So could we instead simplify to this?:
```js
const scheme = "(?:https|http|ftp):\\/\\/";
const noSpaceNorPunctuation = "[^\\s.,?!]";
const hostCharacter = "[a-zA-Z0-9:.\\[\\]-]";
// using non-consuming group here to combine two criteria for the last character
const host = `${hostCharacter}*(?=${hostCharacter})${noSpaceNorPunctuation}`;
// use sub groups so we accept just / or #, but if anything comes after it should not end with punctuation
const pathOrFragment = `(?:[\\/#](?:[^\\s]*${noSpaceNorPunctuation})?)`;
// path or fragment is optional
const urlRegex = `${scheme}${host}${pathOrFragment}?`;
```

This way the last character of the host won't accept any characters not accepted by DNS, like `https://matrix.org<`

src/domain/session/room/timeline/linkify/regex.js
```diff
@@ -0,0 +1,36 @@
+/*
```
bwindels, line 1: Thanks for splitting it up, much more readable like this!

src/domain/session/room/timeline/linkify/linkify.js
```diff
@@ -0,0 +1,112 @@
+import { regex } from "./regex.js";
+
+export function linkify(text, callback) {
+    const matches = text.matchAll(regex);
+    let curr = 0;
+    for (let match of matches) {
+        callback(text.slice(curr, match.index), false);
```
bwindels: perhaps assign it to a variable so it's clear what is happening? Same below.
```suggestion
        const precedingText = text.slice(curr, match.index);
        callback(precedingText, false);
```

src/domain/session/room/timeline/linkify/linkify.js
```diff
@@ -0,0 +1,112 @@
...
+            const out = [{ type: "link", text: "https://foo.bar" },
+            { type: "text", text: "\uD83D\uDE03.com" }];
+            test(assert, link, out);
+        },
+
+        "Link with unicode only after / must linkify": assert => {
+            testLink(assert, "https://foo.bar.com/\uD83D\uDE03");
+        },
+
+        "Link with unicode after fragment without path must linkify": assert => {
```
bwindels, line 110: Great to have so many tests? Can we have a test for `https://matrix.org<` as well which seems to be accepted with the current regex (see below)?

bwindels (changes requested):
Thank you for your changes! nearly there, just a few more comments about the regex.

src/domain/session/room/timeline/MessageBodyBuilder.js
```diff
@@ -0,0 +1,96 @@
...
+
+export class MessageBodyBuilder {
+
+    constructor(message = []) {
+        this._root = message;
+    }
+
+    fromText(text) {
+        const components = text.split("\n");
+        components.flatMap(e => ["\n", e]).slice(1).forEach(e => {
```
bwindels, line 11: flatMap is not supported by IE11, the legacy build _should_ include a polyfill for it from corejs and put it in the legacy bundle. We should double check this is actually happening before merging.

bwindels:
I think we have a winner, will just try it out before and then merge :+1: Thanks for seeing this one out :slightly_smiling_face:

bwindels (approved):
I think we have a winner, will just try it out before and then merge :+1: Thanks for seeing this one out :slightly_smiling_face:

PR #361, opened 2021-05-19
Redactions

### LINKED ISSUE #30: Redactions
I don't think we need the "relations infrastructure" for this, as we when we receive a redaction, we either:
 - have the redacted event in storage, and we remove the fields, etc ...
 - don't have the redacted event. We don't need to do anything, and if we fetch it in the future, we will get the already redacted version.

bwindels:
Fixes #30


src/matrix/room/timeline/Timeline.js
```diff
@@ -69,12 +62,58 @@ export class Timeline {
...
+            e => {
+                const params = updater(e);
+                return params ? params : false;
+            },
+        );
+        // now look in remote entries
+        if (!foundInLocalEntries && pe.relatedEventId) {
+            // TODO: ideally iterate in reverse as target is likely to be most recent,
+            // but not easy through ObservableList contract
+            for (const entry of this._remoteEntries) {
```
bwindels: Also use findAndUpdate here?

PR #365, opened 2021-05-23
Add right panel with Room information

MidhunSureshR:


bwindels:
So, I just tried it out, and it looks really nice!! Exciting to see this :smile:  A few functional things I noticed:
 - it would be great if you switch room, that it keep the details panel open or closed, based on the current URL. See `"open-room"` in `parseUrlPath` in `src/domain/navigation/index.js`.
 - I'd put the button to open the right panel as a menu option in the menu that is already there, especially given there is not much functionality there yet. Also, I think ideally it would be a link rather than a button. You'll probably need to adjust the CSS to have the same styling for links with a given class than for menu option buttons.
 - I'm not sure about showing the room id as a fallback of the room alias... I think if there is no alias, we should not show the subtitle. It might be useful to still (always) show the room id somewhere small & grey though, just so you can relate the id to other things if needed, perhaps below encryption?

Haven't looked at the code recently, will do so later.

src/domain/session/room/RoomViewModel.js
```diff
@@ -285,6 +289,13 @@ export class RoomViewModel extends ViewModel {
             console.error(err.stack);
         }
     }
+
+    toggleDetailsPanel() {
+        const isPanelOpen = !!this.navigation.path.get("details");
+        const link = isPanelOpen ? this.urlCreator.urlUntilSegment("room"):
+        this.urlCreator.urlForSegment("details");
+        window.location.href = link;
```
bwindels: view models should not access any DOM interfaces, like window, as they are assume to work outside of browser environments. The easiest way would probably to expose a url property from this view model.

src/domain/session/SessionViewModel.js
```diff
@@ -112,9 +118,15 @@ export class SessionViewModel extends ViewModel {
...
 
+    get roomInfoViewModel() {
+        return this._roomInfoViewModel;
+    }
+
     _updateGrid(roomIds) {
         const changed = !(this._gridViewModel && roomIds);
         const currentRoomId = this.navigation.path.get("room");
+        // Close right-panel if needed
+        this._toggleRoomInformationPanel();
```
bwindels: you should not have to react to a change in the grid rooms set, as the active room will be a room segment after that. So you can just react to a change in the active room, and it will work for both grid and single rooms.

So, concretely, you can call _toggleRoomInformationPanel from `this.track(currentRoomId.subscribe(roomId => ` on line 62, and in your toggle method check if the room id of the current vm, if any is still the active one.

src/platform/web/ui/session/room/RoomView.js
```diff
@@ -69,14 +69,15 @@ export class RoomView extends TemplateView {
...
             }
             if (vm.canForget) {
-                options.push(Menu.option(vm.i18n`Forget room`, () => vm.forgetRoom()));
+                options.push(Menu.optionWithButton(vm.i18n`Forget room`, () => vm.forgetRoom()));
             }
             if (vm.canRejoin) {
-                options.push(Menu.option(vm.i18n`Rejoin room`, () => vm.rejoinRoom()));
+                options.push(Menu.optionWithButton(vm.i18n`Rejoin room`, () => vm.rejoinRoom()));
             }
+            options.push(Menu.optionWithLink(vm.i18n`Room Details`, vm.detailsLink))
```
bwindels: The font-size (and font-family of the button, but that should be fixed upstream) seems off still for link menu options:
image

src/platform/web/ui/session/rightpanel/RoomInfoView.js
```diff
@@ -0,0 +1,47 @@
+import { TemplateView } from "../../general/TemplateView.js";
+import { text, classNames, tag } from "../../general/html.js";
+import { AvatarView } from "../../avatar.js";
+
+export class RoomInfoView extends TemplateView {
+
+    render(t, vm) {
+        const encryptionString = vm.isEncrypted ? "On" : "Off";
```
bwindels: For all user visible strings, please use
```js
vm.i18n`...`
```
which should make it easier to introduce translations once we decide to support them.

src/domain/navigation/index.js
```diff
@@ -35,9 +35,9 @@ function allowsChild(parent, child) {
             return type === "room" || type === "rooms" || type === "settings";
         case "rooms":
             // downside of the approach: both of these will control which tile is selected
-            return type === "room" || type === "empty-grid-tile";
+            return type === "room" || type === "empty-grid-tile" || type === "details";
```
bwindels: Is this needed?
A `details` segment would always only come after a `room` segment, no?
And room can occur either after session or rooms.
So you'd have:
`session > rooms > room > details` for the right panel in the grid view or
`session > room > details` for the right panel in the single room view.

Possible I'm missing something though.

src/domain/session/RoomGridViewModel.js
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {ViewModel} from "../ViewModel.js";
+import { ViewModel } from "../ViewModel.js";
```
bwindels: I imagine your editor is doing this? The (very limited) code guidelines don't say anything either way in this regard, but most places in the project don't put spaces there and it's perhaps best not to clutter the PR with unrelated stylistic changes like this.

src/domain/session/RoomGridViewModel.js
```diff
@@ -84,7 +84,11 @@ export class RoomGridViewModel extends ViewModel {
         }
         const vmo = this._viewModelsObservables[index];
         if (vmo) {
+            const detailsShown = !!this.navigation.path.get("details");
```
bwindels: Also check the value of the segment here, like you do in parseUrlPath

src/platform/web/ui/session/SessionView.js
```diff
@@ -25,13 +25,15 @@ import {StaticView} from "../general/StaticView.js";
...
+import { RoomInfoView } from "./rightpanel/RoomInfoView.js";
 
 export class SessionView extends TemplateView {
     render(t, vm) {
         return t.div({
             className: {
                 "SessionView": true,
-                "middle-shown": vm => !!vm.activeMiddleViewModel
+                "middle-shown": vm => !!vm.activeMiddleViewModel,
+                "SideBarActive": vm => !!vm.roomInfoViewModel
```
bwindels: Granted that the css naming conventions are wildly inconsistent, but perhaps name this similar to the class for the middle area, e.g. `right-shown`?

src/platform/web/ui/general/Menu.js
```diff
@@ -15,33 +15,58 @@ limitations under the License.
...
 class MenuOption {
-    constructor(label, callback) {
+    constructor(label) {
         this.label = label;
-        this.callback = callback;
         this.icon = null;
     }
 
+    setCallback(callback) {
+        this.callback = callback;
```
bwindels: Consider return `this` like in `setIcon` here so it can be used in a builder-like pattern: `new MenuOption().setCallback().setIcon()` ...

src/domain/navigation/index.js
```diff
@@ -113,6 +113,9 @@ export function parseUrlPath(urlPath, currentNavPath, defaultSessionId) {
                 segments.push(roomsSegmentWithRoom(rooms, roomId, currentNavPath));
             }
             segments.push(new Segment("room", roomId));
+            if (currentNavPath.get("details")?.value) {
```
bwindels, line 116: Would you consider writing a small test below for this behaviour? :pray:

src/domain/session/RoomGridViewModel.js
```diff
@@ -196,7 +200,7 @@ export function tests() {
     }
 
     class RoomViewModelObservableMock extends ObservableValue {
-        async initialize() {}
+        async initialize() { }
```
bwindels: Same, please avoid changes like this

src/domain/session/leftpanel/LeftPanelViewModel.js
```diff
@@ -93,11 +93,13 @@ export class LeftPanelViewModel extends ViewModel {
     }
 
     toggleGrid() {
+        const details = this.navigation.path.get("details");
         if (this.gridEnabled) {
             let path = this.navigation.path.until("session");
             const room = this.navigation.path.get("room");
             if (room) {
                 path = path.with(room);
+                path = path.with(details);
```
bwindels: This will throw if details is not present, so need to check here first.

src/domain/session/leftpanel/LeftPanelViewModel.js
```diff
@@ -106,6 +108,7 @@ export class LeftPanelViewModel extends ViewModel {
             if (room) {
                 path = path.with(this.navigation.segment("rooms", [room.value]));
                 path = path.with(room);
+                path = path.with(details);
```
bwindels: Same as above

src/domain/session/rightpanel/RoomInfoViewModel.js
```diff
@@ -0,0 +1,50 @@
+import { ViewModel } from "../../ViewModel.js";
+import { avatarInitials, getIdentifierColorNumber, getAvatarHttpUrl } from "../../avatar.js";
+
+export class RoomInfoViewModel extends ViewModel {
+    constructor(options) {
+        super(options);
+        this._room = options.room;
+        this._roomSummary = this._room._summary._data;
```
bwindels: wrt to `_summary`, properties that start with an `_` are considered private, and it's a no-no to access them from anywhere but `this` (you might see it happen in tests sometimes, but that's probably being lazy). See below how to get access to the values you need.

src/domain/session/rightpanel/RoomInfoViewModel.js
```diff
@@ -0,0 +1,50 @@
...
+    get canonicalAlias() {
+        return this._roomSummary.canonicalAlias;
+    }
+
+    get name() {
+        return this._roomSummary.name || this._room._heroes?._roomName || this._roomSummary.canonicalAlias;
+    }
+
+    get isEncrypted() {
+        return !!this._roomSummary.encryption;
```
bwindels: you can use `this._room.isEncrypted` here, see `BaseRoom` again.

src/domain/session/rightpanel/RoomInfoViewModel.js
```diff
@@ -0,0 +1,50 @@
...
+    get roomId() {
+        return this._room.id;
+    }
+
+    get canonicalAlias() {
+        return this._roomSummary.canonicalAlias;
+    }
+
+    get name() {
+        return this._roomSummary.name || this._room._heroes?._roomName || this._roomSummary.canonicalAlias;
```
bwindels: you don't need to do this for the name, `this._room.name` (see `BaseRoom.name`) will give you the correct name, taking into account heroes and canonical alias.

src/domain/session/rightpanel/RoomInfoViewModel.js
```diff
@@ -0,0 +1,50 @@
...
+        this._room = options.room;
+        this._roomSummary = this._room._summary._data;
+    }
+
+    get roomId() {
+        return this._room.id;
+    }
+
+    get canonicalAlias() {
+        return this._roomSummary.canonicalAlias;
```
bwindels: Please create a getter in BaseRoom for it.

src/domain/session/rightpanel/RoomInfoViewModel.js
```diff
@@ -0,0 +1,50 @@
...
+    get name() {
+        return this._roomSummary.name || this._room._heroes?._roomName || this._roomSummary.canonicalAlias;
+    }
+
+    get isEncrypted() {
+        return !!this._roomSummary.encryption;
+    }
+
+    get memberCount() {
+        return this._roomSummary.joinCount;
```
bwindels: Please create a getter in BaseRoom for it, perhaps `joinedMemberCount`?

src/domain/session/room/RoomViewModel.js
```diff
@@ -106,6 +106,10 @@ export class RoomViewModel extends ViewModel {
     get timelineViewModel() { return this._timelineVM; }
     get isEncrypted() { return this._room.isEncrypted; }
 
+    get roomDetailsLink() {
+        return this.urlCreator.urlForSegment("details");
```
bwindels: Seems like this is not used?

src/domain/session/room/RoomViewModel.js
```diff
@@ -285,7 +289,11 @@ export class RoomViewModel extends ViewModel {
             console.error(err.stack);
         }
     }
-    
+
+    get detailsLink() {
+        return this.urlCreator.urlForSegment("details");
```
bwindels: Probably best to create the URLs in the constructor and store them in a property. The property on a view model should ideally be very fast (and not create new strings, etc), so the bindings reevaluating them won't slow things down.

src/domain/session/rightpanel/RoomInfoViewModel.js
```diff
@@ -0,0 +1,50 @@
...
+    avatarUrl(size) {
+        return getAvatarHttpUrl(this._room.avatarUrl, size, this.platform, this._room.mediaRepository);
+    }
+
+    get avatarTitle() {
+        return this.name;
+    }
+
+    get closeLink() {
+        return this.urlCreator.urlUntilSegment("room");
```
bwindels: Probably best to create the URLs in the constructor and store them in a property.

src/platform/web/ui/session/room/RoomView.js
bwindels: Also, perhaps put this option first in the menu, so you're less likely to hit (the destructive) "Leave room" option by accident.

src/domain/session/SessionViewModel.js
```diff
@@ -78,6 +80,11 @@ export class SessionViewModel extends ViewModel {
             this._updateLightbox(eventId);
         }));
         this._updateLightbox(lightbox.get());
+
+        const details = this.navigation.observe("details");
+        this.track(details.subscribe(() => this._toggleRoomInformationPanel()));
+        this._toggleRoomInformationPanel();
+
```
bwindels: No need for this empty line at the end of the method probably?

src/domain/session/SessionViewModel.js
```diff
@@ -246,4 +256,21 @@ export class SessionViewModel extends ViewModel {
...
+    _roomFromNavigation() {
+        const roomId = this.navigation.path.get("room")?.value;
+        const room = this._sessionContainer.session.rooms.get(roomId);
+        return room;
+    }
+
+    _toggleRoomInformationPanel() {
+        this._roomInfoViewModel = this.disposeTracked(this._roomInfoViewModel);
+        const room = this._roomFromNavigation();
+        const enable = !!this.navigation.path.get("details");
```
bwindels: Like elsewhere, also check the value of the segment here, if present.

src/domain/session/SessionViewModel.js
```diff
@@ -246,4 +256,21 @@ export class SessionViewModel extends ViewModel {
...
+
+    _roomFromNavigation() {
+        const roomId = this.navigation.path.get("room")?.value;
+        const room = this._sessionContainer.session.rooms.get(roomId);
+        return room;
+    }
+
+    _toggleRoomInformationPanel() {
+        this._roomInfoViewModel = this.disposeTracked(this._roomInfoViewModel);
+        const room = this._roomFromNavigation();
```
bwindels: Do this inside the `if(enable)`?

src/domain/navigation/index.js
```diff
@@ -35,9 +35,9 @@ function allowsChild(parent, child) {
             return type === "room" || type === "rooms" || type === "settings";
         case "rooms":
             // downside of the approach: both of these will control which tile is selected
-            return type === "room" || type === "empty-grid-tile";
+            return type === "room" || type === "empty-grid-tile" || type === "details";
         case "room":
-            return type === "lightbox";
+            return type === "lightbox" || type === "details";
```
bwindels, line 40: Hmm, just realized that opening the lightbox will close the details and vice-versa ... not sure how to solve this, but fine for now. Perhaps lightbox should not be part of the navigation path :thinking: as it is also annoying it ends up in the history, and is more of a modal thing. Anyway .... nothing for you to do here :smile:

src/domain/session/RoomGridViewModel.js
```diff
@@ -84,7 +84,11 @@ export class RoomGridViewModel extends ViewModel {
         }
         const vmo = this._viewModelsObservables[index];
         if (vmo) {
+            const detailsShown = !!this.navigation.path.get("details");
             this.navigation.push("room", vmo.id);
+            if (detailsShown) {
+                this.navigation.push("details", true);
```
bwindels: Every push will cause the navigation to update. It's nice to make all the navigation changes in one go, so I'd suggest creating a new path (with room id and details) like already done in the left panel view model, and then applying it in one go, rather than multiple pushes.

src/platform/web/ui/general/Menu.js
```diff
@@ -15,33 +15,58 @@ limitations under the License.
 */
 
 import {TemplateView} from "./TemplateView.js";
+import { tag } from "./html.js";
 
 export class Menu extends TemplateView {
-    static option(label, callback) {
-        return new MenuOption(label, callback);
+    static optionWithButton(label, callback) {
```
bwindels: This is already used in the composer (and since redactions is merged also in BaseMessageView), so either keep the name, or ensure you've applied the name change everywhere.

src/domain/session/rightpanel/RoomInfoViewModel.js
```diff
@@ -0,0 +1,50 @@
+import { ViewModel } from "../../ViewModel.js";
+import { avatarInitials, getIdentifierColorNumber, getAvatarHttpUrl } from "../../avatar.js";
+
+export class RoomInfoViewModel extends ViewModel {
+    constructor(options) {
+        super(options);
+        this._room = options.room;
```
bwindels: To receive updates on any properties of the room, you'll need to listen to its `change` event and call `emitChange`. Currently, if the room name, avatar or join count, ... would change, the right panel would not update. RoomViewModel does exactly the same, so serves as a good example.

src/platform/web/ui/css/right-panel.css
```diff
@@ -0,0 +1,31 @@
...
+
+.RoomAvatar {
+    display: flex;
+}
+
+.RoomName h2 {
+    text-align: center;
+}
+
+.SidebarRow {
```
bwindels: Were you envisioning that this be used in other sidebars in the app? I'd prefer calling anything related to the right panel that, rather than sidebar. This is consistent with left panel and is more specific which panel is meant.

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -43,14 +41,37 @@ limitations under the License.
...
 
-.hydrogen .avatar.usercolor1 { background-color: var(--usercolor1); }
-.hydrogen .avatar.usercolor2 { background-color: var(--usercolor2); }
-.hydrogen .avatar.usercolor3 { background-color: var(--usercolor3); }
-.hydrogen .avatar.usercolor4 { background-color: var(--usercolor4); }
-.hydrogen .avatar.usercolor5 { background-color: var(--usercolor5); }
-.hydrogen .avatar.usercolor6 { background-color: var(--usercolor6); }
-.hydrogen .avatar.usercolor7 { background-color: var(--usercolor7); }
-.hydrogen .avatar.usercolor8 { background-color: var(--usercolor8); }
+.hydrogen .avatar.usercolor1 {
```
bwindels: I imagine your tooling got ahead of you here, but please don't make unrelated stylistic changes like this. For one, it makes it harder to see where you made actual changes in this file.

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -932,3 +978,87 @@ button.link {
     max-width: 200px;
     width: 100%;
 }
+
+/* Right Panel */
+
+.RoomInfo {
```
bwindels: So, as said, the CSS naming conventions are currently wildly inconsistent. But for CSS classes that are not used in multiple views (which I think applies to all the below?), I do try to use the convention of `ViewName` for the root element, and `ViewName_something` for sub elements.

src/domain/session/rightpanel/RoomInfoViewModel.js
```diff
@@ -0,0 +1,50 @@
+import { ViewModel } from "../../ViewModel.js";
+import { avatarInitials, getIdentifierColorNumber, getAvatarHttpUrl } from "../../avatar.js";
+
+export class RoomInfoViewModel extends ViewModel {
```
bwindels: can we call this `RoomDetails{View|ViewModel}` everywhere? It's consistent with the details segment in the nav/url. Also the CSS classes, see comments in `element/theme.css`

src/domain/session/SessionViewModel.js
```diff
@@ -112,6 +119,10 @@ export class SessionViewModel extends ViewModel {
         return this._roomViewModelObservable?.get();
     }
 
+    get roomInfoViewModel() {
```
bwindels: This would also become `roomDetailsViewModel` ...

src/platform/web/ui/css/themes/element/theme.css
bwindels: As commented in RoomInfoViewModel, I'd suggest the name `RoomDetailsView` here.

I imagine once you add the member list to the right panel. RoomDetailsView would be one possible child of RightPanelView or something, with `MemberListView` being the other possibility.

bwindels (changes requested):
Thank you for your changed @MidhunSureshR! This is shaping up nicely! I did a full code review, don't be intimidated by the many comments, most should be very easy to address.

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -932,3 +978,87 @@ button.link {
...
+    width: 100%;
+}
+
+.RoomName h2 {
+    margin-bottom: 4px;
+    font-size: 1.8rem;
+}
+
+.SidebarRow {
+    width: 90%;
```
bwindels: Wonder what this is for? It is adding space on the right, making it look somewhat unbalanced:
image

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -932,3 +978,87 @@ button.link {
     max-width: 200px;
     width: 100%;
 }
+
+/* Right Panel */
+
+.RoomInfo {
+    background: rgba(245, 245, 245, 0.90);
+    padding: 10px;
```
bwindels: I'd go for 16px here. Our designer recommended me to always use values multiple of 4px in any case, so you end up with a 4px grid that looks more coherent.

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -932,3 +978,87 @@ button.link {
...
+}
+
+.RoomAvatar {
+    margin-top: 44px;
```
bwindels: Hmm, not sure about this. I think I prefer how it looks without this (and the other changes I suggested):

image

bwindels (changes requested):
Had a look at the styling as well, mostly looks good, just a few suggestions.

src/platform/web/ui/session/rightpanel/RoomDetailsView.js
```diff
@@ -0,0 +1,51 @@
...
+
+    _createRoomAliasDisplay(vm) {
+        return vm.canonicalAlias ? tag.div({className: "RoomDetailsView_id"}, [text(vm.canonicalAlias)]) :
+            "";
+    }
+
+    _createRightPanelRow(t, label, labelClass, value) {
+        const labelClassString = classNames({RoomDetailsView_label: true, ...labelClass});
+        return t.div({className: "RoomDetailsView_row"}, [
+            t.div({className: labelClassString}, [text(label)]),
```
bwindels: Just a small note that you don't need `text()` here. It's only when working with node bindings where you explicitly need to return a DOM node you need to do this. t.<element>() will convert any children strings to text nodes.

src/domain/session/room/RoomViewModel.js
```diff
@@ -283,10 +283,16 @@ export class RoomViewModel extends ViewModel {
             console.error(err.stack);
         }
     }
-    
+
     get composerViewModel() {
         return this._composerVM;
     }
+
+    toggleDetailsPanel() {
```
bwindels: the implementation looks like it will just open rather than toggle? Change name?

src/domain/session/rightpanel/RoomDetailsViewModel.js
```diff
@@ -0,0 +1,51 @@
+import {ViewModel} from "../../ViewModel.js";
+import {avatarInitials, getIdentifierColorNumber, getAvatarHttpUrl} from "../../avatar.js";
+
+export class RoomDetailsViewModel extends ViewModel {
+    constructor(options) {
+        super(options);
+        this._room = options.room;
+        this._room.on("change", () => this.emitChange());
```
bwindels: You'll need to unsubscribe in the dispose method you override from ViewModel, which also means you can't use an anonymous lambda here. See RoomViewModel

bwindels (commented):
Thank you for addressing all the comments, just went over those and resolved accordingly. Came across one issue and a little remaining nit. I'll do another review of the whole thing (rather than just going over the comments as I did now), but I think we're mostly good to go.

bwindels:
Can you also fix some of the lint warnings in your new code? The current warning count is 16 and it'd be good if it didn't go up.

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -660,14 +662,19 @@ button.link {
...
+
+.menu .menu-item{
     border-radius: 4px;
-    display: block;
     border: none;
-    width: 100%;
     background-color: transparent;
     text-align: left;
     padding: 8px 32px 8px 8px;
+    font-size: 1.6rem;
```
bwindels: This should be 15px aka `1.5rem`

src/platform/web/ui/css/layout.css
```diff
@@ -65,7 +72,7 @@ the layout viewport up without resizing it when the keyboard shows */
 .middle .close-middle { display: none; }
 /* mobile layout */
 @media screen and (max-width: 800px) {
-    .SessionView:not(.middle-shown) {
+    .SessionView:not(.middle-shown, .right-shown) {
```
bwindels: selector lists in `:not` are not supported in IE [apparently](https://developer.mozilla.org/en-US/docs/Web/CSS/:not). Can you write this out?

bwindels (changes requested):
Looking great, found a few more things, but after that we're good to merge I think.

src/domain/session/RoomGridViewModel.js
```diff
@@ -183,6 +189,7 @@ export class RoomGridViewModel extends ViewModel {
 
 import {createNavigation} from "../navigation/index.js";
 import {ObservableValue} from "../../observable/ObservableValue.js";
+import { Segment } from "../navigation/Navigation.js";
```
bwindels: Nit: only the imports needed for the tests should be below here, the others up at the top. You can also avoid the import by doing `this.navigation.segment(...)` fwiw.

src/platform/web/ui/css/layout.css
```diff
@@ -72,7 +72,7 @@ the layout viewport up without resizing it when the keyboard shows */
 .middle .close-middle { display: none; }
 /* mobile layout */
 @media screen and (max-width: 800px) {
-    .SessionView:not(.middle-shown, .right-shown) {
+    .SessionView:not(.middle-shown) {
```
bwindels: is the `.right-shown` not needed then in here?

bwindels (approved):

PR #369, opened 2021-05-28
Replace ellipsis character in RoomView header with svg icon

MidhunSureshR:


src/platform/web/ui/session/room/RoomView.js
```diff
@@ -47,7 +47,7 @@ export class RoomView extends TemplateView {
                 t.button({
                     className: "button-utility room-options",
                     onClick: evt => this._toggleOptionsMenu(evt)
```
bwindels, line 50: Could you add an `aria-label` with
```
vm.i18`Room options`
```
as well?

bwindels (approved):
Looks great, just one small thing

PR #373, opened 2021-06-02
Fix race between /send and /sync

bwindels:


src/matrix/storage/idb/StorageFactory.js
```diff
@@ -21,14 +21,18 @@ import { schema } from "./schema.js";
...
+const openDatabaseWithSessionId = function(sessionId, idbFactory) {
+    return openDatabase(sessionName(sessionId), createStores, schema.length, idbFactory);
+}
 
 async function requestPersistedStorage() {
-    if (navigator?.storage?.persist) {
-        return await navigator.storage.persist();
-    } else if (document.requestStorageAccess) {
+    // don't assume browser so we can run in node with fake-idb
+    const glob = this;
```
MadLittleMods, line 30: This doesn't work in strict mode where `this` is `undefined`. The SDK `hydrogen.cjs.js` has  `"use strict";` at the top.

Docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#function_context

PR #384, opened 2021-06-03
Reactions

### LINKED ISSUE #132: Reactions

bwindels:
Fixes #132


src/matrix/room/timeline/entries/BaseEventEntry.js
```diff
@@ -39,27 +48,38 @@ export class BaseEventEntry extends BaseEntry {
     }
 
     /**
-        aggregates local relation.
+        aggregates local relation or local redaction of remote relation.
         @return [string] returns the name of the field that has changed, if any
     */
     addLocalRelation(entry) {
-        if (entry.eventType === REDACTION_TYPE) {
+        if (entry.eventType === REDACTION_TYPE && entry.relatedEventId === this.id) {
```
bwindels: this will break local echo for redactions when the target is a pending event that is sending but doesn't have a remote id yet

src/matrix/room/timeline/entries/BaseEventEntry.js
```diff
@@ -80,4 +111,35 @@ export class BaseEventEntry extends BaseEntry {
             // so don't clear _pendingRedactions here
         }
     }
-}
\ No newline at end of file
+
+    get pendingRedaction() {
```
bwindels, line 143: this looks to be unused

bwindels: might still be useful if you want to get the redaction reason during local echo?

src/platform/web/ui/css/themes/element/timeline.css
```diff
@@ -211,6 +213,42 @@ only loads when the top comes into view*/
...
+    user-select: none;
+    vertical-align: middle;
+}
+
+.Timeline_messageReactions button.haveReacted {
+    background-color: #e9fff9;
+    border-color: #0DBD8B;
+}
+
+@keyframes glow-reaction-border {
```
bwindels, line 244: respect `prefers-reduced-motion` here?

src/domain/session/room/timeline/ReactionsViewModel.js
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    async toggleReaction() {
+        if (this._isToggling) {
+            return;
+        }
+        this._isToggling = true;
+        try {
+            // TODO: should some of this go into BaseMessageTile?
```
bwindels: it probably should

src/platform/web/ui/css/themes/element/timeline.css
bwindels: that only seems to be adviced for animations that change geometry or position, not the case here

PR #390, opened 2021-06-04
Allow registering and logging in as guest

rxl881:
This PR is intended to enable (registering and) logging in as a guest user.

When `guest === true` is passed to startWithLogin the username and password parameters will be ignored (should be passed as `null`) and guest registration and login will be attempted. On success the registered username and accessToken will be returned to match a normal user / password login.

### DIFF
--- a/src/matrix/SessionContainer.js
+++ b/src/matrix/SessionContainer.js
@@ -97,7 +97,7 @@ export class SessionContainer {
         });
     }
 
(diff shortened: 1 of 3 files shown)


bwindels:
This is now outdated and @rxl881 has indicated out of band that he doesn't need this functionality anymore, so when we do need it, we can do it again based on the latest API of SessionContainer.

PR #392, opened 2021-06-08
Documentation for Hydrogen

MidhunSureshR:

### DIFF
--- a/documentation/UI/render-dom-elements.md
+++ b/documentation/UI/render-dom-elements.md
@@ -1,0 +1,18 @@
+There are two options to render DOM elements:
+- Use `tag` from `ui/general/html.js`
+- Use `TemplateBuilder` object (t) from the render function in the view.
+
+Although syntactically similar, they are not functionally equivalent.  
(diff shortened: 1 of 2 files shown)


documentation/UI/render-dom-elements.md
```diff
@@ -0,0 +1,18 @@
+There are two options to render DOM elements:
+- Use `tag` from `ui/general/html.js`
+- Use `TemplateBuilder` object (t) from the render function in the view.
```
bwindels: Not any view, just a TemplateView

Also, I'd perhaps structure this a bit differently. First introduce `tag` as just a quicker way to create dom nodes, then add templateview on top of that and say it adds bindings of all kinds and event handlers, with a footnote that tag and t for template builder should not be confused, as the bindings will be ignored by tag.

MidhunSureshR: I will restructure this as you suggested.

bwindels:
I'll merge this, we can improve it later but no use keeping this PR open.

PR #395, opened 2021-06-18
Add memberlist right-panel to Hydrogen

MidhunSureshR:


bwindels:
For https://github.com/vector-im/hydrogen-web/issues/403

src/domain/session/leftpanel/LeftPanelViewModel.js
```diff
@@ -93,8 +93,13 @@ export class LeftPanelViewModel extends ViewModel {
     }
 
     _pathForDetails(path) {
+        let _path = path;
         const details = this.navigation.path.get("details");
-        return details?.value ? path.with(details) : path;
+        if (details?.value) {
```
bwindels: This will need to be generic to copy any segments over that come after a `right-panel` segment.

src/domain/session/rightpanel/comparator.js
```diff
@@ -0,0 +1,98 @@
+import {PowerLevels} from "../../../matrix/room/timeline/PowerLevels.js";
```
bwindels: Perhaps put this file (comparator.js) in a `members` subdirectory?

src/domain/session/rightpanel/disambiguator.js
```diff
@@ -0,0 +1,147 @@
+export class Disambiguator {
```
bwindels: same here, can be in `members` sub dir

src/domain/session/rightpanel/disambiguator.js
```diff
@@ -0,0 +1,147 @@
...
+            this._unDisambiguate(vm, value);
+            if (value.length === 1) { this._flatten(previousName, value); }
+        } else {
+            this._map.delete(previousName);
+        }
+    }
+
+    _updateMap(vm) {
+        const name = vm.name;
+        if (this._map.has(name)) {
```
bwindels: You can prevent the double lookup here by doing `get` immediately and check if the result is truthy

src/domain/session/rightpanel/disambiguator.js
```diff
@@ -0,0 +1,147 @@
...
+            }
+        } else {
+            this._map.set(name, vm);
+        }
+    }
+
+    disambiguate(vm) {
+        if (!vm.nameChanged) { return; }
+        this._handlePreviousName(vm);
+        this._updateMap(vm);
```
bwindels: You could return the array of `vm`s from this method (only if it is an array), so you don't need to look it up again here.

src/domain/session/rightpanel/disambiguator.js
```diff
@@ -0,0 +1,147 @@
...
+        const idx = array.indexOf(vm);
+        if (idx !== -1) {
+            const [removed] = array.splice(idx, 1);
+            removed.setDisambiguation(false);
+        }
+    }
+
+    _handlePreviousName(vm) {
+        const previousName = vm.previousName;
+        if (!previousName) { return; }
```
bwindels: Perhaps check that previousName is not a string or is not undefined rather than falsy in case somebody has an empty name

src/domain/session/rightpanel/disambiguator.js
```diff
@@ -0,0 +1,147 @@
+export class Disambiguator {
+    constructor() {
+        this._map = new Map();
+    }
+
+    _flatten(name, array) {
+        const vm = array.pop();
```
bwindels: Can we write this inline in `_handlePreviousName` perhaps? I find it a bit harder to read when this is disconnected from the `length === 1` check. Then you could do `array[0]` for more clarity.

src/platform/web/ui/session/rightpanel/LoadingView.js
```diff
@@ -0,0 +1,8 @@
+import {TemplateView} from "../../general/TemplateView.js";
+import {spinner} from "../../common.js";
+
+export class LoadingView extends TemplateView {
```
bwindels: I wonder if it is worth it to create a generic loading view that accepts a message and perhaps an extra css class, and by default just center the message and spinner on both axis, like `TimelineLoadingView`.

src/platform/web/ui/session/rightpanel/MemberListView.js
```diff
@@ -0,0 +1,9 @@
+import {TemplateView} from "../../general/TemplateView.js";
+import {ListView} from "../../general/ListView.js";
+import {MemberTileView} from "./MemberTileView.js";
+
+export class MemberListView extends TemplateView {
+    render(t, vm) {
+        return t.view(new ListView({list: vm.memberTileViewModels, className:"MemberListView"}, tileViewModel => new MemberTileView(tileViewModel)));
```
bwindels: You can just define MemberListView as a sub class of `ListView`, providing a convenience constructor. Or do you plan to add more things to this view?

src/matrix/room/BaseRoom.js
```diff
@@ -378,6 +378,10 @@ export class BaseRoom extends EventEmitter {
         return this._summary.data.membership;
     }
 
+    get powerLevels() {
+        return this._timeline.powerLevels;
```
bwindels: `this._timeline` is only set if the timeline is open, which will be fine for using it in the in the member list, but it will throw if `openTimeline` is not called before `loadMemberlist`. Ideally we'd:

 - lazy load the power levels in BaseRoom either when loading the members or opening the timeline, and pass them into Timeline and Memberlist.
 - if loaded, update the power levels in afterSync of Room
 - resort the member list when the power levels change in afterSync.

bwindels (commented):
This is looking great! I tried it out and the sorting and disambiguation seemed to work quite well.
I came across a few things in this first review, see the comments.

src/platform/web/ui/session/rightpanel/MemberListView.js
MidhunSureshR: Will keep this in mind. But some of this is bound to change with lazy loading I think.

src/platform/web/ui/session/rightpanel/LoadingView.js
MidhunSureshR: This is pretty much what I thought too. We'd then replace TimelineLoadingView with the generic loading view.

src/domain/session/leftpanel/LeftPanelViewModel.js
MidhunSureshR: Yep. Haven't gotten around to making the navigation work everywhere. Planning on doing it right after lazy loading. 👍

src/domain/session/rightpanel/MemberListViewModel.js
```diff
@@ -0,0 +1,35 @@
...
+    _mapTileViewModels(members) {
+        const mapper = (member, emitChange) => {
+            const mediaRepository = this.mediaRepository;
+            const vm = new MemberTileViewModel(this.childOptions({member, emitChange, mediaRepository}));
+            this.nameDisambiguator.disambiguate(vm);
+            return vm;
+        }
+        const updater = (vm, params, newMember) => {
+            vm.updateFrom(newMember);
+            this.nameDisambiguator.disambiguate(vm);
```
MidhunSureshR, line 37: Noticed something here. `MappedMap` won't run the mapper in `onRemove`. This will cause any user who leaves and rejoins a room to appear disambiguated.

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,128 @@
...
+
+    _calculateInitialRenderCount() {
+        return Math.ceil(this._height / this._itemHeight);
+    }
+
+    loadList() {
+        if (!this._list) {
+            return;
+        }
+        this._subscription = this._list.subscribe(this);
```
bwindels, line 172: any reason you can't leave this up to the parent class by calling `super`? Should be enough to just override `onAdd`, ...?

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,128 @@
...
+        const start = this._expanded ? this._initialStart : this.start;
+        return idx >= start && idx <= this.end;
+    }
+}
+
+export class LazyListView extends ListView {
+    constructor({itemHeight, height, appendCount = 5, ...options}, childCreator) {
+        super(options, childCreator);
+        this._itemHeight = itemHeight;
+        this._height = height;
```
bwindels: I'd try to determine this inside this component using the DOM rather than inject (& hardcode) it. It does mean you might need to make the initial rendering async as in render at first you'd only attach the `ul` to measure the height of it, and only then mount the needed amount of children. In my experience, doing a double `await new Promise(r => requestAnimationFrame(r))` does a good job of reading a reliably `clientHeight`.

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,128 @@
...
+
+    _isFullyScrolled() {
+        return this._root.scrollHeight - Math.abs(this._root.scrollTop) === this._root.clientHeight;
+    }
+
+    _renderMoreIfNeeded() {
+        if (!this._isFullyScrolled()) {
+            return;
+        }
+        this._range.expandFromEnd(this._appendCount);
```
bwindels: I think this will add `n` (default of 5) items *every* time you scroll a pixel? I'd like to interpret this number as a padding rather, where you have the range the corresponds to the current viewport and you add `n` items on either end (limited by 0 and the length of the list on either end of course).

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,128 @@
...
+
+class Range {
+    constructor(start = 0, end = 0) {
+        this.start = start;
+        this.end = end;
+        this._expanded = false;
+    }
+
+    _onInitialExpand() {
+        if (this._expanded) { return; }
```
bwindels: What does expand mean here? Move the window of the range to react to scrolling? Is it inspired [on this](https://github.com/matrix-org/matrix-react-sdk/blob/9d6d8fc666855ca0c06c71ccb30c74ac4fc8fd12/src/components/views/elements/LazyRenderList.js#L39)? It seems to do something somewhat different though.

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,128 @@
...
+        const fromInRange = this._range.contains(fromIdx);
+        const toInRange = this._range.contains(toIdx);
+        if (fromInRange && toInRange) {
+            super.onMove(fromIdx, toIdx, value);
+        }
+        else if (fromInRange && !toInRange) {
+            this.onBeforeListChanged();
+            const [child] = this._childInstances.splice(fromIdx, 1);
+            child.root().remove();
+            this.onListChanged();
```
bwindels, line 251: you'll need to render an item at the end of the list though to keep the count the same

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,128 @@
...
+        }
+        else if (!fromInRange && toInRange) {
+            this.onBeforeListChanged();
+            const child = this._childCreator(value);
+            this._childInstances.splice(toIdx, 0, child);
+            insertAt(this._root, toIdx, mountView(child, this._mountArgs));
```
bwindels: same here, you'll need to remove an item at the end of the list to keep the count the same

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    _onInitialExpand() {
+        if (this._expanded) { return; }
+        this._initialStart = this.start;
+        this._expanded = true;
+    }
+
+    expandFromEnd(units) {
+        this._onInitialExpand();
```
bwindels: I'm failing to understand what the idea here is

bwindels (commented):
Nice work so far, left some comments on the lazy list view.

src/domain/session/rightpanel/MemberListViewModel.js
bwindels: Ah yes, can you check for this in the mapper when it is run in onAdd?

src/platform/web/ui/general/LazyListView.js
MidhunSureshR: Because `super.loadList()` also inserts all the elements from list into `_root`.

src/platform/web/ui/session/rightpanel/MemberTileView.js
```diff
@@ -0,0 +1,11 @@
+import {TemplateView} from "../../general/TemplateView.js";
+import {AvatarView} from "../../avatar.js";
+
+export class MemberTileView extends TemplateView {
+    render(t, vm) {
+        return t.div({ className: "MemberTileView" }, [
```
bwindels: the root element should probably be an `<li>` here as it's mounted within an `<ul>`.

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,250 @@
...
+        const renderCount = Math.min(visibleItems, itemsAfterTop);
+        const bottomCount = itemsAfterTop - renderCount;
+        return new ItemRange(topCount, renderCount, bottomCount);
+    }
+
+    _renderIfNeeded() {
+        const range = this._getVisibleRange();
+        const intersectRange = range.expand(this._overflowMargin);
+        const renderRange = range.expand(this._overflowItems);
+        const listHasChangedSize = !!this._renderRange && this._list.length !== this._renderRange.totalSize();
```
bwindels: `listHasChangedSize` can be removed here and in the check below, as any changes to the list will be handled in the onAdd/onMove/onRemove/... handlers.

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,250 @@
...
+        map index from list to index in rendered range
+        eg: if the index range of this._list is [0, 200] and we have rendered
+        elements in range [50, 60] then index 50 in list must map to index 0
+        in DOM tree/childInstance array.
+        */
+        return idx - this.topCount;
+    }
+}
+
+export class LazyListView extends ListView {
```
bwindels, line 57: I wonder how much behaviour is still used from `ListView`? Do you think it could be simpler to just implement this from scratch?

src/matrix/room/BaseRoom.js
```diff
@@ -378,8 +381,44 @@ export class BaseRoom extends EventEmitter {
         return this._summary.data.membership;
     }
 
-    get powerLevels() {
-        return this._timeline.powerLevels;
+    async loadPowerLevels() {
```
bwindels: this should be private (e.g. _load...)

src/matrix/room/BaseRoom.js
```diff
@@ -378,8 +381,44 @@ export class BaseRoom extends EventEmitter {
...
+        const timelineReader = new TimelineReader({
```
bwindels: The timeline reader is not needed for the power levels. This is the part the reads the messages in the Timeline class.

src/matrix/room/BaseRoom.js
```diff
@@ -378,8 +381,44 @@ export class BaseRoom extends EventEmitter {
...
+                membership: this.membership
+            });
+        } else {
+            const membership = this.membership;
+            return new PowerLevels({ownUserId: this._user.id, membership});
+        }
+    }
+
+    async observePowerLevels() {
+        let observable = this._powerLevels;
```
bwindels, line 425: as the function currently is, if a second call is made before `load...` finishes for the first call, it will load the power levels twice. I think you will need to cache the promise from `load...` and store it on `this` and await that if it is set.

src/matrix/room/BaseRoom.js
```diff
@@ -378,8 +381,44 @@ export class BaseRoom extends EventEmitter {
...
-    get powerLevels() {
-        return this._timeline.powerLevels;
+    async loadPowerLevels() {
+        const timelineReader = new TimelineReader({
+            roomId: this._roomId,
+            storage: this._storage,
+            fragmentIdComparer: this._fragmentIdComparer
+        });
+        const txn = await this._storage.readTxn(
+            timelineReader.readTxnStores.concat(this._storage.storeNames.roomMembers, this._storage.storeNames.roomState)
```
bwindels: the transaction should only use the `roomState` store, as that is the only store you'll be reading from here.

src/platform/web/ui/general/LazyListView.js
MidhunSureshR: Very little. It would me much simpler to implement this from scratch indeed. You can see in some places (like `loadList`) where the code gets a little crufty because we have to support the super class.

src/matrix/room/BaseRoom.js
```diff
@@ -404,11 +404,14 @@ export class BaseRoom extends EventEmitter {
     }
 
     async observePowerLevels() {
+        if (this._powerLevelLoading) { await this._powerLevelLoading; }
```
bwindels, line 424: new members should be set to null in the constructor, I think this could cause v8 to de-optimize our code if the hidden class doesn't match any more.

src/domain/session/leftpanel/LeftPanelViewModel.js
```diff
@@ -92,9 +93,10 @@ export class LeftPanelViewModel extends ViewModel {
         }
     }
 
-    _pathForDetails(path) {
-        const details = this.navigation.path.get("details");
-        return details?.value ? path.with(details) : path;
+    _pathForRightPanel(path) {
```
bwindels: why not just inline `path = addPanelIfNeeded(this.navigation, path);` where this is used?

src/platform/web/ui/session/rightpanel/RightPanelView.js
```diff
@@ -0,0 +1,34 @@
...
+
+export class RightPanelView extends TemplateView {
+    render(t) {
+        const viewFromType = {
+            "room-details": RoomDetailsView,
+            "member-list": MemberListView
+        };
+        return t.div({ className: "RightPanelView" },
+            [
+                t.mapView(vm => vm.activeViewModel && vm, vm => vm ? new ButtonsView(vm) : null),
```
bwindels: you could use `ifView` here

src/platform/web/ui/css/layout.css
```diff
@@ -208,3 +208,13 @@ the layout viewport up without resizing it when the keyboard shows */
     min-height: 0;
     overflow-y: auto;
 }
+
+.LazyListParent {
+    height: 100vh;
```
bwindels: Not sure why this doesn't make the lazy list too tall, but 100vh doesn't seem correct as there are buttons above it. This would be better:
```suggestion
    flex: 1;
```

src/platform/web/ui/css/right-panel.css
```diff
@@ -1,6 +1,13 @@
...
+.RightPanelView{
     grid-area: right;
+    min-height: 0;
+    display: flex;
+    flex-direction: column;
+}
+
+.RoomDetailsView {
     flex-direction: column;
+    height: 100%;
```
bwindels: same here
```suggestion
    flex: 1;
```

src/domain/session/rightpanel/MemberListViewModel.js
```diff
@@ -0,0 +1,46 @@
...
+        this.mediaRepository = options.mediaRepository;
+    }
+
+    get type() { return "member-list"; }
+
+    get shouldShowBackButton() { return true; }
+
+    get previousSegmentName() { return "details"; }
+
+    _filterJoinedMembers(members) {
```
bwindels: not sure it's worth to put this in a method? seems easier to read if the whole chain is in one place ...

src/platform/web/ui/session/rightpanel/MemberListView.js
MidhunSureshR: Nothing changed with lazy loading :sweat_smile:

src/platform/web/ui/general/LazyListView.js
MidhunSureshR: `listHasChangedSize` is removed but we still need an option to force render for the time being until we optimize onAdd/onRemove.

src/domain/session/rightpanel/RightPanelViewModel.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    _hookUpdaterToSegment(segment, viewmodel, argCreator) {
+        const observable = this.navigation.observe(segment);
+        const updater = this._setupUpdater(segment, viewmodel, argCreator);
+        this.track(observable.subscribe(() => updater()));
+    }
+
+    _setupUpdater(segment, viewmodel, argCreator) {
+        const updater = async (skipDispose = false) => {
```
bwindels, line 33: why do we need this flag? wouldn't only one of the options have `enable` true below?

src/domain/session/rightpanel/comparator.js
```diff
@@ -0,0 +1,97 @@
...
+
+export function createMemberComparator(powerLevels) {
+    const collator = new Intl.Collator();
+    const removeCharacter = string => string.charAt(0) === "@"? string.slice(1) : string;
```
bwindels: singling out `@` seems a bit arbitrary here, but probably fine for now.

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,251 @@
...
+        /*
+        Wait two frames for the return from mount() to be inserted into DOM.
+        This should be enough, but if this gives us trouble we can always use
+        MutationObserver.
+        */
+        await new Promise(r => requestAnimationFrame(r));
+        await new Promise(r => requestAnimationFrame(r));
+
+        this._height = this._parent.clientHeight;
+        if(this._height === 0) { throw "LazyListView could not calculate parent height."}
```
bwindels: this method should not throw as it is not awaited from `mount`, better just log?

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,251 @@
...
+        const range = this._getVisibleRange();
+        const renderRange = range.expand(this._overflowItems);
+        this._renderRange = renderRange;
+        this._renderElementsInRange();
+    }
+
+    _renderItems(items) {
```
bwindels: since this method being called from anywhere but _renderElementsInRange would break things (as _renderElementsInRange clears the DOM), should we merge the two?

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    _renderAdditionalElement(fromIdx, toIdx) {
+        const {topCount, renderCount} = this._renderRange;
+        const childFromIndex = index => this._childCreator(this._list.get(index));
+        if (toIdx < fromIdx) {
+            // Element is moved up the list, so render element from top boundary
+            const index = topCount;
+            const child = childFromIndex(index);
+            // Modify childInstances here
```
bwindels: should we do this in this PR? Sounds like it would create bugs otherwise?

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,251 @@
...
+    recreateItem(idx, value) {
+        if (this._renderRange.containsIndex(idx)) {
+            const normalizedIdx = this._renderRange.normalize(idx);
+            super.recreateItem(normalizedIdx, value)
+        }
+    }
+
+    _renderAdditionalElement(fromIdx, toIdx) {
+        const {topCount, renderCount} = this._renderRange;
+        const childFromIndex = index => this._childCreator(this._list.get(index));
```
bwindels: `BaseObservableList` doesn't define a `get` method ... should we also use an iterator here for now?

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,251 @@
...
+        if (toIdx < fromIdx) {
+            // Element is moved up the list, so render element from top boundary
+            const index = topCount;
```
bwindels, line 215: I don't understand why toIdx and fromIdx are seemingly being ignored here. If an element moves within the current rendered range, I'd think we'd need to take the specific indices into account?

src/platform/web/ui/general/LazyListView.js
bwindels: (also, a lambda for this seems overkill?)

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,251 @@
...
+    }
+    
+    recreateItem(idx, value) {
+        if (this._renderRange.containsIndex(idx)) {
+            const normalizedIdx = this._renderRange.normalize(idx);
+            super.recreateItem(normalizedIdx, value)
+        }
+    }
+
+    _renderAdditionalElement(fromIdx, toIdx) {
```
bwindels: Can we give this a name more specific to moving an item? Given it's arguments, it looks like it would only apply in that scenario.

src/platform/web/ui/general/LazyListView.js
bwindels: Oooh, I see. This is about adding an item at either edge after removing one to keep the count in the rendered range the same. Perhaps we can make this clearer?

src/platform/web/ui/session/rightpanel/LoadingView.js
bwindels: Ok, but we're not doing it in this PR I suppose? fine be me

bwindels (commented):
looking good, we're almost there!

src/domain/session/rightpanel/RightPanelViewModel.js
MidhunSureshR: We have to run the `updater` for both the segments ("memberlist" and "details") once initially because otherwise by the time the `RightPanelViewModel` is created and the segment is observed, the navigation event would have already fired.
Only one of the initial `updater` runs would actually create the `activeViewModel` but we don't want the second initial updater call to dispose of the view model created by an earlier initial updater call.

src/domain/session/rightpanel/comparator.js
MidhunSureshR: Perhaps removing the explaining comment from here just hours ago was a mistake :smile:
This is so that we can spread the members whose name is just their user id (starting with @) from clubbing together at the top of the list. By ignoring the starting @, these members are alphabetically sorted on their user id.

src/platform/web/ui/general/LazyListView.js
MidhunSureshR: Have merged the two for the time being. The follow up PR will change the render code anyway.

src/platform/web/ui/general/LazyListView.js
```diff
@@ -0,0 +1,251 @@
...
+            const index = topCount + renderCount - 1;
+            const child = childFromIndex(index);
+            this._root.appendChild(mountView(child, this._mountArgs));
+        }
+    }
+
+    _removeAdditionalElement(fromIdx, toIdx) {
+        if (toIdx < fromIdx) {
+            // Element comes from the bottom, so remove element at bottom
+            this._root.lastChild.remove();
```
MidhunSureshR: Unmount child and modify `childInstances` here.

MidhunSureshR:
~~TODO: Rebase against master to incorporate avatar fix and anything else.~~

src/platform/web/ui/general/LazyListView.js
MidhunSureshR: > `BaseObservableList` doesn't define a `get` method ... should we also use an iterator here for now?

Eventually we'd want a `get` method for `BaseObservableList` no? We sure can do iterator for now.

src/domain/session/rightpanel/MemberListViewModel.js
MidhunSureshR: Fixed for now but there are related issues. For instance if user-A and user-B are disambiguated and user-B leaves the room, we'd ideally want user-A to be un-disambiguated (which we can't really do atm because the disambiguator does not get informed when user-B leaves).

bwindels: good point, I think this is fine though for now.

src/domain/session/rightpanel/RightPanelViewModel.js
bwindels: right, makes sense now :+1:

src/platform/web/ui/general/LazyListView.js
bwindels: Yeah, probably. Just need to think if it can be implemented well in all observable list implementations (e.g. filtering changes the index).

bwindels:
Fantastic work, this is good to go!

PR #404, opened 2021-06-28
Handle avatar image error

### LINKED ISSUE #402: Avatars sometime show broken images
We should make sure that image links in avatars are valid.

MidhunSureshR:
Fixes #402 

### DIFF
--- a/src/platform/web/Platform.js
+++ b/src/platform/web/Platform.js
@@ -36,6 +36,7 @@ import {BlobHandle} from "./dom/BlobHandle.js";
 import {hasReadPixelPermission, ImageHandle, VideoHandle} from "./dom/ImageHandle.js";
 import {downloadInIframe} from "./dom/download.js";
 import {Disposables} from "../../utils/Disposables.js";
+import {handleAvatarError} from "./ui/avatar.js";
 
 function addScript(src) {
     return new Promise(function (resolve, reject) {
@@ -189,6 +190,7 @@ export class Platform {
                 this._disposables.track(disposable);
             }
         }
+        this._container.addEventListener("error", handleAvatarError, true);
         window.__hydrogenViewModel = vm;
         const view = new RootView(vm);
         this._container.appendChild(view.mount());
--- a/src/platform/web/Platform.js
+++ b/src/platform/web/Platform.js
@@ -191,6 +191,7 @@ export class Platform {
             }
         }
         this._container.addEventListener("error", handleAvatarError, true);
+        this._disposables.track(() => this._container.removeEventListener("error", handleAvatarError, true));
         window.__hydrogenViewModel = vm;
         const view = new RootView(vm);
         this._container.appendChild(view.mount());
--- a/src/platform/web/ui/AvatarView.js
+++ b/src/platform/web/ui/AvatarView.js
@@ -68,17 +68,19 @@ export class AvatarView extends BaseUpdateView {
                 this._root.replaceChild(renderImg(vm, this._size), this._root.firstChild);
                 this._root.classList.remove(bgColorClass);
             } else {
-                this._root.replaceChild(text(vm.avatarLetter), this._root.firstChild);
+                this._root.textContent = vm.avatarLetter;
                 this._root.classList.add(bgColorClass);
             }
         }
         const hasAvatar = !!vm.avatarUrl(this._size);
         if (this._avatarTitleChanged() && hasAvatar) {
-            const img = this._root.firstChild;
-            img.setAttribute("title", vm.avatarTitle);
+            const element = this._root.firstChild;
+            if (element.tagName === "IMG") {
+                element.setAttribute("title", vm.avatarTitle);
+            }
         }
         if (this._avatarLetterChanged() && !hasAvatar) {
-            this._root.firstChild.textContent = vm.avatarLetter;
+            this._root.textContent = vm.avatarLetter;
         }
     }
 }
--- a/src/platform/web/ui/AvatarView.js
+++ b/src/platform/web/ui/AvatarView.js
@@ -1,0 +1,84 @@
+import {BaseUpdateView} from "./general/BaseUpdateView.js";
+import {renderStaticAvatar, renderImg} from "./avatar.js";
+import {text} from "./general/html.js";
+
+/*
+optimization to not use a sub view when changing between img and text
+because there can be many many instances of this view
+*/
+
+export class AvatarView extends BaseUpdateView {
+    /**
+     * @param  {ViewModel} value   view model with {avatarUrl, avatarColorNumber, avatarTitle, avatarLetter}
+     * @param  {Number} size
+     */
+    constructor(value, size) {
+        super(value);
+        this._root = null;
+        this._avatarUrl = null;
+        this._avatarTitle = null;
+        this._avatarLetter = null;
+        this._size = size;
+    }
+
+    _avatarUrlChanged() {
+        if (this.value.avatarUrl(this._size) !== this._avatarUrl) {
+            this._avatarUrl = this.value.avatarUrl(this._size);
+            return true;
+        }
+        return false;
+    }
+
+    _avatarTitleChanged() {
+        if (this.value.avatarTitle !== this._avatarTitle) {
+            this._avatarTitle = this.value.avatarTitle;
+            return true;
+        }
+        return false;
+    }
+
+    _avatarLetterChanged() {
+        if (this.value.avatarLetter !== this._avatarLetter) {
+            this._avatarLetter = this.value.avatarLetter;
+            return true;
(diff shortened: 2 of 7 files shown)


src/platform/web/ui/avatar.js
```diff
@@ -65,6 +70,8 @@ export class AvatarView extends BaseUpdateView {
         this._avatarLetterChanged();
         this._avatarTitleChanged();
         this._root = renderStaticAvatar(this.value, this._size);
+        const image = this._root.firstChild;
+        image?.addEventListener("error", () => this.setAvatarError());
```
bwindels: Perhaps also listen for load here and in both handlers, remove the listeners of both, so once an image is loaded or failed to load, there are no listeners anymore.

src/platform/web/ui/avatar.js
```diff
@@ -65,6 +70,8 @@ export class AvatarView extends BaseUpdateView {
         this._avatarLetterChanged();
         this._avatarTitleChanged();
         this._root = renderStaticAvatar(this.value, this._size);
+        const image = this._root.firstChild;
```
bwindels: Can we also do this for static avatars (e.g. done with renderStaticAvatar and not binding to the view model)? We'll need to handle both separately though I think, as we need to know in update if the avatar failed to load to add or remove the background color class.

bwindels (commented):
Thanks! Looks mainly good, just a few small nits.

src/platform/web/ui/avatar.js
```diff
@@ -57,6 +57,20 @@ export class AvatarView extends BaseUpdateView {
...
+        const handleAvatarError = (e) => {
+            const image = e.target;
+            image.removeEventListener("load", removeErrorHandler);
+            this._setAvatarError();
+        };
+        const removeErrorHandler = (e) => {
+            const image = e.target;
+            image.removeEventListener("error", handleAvatarError);
+        };
+        image?.addEventListener("error", handleAvatarError);
```
bwindels: don't you need to remove both handlers on either event? I think you can use the same handler for both events even, and set `this._avatarError` to `e.type === "error";`.

MidhunSureshR: Oops! Sorry about that

src/platform/web/Platform.js
```diff
@@ -189,6 +190,7 @@ export class Platform {
                 this._disposables.track(disposable);
             }
         }
+        this._container.addEventListener("error", handleAvatarError, true);
```
bwindels, line 193: Can you add a closure to disposables to remove this again when dispose gets called? `adaptUIOnVisualViewportResize` has an example of this.

src/platform/web/ui/avatar.js
```diff
@@ -108,16 +25,29 @@ export function renderStaticAvatar(vm, size, extraClasses = undefined) {
...
         [`size-${size}`]: true,
-        [`usercolor${vm.avatarColorNumber}`]: !hasAvatar,
+        [`usercolor${vm.avatarColorNumber}`]: true
     });
     if (extraClasses) {
         avatarClasses += ` ${extraClasses}`;
     }
     const avatarContent = hasAvatar ? renderImg(vm, size) : text(vm.avatarLetter);
-    return tag.div({className: avatarClasses}, [avatarContent]);
+    return tag.div({className: avatarClasses, 'data-avatar-letter': vm.avatarLetter}, [avatarContent]);
```
bwindels: You could only set the letter if `hasAvatar` is true

src/platform/web/ui/avatar.js
```diff
@@ -108,16 +25,29 @@ export function renderStaticAvatar(vm, size, extraClasses = undefined) {
...
+    const element = e.target;
+    const parent = element.parentElement;
+    return element.tagName === "IMG" && parent.classList.contains("avatar");
+}
+
+export function handleAvatarError(e) {
+    if (!isAvatarEvent(e)) { return; }
+    const parent = e.target.parentElement;
+    const avatarLetter = parent.getAttribute("data-avatar-letter");
+    parent.textContent = avatarLetter;
```
bwindels, line 59: Wouldn't you also need to hide the img somehow? (and unhide if the binding in AvatarView changes the `src` property)

src/platform/web/ui/AvatarView.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    update(vm) {
+        // important to always call _...changed for every prop 
+        if (this._avatarUrlChanged()) {
+            // avatarColorNumber won't change, it's based on room/user id
+            const bgColorClass = `usercolor${vm.avatarColorNumber}`;
+            if (vm.avatarUrl(this._size)) {
+                this._root.replaceChild(renderImg(vm, this._size), this._root.firstChild);
+                this._root.classList.remove(bgColorClass);
```
bwindels, line 69: The background color class should not be removed or set here I imagine, but just always left as it was by the initial render in `renderStaticAvatar`. You would however need to add/remove a class to show or hide the image when it fails to load I think?

bwindels (commented):
Looking good, just a few more things.

src/platform/web/ui/avatar.js
bwindels: textContent already clears the image

src/platform/web/ui/AvatarView.js
```diff
@@ -0,0 +1,84 @@
...
+                this._root.classList.add(bgColorClass);
+            }
+        }
+        const hasAvatar = !!vm.avatarUrl(this._size);
+        if (this._avatarTitleChanged() && hasAvatar) {
+            const img = this._root.firstChild;
+            img.setAttribute("title", vm.avatarTitle);
+        }
+        if (this._avatarLetterChanged() && !hasAvatar) {
+            this._root.firstChild.textContent = vm.avatarLetter;
```
bwindels: Can you do `this._root.textContent = ...` here as well like you did in handleAvatarError?

src/platform/web/ui/AvatarView.js
```diff
@@ -0,0 +1,84 @@
...
+                this._root.replaceChild(renderImg(vm, this._size), this._root.firstChild);
+                this._root.classList.remove(bgColorClass);
+            } else {
+                this._root.replaceChild(text(vm.avatarLetter), this._root.firstChild);
+                this._root.classList.add(bgColorClass);
+            }
+        }
+        const hasAvatar = !!vm.avatarUrl(this._size);
+        if (this._avatarTitleChanged() && hasAvatar) {
+            const img = this._root.firstChild;
```
bwindels: Can you check here now if this is indeed an image? As we might have removed it on error ...

src/platform/web/ui/AvatarView.js
```diff
@@ -0,0 +1,84 @@
...
+    update(vm) {
+        // important to always call _...changed for every prop 
+        if (this._avatarUrlChanged()) {
+            // avatarColorNumber won't change, it's based on room/user id
+            const bgColorClass = `usercolor${vm.avatarColorNumber}`;
+            if (vm.avatarUrl(this._size)) {
+                this._root.replaceChild(renderImg(vm, this._size), this._root.firstChild);
+                this._root.classList.remove(bgColorClass);
+            } else {
+                this._root.replaceChild(text(vm.avatarLetter), this._root.firstChild);
```
bwindels: Can you do `this._root.textContent = ...` here as well like you did in handleAvatarError?

src/platform/web/ui/avatar.js
```diff
@@ -108,16 +25,36 @@ export function renderStaticAvatar(vm, size, extraClasses = undefined) {
...
+    const parent = element.parentElement;
+    return element.tagName === "IMG" && parent.classList.contains("avatar");
+}
+
+export function handleAvatarError(e) {
+    if (!isAvatarEvent(e)) { return; }
+    const parent = e.target.parentElement;
+    const avatarColorNumber = parent.getAttribute("data-avatar-color");
+    parent.classList.add(`usercolor${avatarColorNumber}`);
+    const avatarLetter = parent.getAttribute("data-avatar-letter");
```
bwindels, line 58: actually, looks like you want to use innerText to make the  go away in IE11
https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext

bwindels (approved):
Looks good, thanks! Did you test that it works as well?

MidhunSureshR:
buggy
Before
memberlist
Works with memberlist :heavy_check_mark: 
timeline
Works with timeline :heavy_check_mark: 
transparent
Adds white background by default :heavy_check_mark:
Is that good enough :thinking: or do we need to test with a dummy view with invalid avatars?

bwindels:
That's good enough for testing for me! Let's merge it!

PR #405, opened 2021-06-28
Ensure room color in DM, timeline, and left panel is the same.

### LINKED ISSUE #329: Default avatar for DM has different color in room list and room header than in timeline

DanilaFe:
I believe this doesn't work with named DMs, since it the `_heroes` property is not present in that case. I'm not sure what the intended behavior would be in this situation.

### DIFF
--- a/src/domain/session/leftpanel/RoomTileViewModel.js
+++ b/src/domain/session/leftpanel/RoomTileViewModel.js
@@ -15,6 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import {getIdentifierColorNumber} from "../../avatar.js";
 import {BaseTileViewModel} from "./BaseTileViewModel.js";
 
 export class RoomTileViewModel extends BaseTileViewModel {
@@ -73,6 +74,10 @@ export class RoomTileViewModel extends BaseTileViewModel {
             return nameCmp;
         }
         return timeDiff;
+    }
+
+    get avatarColorNumber() {
+        return getIdentifierColorNumber(this._room.avatarColorId);
     }
 
     get isUnread() {
--- a/src/domain/session/leftpanel/RoomTileViewModel.js
+++ b/src/domain/session/leftpanel/RoomTileViewModel.js
@@ -15,7 +15,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {getIdentifierColorNumber} from "../../avatar.js";
 import {BaseTileViewModel} from "./BaseTileViewModel.js";
 
 export class RoomTileViewModel extends BaseTileViewModel {
@@ -74,10 +73,6 @@ export class RoomTileViewModel extends BaseTileViewModel {
             return nameCmp;
         }
         return timeDiff;
-    }
-
-    get avatarColorNumber() {
-        return getIdentifierColorNumber(this._room.avatarColorId);
     }
 
     get isUnread() {
--- a/src/matrix/room/RoomSummary.js
+++ b/src/matrix/room/RoomSummary.js
@@ -239,15 +239,6 @@ export class SummaryData {
         return !this.name && !this.canonicalAlias && this.heroes && this.heroes.length > 0;
     }
 
-    get avatarColorId() {
-        if (this.heroes && this.heroes.length === 1) {
-            for (const hero of this.heroes) {
-                return hero;
-            }
-        }
-        return this.roomId;
-    }
-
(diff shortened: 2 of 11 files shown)


bwindels (commented):
This is looking very nice, thanks! There is a slight complication though:

In the room list, you can see 3 kinds of rooms: joined rooms (Room, extending BaseRoom), left or archived rooms (ArchivedRoom extending BaseRoom) and invites (Invite, just extending EventEmitter).

Ideally, the avatar color would not change as you move through these different memberships. The way a joined room is handled looks great to me, it would be nice if we could also handle Invite and ArchivedRoom as well.

For an Invite, you'll need to store the avatarColorId of the heroes in Invite._createData and just create a getter for it.

For ArchivedRoom, the actual heroe userids will be stored in the `joinedSummaryData` (an instance of RoomSummary) that is adopted from the `Room` when you leave the room. With just those user ids, you should be able to implement avatarColorId n the same fashion as you already did in `Heroes`.

Once all 3 of these room classes have a avatarColorId property, you can move the avatarColorNumber from RoomTileViewModel to the parent class, BaseTileViewModel, and make it work for both InviteTileViewModel and RoomTileViewModel (used for both joined and archived rooms).

DanilaFe:
~I still have a few questions:~

1. ~The `_summary` property is present all the way back to `BaseRoom`. Is there a source of truth, or do they have to be identical? If the `_summary` is valid for `BaseRoom` and `Room`, perhaps the cleanest move is to make `avatarColorId` on `BaseRoom` consult `RoomSummary`, and be done.~
2. ~What about `Invite`, which receives both `summaryData` (which contains `heroes`) and `heroes`? Is there some precedence here to which should be used to decide the `avatarColorId`, or are they always the same?~

__Edit__: To answer my own question, it seems like `summary`'s `heroes` are just the hero IDs, while `BaseRoom._heroes` contains other data. We don't really need the other data, so we may as well use the summary (which appears always up-to-date, and present in all rooms).  I've implemented this in d0f70cbd.

bwindels:
Fixes https://github.com/vector-im/hydrogen-web/issues/329

src/matrix/room/RoomSummary.js
```diff
@@ -239,6 +239,15 @@ export class SummaryData {
         return !this.name && !this.canonicalAlias && this.heroes && this.heroes.length > 0;
     }
 
+    get avatarColorId() {
+        if (this.heroes && this.heroes.length === 1) {
+            for (const hero of this.heroes) {
```
bwindels: you can just return this.heroes[0] here, as it is an array and not just a mere iterable.

src/domain/session/leftpanel/RoomTileViewModel.js
```diff
@@ -75,6 +76,10 @@ export class RoomTileViewModel extends BaseTileViewModel {
         return timeDiff;
     }
 
+    get avatarColorNumber() {
```
bwindels: You can put this in the parent class, `BaseTileViewModel`, which already has this getter. There, this._avatarSource is either an Invite or a subclass of BaseRoom, so `this._avatarSource.avatarColorId` will handle all cases.

src/matrix/room/RoomSummary.js
bwindels: can you also add a jsdoc comment here with a description of what this property is for, please?

bwindels (commented):
Looking good, thanks! Just a few more little things.

bwindels:
> I believe this doesn't work with named DMs, since it the `_heroes` property is not present in that case. I'm not sure what the intended behavior would be in this situation.

That's fine btw. In most cases (apart from the IRC bridge you mentioned) the room would not be named after the other person in it, and it wouldn't not be confusing to have that room avatar be a different color than the person in the timeline.

src/matrix/room/RoomSummary.js
```diff
@@ -239,6 +239,15 @@ export class SummaryData {
         return !this.name && !this.canonicalAlias && this.heroes && this.heroes.length > 0;
     }
 
+    get avatarColorId() {
```
bwindels: Actually, I think your first idea to handle this in `Heroes` might be the right one. As you can see in `Invite.writeSync`, the heroes are derived from the invite state, as room invites never contain the `m.summary` section. So `this.heroes` will never be an array for an invite in the summary data. Given we already do the work of deriving the `Heroes`, better reuse that and put this property in there I think.

bwindels (approved):
Looks great, thanks!

PR #407, opened 2021-07-01
Formatted messages

DanilaFe:
Just  getting started on this. See [FORMATTED](https://github.com/vector-im/hydrogen-web/blob/677bd2984b74aef45ce6151fd135035dc66b2e6b/src/domain/session/room/timeline/FORMATTED.md) for my implementation ideas.

### DIFF
--- a/src/domain/session/room/timeline/FORMATTED.md
+++ b/src/domain/session/room/timeline/FORMATTED.md
@@ -1,0 +1,19 @@
+# Ideas for formatted messages
+
+* Seems like a good idea to take some
+  inspiration from [Pandoc's AST](https://hackage.haskell.org/package/pandoc-types-1.22/docs/Text-Pandoc-Definition.html#t:Block).
+  Then, much like Pandoc AST, we can turn our representation into
+  markdown in the editor, or HTML in the view.
+* As such, we represent formatting by nesting parts
+  (we'd have a `ItalicsPart`, `BoldPart`, etc.)
+* We keep the "inline"/"block" distinction, but only
+  track it as a property, so that we can avoid adding
+  block parts to elements that cannot contain blocks
+  (headers, for instance, cannot contain blocks, but
+  lists -- themselves blocks -- can).
+* When parsing, we may need some sort of "permanent" context:
+  if we're parsing a header, and we are inside 3 layers of other
+  "inline" things (italics, strikethrough, and bold, for example),
+  and we encounter a block, that's still not valid.
+  * Element seems to not care at all about the validity of what
+    it's parsing. Do we assume the HTML is well-formatted, then?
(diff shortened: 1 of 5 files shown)


src/domain/session/room/timeline/FORMATTED.md
```diff
@@ -0,0 +1,19 @@
...
+  Then, much like Pandoc AST, we can turn our representation into
+  markdown in the editor, or HTML in the view.
+* As such, we represent formatting by nesting parts
+  (we'd have a `ItalicsPart`, `BoldPart`, etc.)
+* We keep the "inline"/"block" distinction, but only
+  track it as a property, so that we can avoid adding
+  block parts to elements that cannot contain blocks
+  (headers, for instance, cannot contain blocks, but
+  lists -- themselves blocks -- can).
+* When parsing, we may need some sort of "permanent" context:
```
bwindels, line 20: Indeed, that's how I did this in element web fwiw: https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/editor/deserialize.ts#L94-L99

src/platform/web/ui/session/room/timeline/TextMessageView.js
```diff
@@ -36,18 +36,29 @@ export class TextMessageView extends BaseMessageView {
  * Map from part to function that outputs DOM for the part
  */
 const formatFunction = {
+    header: headerBlock => tag["h" + Math.min(6,headerBlock.level)]({}, renderParts(headerBlock.inlines)),
```
bwindels, line 39: Should the min be on the model, or is your reasoning that it's a HTML limitation to only allow 6 nested headers?

src/domain/session/room/timeline/FORMATTED.md
```diff
@@ -0,0 +1,19 @@
...
+  if we're parsing a header, and we are inside 3 layers of other
+  "inline" things (italics, strikethrough, and bold, for example),
+  and we encounter a block, that's still not valid.
+  * Element seems to not care at all about the validity of what
+    it's parsing. Do we assume the HTML is well-formatted, then?
```
bwindels, line 25: I would think that `DOMParser` will try to wrangle the html into a valid DOM, but could be mistaken.

src/platform/web/ui/session/room/timeline/TextMessageView.js
DanilaFe: Indeed, my assumption was that 6 is an HTML limitation. I think that there's no real reason to enforce the magic number 6 on our "format-independent" data (although we'd probably be hard-pressed to find a source format that does use more than 6 headers)

src/domain/session/room/timeline/FORMATTED.md
DanilaFe: It does, and so we need not check for validity.

DanilaFe:
Something's up with this pull request; it's showing classes that have been deleted, and duplicate commit. I'm going to close and re-open. Sorry for the noise.

PR #408, opened 2021-07-02
Formatted messages

### LINKED ISSUE #406: Display formatted messages correctly

DanilaFe:
Re-opening of #407, which is misbehaving.


src/domain/session/room/timeline/tiles/TextTile.js
```diff
@@ -16,18 +16,39 @@ limitations under the License.
...
+        let val = content?.[key] || fallback;
+        if (!val && val !== "") { // empty string is falsy, but OK here.
+            return null;
+        }
         if (content.msgtype === "m.emote") {
-            body = `* ${this.displayName} ${body}`;
+            val = `* ${this.displayName} ${body}`;
         }
-        return body;
+        return { string: val, format };
```
bwindels: Perhaps add another method here, `_getBodyFormat`, rather than returning an object?

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,285 @@
...
+
+function parseImage(options, node) {
+    const { result, mediaRepository } = options;
+    const src = result.getAttributeValue(node, "src") || "";
+    const url = mediaRepository.mxcUrl(src);
+    // We just ignore non-mxc `src` attributes.
+    if (!url) {
+        return null;
+    }
+    const width = result.getAttributeValue(node, "width");
```
bwindels: probably best to parse this to an int to validate the value somewhat

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,285 @@
...
+    const height = result.getAttributeValue(node, "height");
+    const alt = result.getAttributeValue(node, "alt");
+    const title = result.getAttributeValue(node, "title");
+    return new ImagePart(url, { width, height, alt, title });
```
bwindels: curious why you put these in a separate object rather than just put them on ImagePart as individual properties?

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,285 @@
...
+            const input = '<ul start="3"><li>Lorem</li><li>Ipsum</li></ul>';
+            const output = [
+                new ListBlock(null, [
+                    [ new TextPart("Lorem") ],
+                    [ new TextPart("Ipsum") ]
+                ])
+            ];
+            test(assert, input, output);
+        },
+        /* Doesnt work: HTML library doesn't handle <pre><code> properly.
```
bwindels, line 490: node-html-parser? Do you know what it doesn't like about it?

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,285 @@
...
+function parseNode(options, node) {
+    const { result } = options;
+    if (result.isTextNode(node)) {
+        return new TextPart(result.getNodeText(node));
+    } else if (result.isElementNode(node)) {
+        const f = nodes[result.getNodeElementName(node)];
+        if (!f) {
+            return null;
+        }
+        const children = f.descend ? parseNodes(options, node.childNodes) : null;
```
bwindels: I think this might have slipped the opaque handle approach:
```suggestion
        const children = f.descend ? parseNodes(options, result.getChildNodes(node)) : null;
```

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,285 @@
...
+ * parsed just like any other node, and fed as a second argument
+ * to `parsefn`. If not, the node's children are either to be ignored
+ * (as in <pre>) or processed specially (as in <ul>).
+ *
+ * The `parsefn` combines a node's data and its children into
+ * an internal representation node.
+ */
+const nodes = buildNodeMap();
+
+function parseNode(options, node) {
```
bwindels: you could consider putting all of these parse functions in a class so you don't have to pass the options as an argument everywhere but just have `this.parseResult...`

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,285 @@
...
+    const { result } = options;
+    if (result.isTextNode(node)) {
+        return new TextPart(result.getNodeText(node));
+    } else if (result.isElementNode(node)) {
+        const f = nodes[result.getNodeElementName(node)];
+        if (!f) {
+            return null;
+        }
+        const children = f.descend ? parseNodes(options, node.childNodes) : null;
+        return f.parsefn(options, node, children);
```
bwindels: for some parts, it could make sense to only allow a set of child parts, like a list only containing list items. I guess you could filter them in the parse function, but if you get the children and iterate them in the parse function, you can filter before you parse. some for not allowing block elements for formatrule.

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,285 @@
...
+export function parseHTMLBody({ mediaRepository, platform }, html) {
+    const parseResult = platform.parseHTML(sanitizeHtml(html, sanitizeConfig));
+    const options = { result: parseResult, mediaRepository };
+    const parts = parseNodes(options, parseResult.rootNodes);
+    return new MessageBody(html, parts);
+}
+
+import parse from '../../../../../lib/node-html-parser/index.js';
+
+export class HTMLParseResult {
```
bwindels: if this is only used in the test, better put it in the `tests` function

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,285 @@
...
+        "a": ["name", "target", "href"],
+        "img": ["width", "height", "alt", "title", "src"],
+        "ol": ["start"],
+        "code": ["class"]
+    },
+    allowedSchemes: [ "http", "https", "ftp", "mailto", "tel", "mxc" ]
+};
+
+export function parseHTMLBody({ mediaRepository, platform }, html) {
+    const parseResult = platform.parseHTML(sanitizeHtml(html, sanitizeConfig));
```
bwindels: Perhaps the sanitization step can be done within platform.parseHTML, so we can use platform specific libraries (like dompurify). I don't think we need such a specific configuration either, as we already filter stuff with our parsing code afterwards. It's mainly meant to catch malicious attributes where we have a bug in the parsing of the attributes in our own parsing code.

bwindels (commented):
Looks great, and nearly there it looks. Have mainly stylistic concerns.

src/domain/session/room/timeline/deserialize.js
DanilaFe: :facepalm:

src/domain/session/room/timeline/deserialize.js
DanilaFe: the `code` element becomes text. Thus, we end up with `<code>3+x</code>` instead of `3+x`. This make `parseCode` fail, because it expects a `code` element, and not a text node.

src/domain/session/room/timeline/deserialize.js
DanilaFe: I've switched to using `DOMPurify` in the platform layer. I was wondering, though: doesn't this offload more work onto the platform implementation? Unless, of course, you have in mind not sanitizing anything at all on non-web platforms.

src/domain/session/room/timeline/deserialize.js
DanilaFe: Here, `descend` was used to indicate nodes that have children the parsing of which does not depend on the parent (for example, children of `p` tags don't care that they're in a `p` tag). For nodes that are more particular about their children (and might reject them), `descend` is set to false, `parseNodes` is never called, and the iteration occurs in the parse function.

Either way, this is now out of date; in refactoring the deserializer to be a class, I've moved  to using `switch` statements on the node tag, and explicitly parsing children only when necessary. 

Also, I'm currently operating under the assumption that the DOMParser "rearranges" ill-formed HTML (for instance, it will auto-close `p` tags if it encounters another block element inside it). This seems to hold true with the Node HTML parser, too (I will add a test). Thus, there's no need to worry about block elements where they don't belong.

src/domain/session/room/timeline/deserialize.js
bwindels: right, it has special handling to treat everything in `<pre>` as character data... not sure if that's what the HTML spec says, but :man_shrugging:

src/domain/session/room/timeline/deserialize.js
```diff
@@ -13,140 +13,137 @@ import sanitizeHtml from "../../../../../lib/sanitize-html/index.js"
...
+    }
 
-function parseList(options, node) {
-    const { result } = options;
-    let start = null;
-    if (result.getNodeElementName(node) === "OL") {
-        // Will return 1 for, say, '1A', which may not be intended?
-        start = parseInt(result.getAttributeValue(node, "start")) || 1;
+    parseLink(node, children) {
+        // TODO Not equivalent to `node.href`!
```
bwindels: What's the difference between href and getAttribute(href)?

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,270 @@
...
+                const children = this.parseChildNodes(node);
+                return new FormatPart(tag, children);
+            }
+        }
+    }
+
+    parseNode(node) {
+        const result = this.result;
+        if (result.isTextNode(node)) {
+            return new TextPart(result.getNodeText(node));
```
bwindels: We should still parse links in the text, like we do for plain text messages.

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,270 @@
...
+        if (result.getNodeElementName(node) === "OL") {
+            // Will return 1 for, say, '1A', which may not be intended?
+            start = parseInt(result.getAttributeValue(node, "start")) || 1;
+        }
+        const nodes = [];
+        for (const child of result.getChildNodes(node)) {
+            if (result.getNodeElementName(child) !== "LI") {
+                continue;
+            }
+            const item = this.parseNodes(result.getChildNodes(child));
```
bwindels: this will be an array of parts, rather than a single part or item?

DanilaFe: This is intentional. A "list item" is a list of nodes (for instance, a list of paragraphs within the list). The Pandoc equivalent is `BulletList [[Block]]`.

src/domain/session/room/timeline/deserialize.js
DanilaFe: `href` returns the computed destination URL, whereas `getAttribute('href')` returns the literal contents of the `href` attribute. Element uses `href`, I believe.

bwindels: I see. We probably should not allow relative URLs at all, so not really an issue I suppose?

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,398 @@
...
+
+class Deserializer {
+    constructor(result, mediaRepository) {
+        this.result = result;
+        this.mediaRepository = mediaRepository;
+    }
+
+    parseLink(node, children) {
+        const href = this.result.getAttributeValue(node, "href");
+        if (!href || !href.match(/^[a-z]+:[\/]{2}/i)) {
```
DanilaFe: Relative URLs are thrown away here. Of note, though, is that Element allows relative URLs.

src/platform/web/ui/session/room/timeline/TextMessageView.js
```diff
@@ -32,22 +32,86 @@ export class TextMessageView extends BaseMessageView {
...
+    return f(part);
+}
+
+function renderParts(parts) {
+    return Array.from(parts, renderPart);
+}
+
 class BodyView extends StaticView {
     render(t, messageBody) {
         const container = t.span();
```
DanilaFe: As things stand, all the block elements end up inside the span. That doesn't seem correct (we just recently talked about something like this in the intermediate representation), and the subsequent timestamp looks a bit weird.

src/domain/session/room/timeline/deserialize.js
DanilaFe: We now turn relative links into plain spans (see below comment).

scripts/post-install.js
```diff
@@ -76,6 +77,18 @@ async function populateLib() {
     for (const file of ["olm.js", "olm.wasm", "olm_legacy.js"]) {
         await fs.symlink(path.join(olmSrcDir, file), path.join(olmDstDir, file));
     }
+    // transpile node-html-parser to esm
+    await fs.mkdir(path.join(libDir, "node-html-parser/"));
+    await commonjsToESM(
```
bwindels, line 82: I guess the `rollup/plugin-json` is needed because node-html-parser requires json somewhere?

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,457 @@
...
+
+    parseCodeBlock(node) {
+        const result = this.result;
+        let codeNode;
+        for (const child of result.getChildNodes(node)) {
+            codeNode = child;
+            break;
+        }
+        if (!this._ensureElement(codeNode, "CODE")) {
+            return null;
```
bwindels: we should probably do something with just `<pre>` tags too ... We could consider the `<code>` tag optional, e.g. to set the language.

src/domain/session/room/timeline/deserialize.js
```diff
@@ -0,0 +1,457 @@
...
+            // Will return 1 for, say, '1A', which may not be intended?
+            start = parseInt(result.getAttributeValue(node, "start")) || 1;
+        }
+        const nodes = [];
+        for (const child of result.getChildNodes(node)) {
+            if (result.getNodeElementName(child) !== "LI") {
+                continue;
+            }
+            const item = this.parseAnyNodes(result.getChildNodes(child));
+            nodes.push(item);
```
bwindels: items rather than nodes?

src/domain/session/room/timeline/pills.js
```diff
@@ -0,0 +1,12 @@
+const baseUrl = 'https://matrix.to';
+const linkPrefix = `${baseUrl}/#/`;
+
+export function parsePillLink(link) {
+    if (!link.startsWith(linkPrefix)) {
+        return null;
+    }
+    const contents = link.substring(linkPrefix.length);
+    if (contents[0] === '@') {
+        return { userId: contents }
```
bwindels: I see why you extracted this to its own module (to separate pill handling from message parsing), but pills only really appear in messages, so wonder if it's worth it. (and also this allocates another temp object :)

src/platform/web/ui/session/room/timeline/TextMessageView.js
```diff
@@ -14,41 +14,101 @@ See the License for the specific language governing permissions and
...
+    }
+}
+
+function renderList(listBlock) {
+    const items = listBlock.items.map(item => tag.li({}, renderParts(item)));
+    const start = listBlock.startOffset;
+    if (start) {
+        return tag.ol({ start }, items);
+    } else {
+        return tag.ul({}, items);
```
bwindels: you can just do tag.ul(items) fwiw

src/platform/web/ui/session/room/timeline/TextMessageView.js
```diff
@@ -14,41 +14,101 @@ See the License for the specific language governing permissions and
...
+    }
+}
+
+function renderImage(imagePart) {
+    const attributes = Object.assign(
+        { src: imagePart.src },
```
bwindels: this allocates 5 objects ... perhaps there's a simpler way?

src/platform/web/ui/session/room/timeline/TextMessageView.js
```diff
@@ -14,41 +14,101 @@ See the License for the specific language governing permissions and
...
+        imagePart.width && { width: imagePart.width },
+        imagePart.height && { height: imagePart.height },
+        imagePart.alt && { alt: imagePart.alt },
+        imagePart.title && { title: imagePart.title }
+    );
+    return tag.img(attributes, []);
```
bwindels: you can just do tag.img(attributes) fwiw

src/domain/session/room/timeline/pills.js
```diff
@@ -0,0 +1,12 @@
+const baseUrl = 'https://matrix.to';
```
bwindels: Can you add license headers for the matrix foundation to all the new files you added please?

src/domain/session/room/timeline/tiles/TextTile.js
```diff
@@ -16,18 +16,42 @@ limitations under the License.
...
     }
 
-    _parseBody(bodyString) {
-        return parsePlainBody(bodyString);
+    _getBody() {
+        return this._getFormattedBody() || this._getPlainBody();
+    }
+
+    _getBodyFormat() {
+        return this._getContent()?.["formatted_body"] ? "html" : "plain"
```
bwindels: we should probably check `content.format === "org.matrix.custom.html"`  here?

src/domain/session/room/timeline/tiles/BaseTextTile.js
```diff
@@ -21,26 +21,35 @@ export class BaseTextTile extends BaseMessageTile {
...
+        // Even if the body hasn't changed, but the format has, we need
+        // to re-fill our cache.
+        if (!this._messageBody || this._messageBody.sourceString !== body || this._messageFormat !== format) {
             // body with markup is an array of parts,
             // so we should not recreate it for the same body string,
             // or else the equality check in the binding will always fail.
             // So cache it here.
-            this._messageBody = this._parseBody(body);
+            this._messageBody = this._parseBody(body, format);
+            this._messageFormat = body.format;
```
bwindels: should just be format?

src/domain/session/room/timeline/tiles/EncryptedEventTile.js
```diff
@@ -33,13 +33,15 @@ export class EncryptedEventTile extends BaseTextTile {
...
         const code = decryptionError?.code;
+        let string;
         if (code === "MEGOLM_NO_SESSION") {
-            return this.i18n`The sender hasn't sent us the key for this message yet.`;
+            string = this.i18n`The sender hasn't sent us the key for this message yet.`;
         } else {
-            return decryptionError?.message || this.i18n`Could not decrypt message because of unknown reason.`;
+            string = decryptionError?.message || this.i18n`Could not decrypt message because of unknown reason.`;
         }
+        return { string, format: "plain" };
```
bwindels: need to still split this up on getBody and getBodyFormat?

src/domain/session/room/timeline/tiles/TextTile.js
```diff
@@ -16,18 +16,42 @@ limitations under the License.
...
+    }
+
+    _parseBody(body, format) {
+        if (format === "html") {
```
bwindels: Perhaps we can even use an enum for the format? see src/utils/enum.js ...

src/domain/session/room/timeline/tiles/TextTile.js
```diff
@@ -16,18 +16,42 @@ limitations under the License.
...
+    }
+
+    _getFormattedBody() {
+        return this._getContentString("formatted_body");
     }
 
-    _parseBody(bodyString) {
-        return parsePlainBody(bodyString);
+    _getBody() {
+        return this._getFormattedBody() || this._getPlainBody();
```
bwindels: can we check the format here rather than the presence of a property in contents?

src/platform/web/ui/session/room/timeline/TextMessageView.js
```diff
@@ -14,41 +14,101 @@ See the License for the specific language governing permissions and
...
+    children.push(tag.tbody({}, rows));
+    return tag.table({}, children);
 }
 
 /**
  * Map from part to function that outputs DOM for the part
  */
 const formatFunction = {
+    header: headerBlock => tag["h" + Math.min(6,headerBlock.level)]({}, renderParts(headerBlock.inlines)),
+    codeblock: codeBlock => tag.pre({}, tag.code({}, text(codeBlock.text))),
```
bwindels: same here, attributes are optional

bwindels (commented):
looking great, nearly there!

scripts/post-install.js
DanilaFe: It does, yes. I think there's a JSON list of allow nodes or something of that sort.

src/platform/web/ui/session/room/timeline/TextMessageView.js
bwindels: Have fixed this by using mapSideEffect, if we ever decide to put the time elsewhere we can get rid of this hack-ish thing.

bwindels (approved):
Thanks, excellent work, this makes it feel so much more grown up!

PR #410, opened 2021-07-17
Fix error on invites in room list panel.

DanilaFe:
Frustratingly, the null-containing invites are saved to storage and will not be fixed by this.


src/matrix/room/Invite.js
```diff
@@ -58,7 +58,7 @@ export class Invite extends EventEmitter {
 
     /** @see BaseRoom.avatarColorId */
     get avatarColorId() {
-        return this._inviteData.avatarColorId;
+        return this._inviteData.avatarColorId || this.id;
```
DanilaFe, line 61: Invites are only process once, when they are received, so if a client inserted `null` into storage, that `null` stays there. This line could be used to mitigate the issue.

bwindels (approved):
Looks good! Thanks for the fix!

PR #412, opened 2021-07-19
Member panel - PR 1 - Add method to fetch member

MidhunSureshR:

### DIFF
--- a/src/matrix/room/BaseRoom.js
+++ b/src/matrix/room/BaseRoom.js
@@ -21,7 +21,7 @@ import {RelationWriter} from "./timeline/persistence/RelationWriter.js";
 import {Timeline} from "./timeline/Timeline.js";
 import {FragmentIdComparer} from "./timeline/FragmentIdComparer.js";
 import {WrappedError} from "../error.js"
-import {fetchOrLoadMembers, loadMember} from "./members/load.js";
+import {fetchOrLoadMembers, fetchOrLoadMember} from "./members/load.js";
 import {MemberList} from "./members/MemberList.js";
 import {Heroes} from "./members/Heroes.js";
 import {EventEntry} from "./timeline/entries/EventEntry.js";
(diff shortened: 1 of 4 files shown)


src/matrix/room/members/load.js
```diff
@@ -90,3 +90,9 @@ export async function fetchOrLoadMembers(options, logger) {
         return loadMembers(options);
     }
 }
+
+export async function loadMember({roomId, userId, storage}) {
```
MidhunSureshR: We need a `fetchOrLoadMember` so that we can  show the details of a member who has not yet been fetched (roomSummary.hasFetchedMembers = false; since memberlist has not been loaded yet).  
We could do this using the [/profile](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-profile-userid) endpoint.

This is only a problem when switching to another room which has not fetched the memberlist yet but does contain the member.

bwindels: Yeah, I think the best approach here would be:

1. try read the member
2. if found, return
2. if not found, fetch it with https://matrix.org/docs/spec/client_server/latest#get-matrix-client-r0-rooms-roomid-state-eventtype-statekey
3. store the member
4. return

The idea being that if you've opened the member panel before, it will be available offline the next time you open it.

src/matrix/room/BaseRoom.js
```diff
@@ -214,6 +215,26 @@ export class BaseRoom extends EventEmitter {
...
+        }
+        const mapMember = this._observedMembers.get(userId);
+        if (mapMember) {
+            // Hit, we're already observing this member
+            return mapMember;
+        }
+        // Miss, load from storage and set in map
+        const member = await loadMember({roomId: this._roomId, userId, storage: this._storage});
+        if (!member) {
+            return false;
```
bwindels: I'd return null here, so the return type is always the "same"

bwindels (commented):
this looks excellent, including updates from sync, nice! I would add fetching too while at it though.

src/matrix/room/members/load.js
```diff
@@ -90,3 +90,49 @@ export async function fetchOrLoadMembers(options, logger) {
...
+        // We haven't fetched the memberlist yet; so ping the hs to see if this member does exist
+        return logger.wrapOrRun(options.log, "fetchMember", log => fetchMember(options, log));
+    }
+    return member;
+}
+
+async function loadMember({roomId, userId, storage}) {
+    const txn = await storage.readTxn([storage.storeNames.roomMembers,]);
+    const member = await txn.roomMembers.get(roomId, userId);
+    return member? new RoomMember(member) : undefined;
```
bwindels: perhaps return null here, which is usually for "not found"

bwindels (approved):

PR #417, opened 2021-07-20
Member Panel - PR 2 - UI

MidhunSureshR:


src/platform/web/ui/session/rightpanel/MemberDetailsView.js
```diff
@@ -0,0 +1,54 @@
...
+import {TemplateView} from "../../general/TemplateView.js";
+
+export class MemberDetailsView extends TemplateView {
+    render(t, vm) {
+        const encryptionMessages = {
+            true: vm.i18n`Messages in this room are end-to-end encrypted.`,
+            false: vm.i18n`Messages in this room are not end-to-end encrypted.`
+        };
+        return t.div({className: "MemberDetailsView"},
+            [   t.view(new AvatarView(vm, 128)),
```
MidhunSureshR, line 23: Avatars loaded here are blurry; they're smaller than the requested size. Probably a synapse bug.

src/domain/navigation/index.js
```diff
@@ -132,10 +132,12 @@ export function parseUrlPath(urlPath, currentNavPath, defaultSessionId) {
...
             segments.push(new Segment("room", roomId));
-            if (currentNavPath.get("details")?.value) {
-                pushRightPanelSegment(segments, "details");
-            } else if (currentNavPath.get("members")?.value) {
-                pushRightPanelSegment(segments, "members");
+            // Add right-panel segments from previous path
+            const previousSegments = currentNavPath.segments;
+            const i = previousSegments.findIndex(s => s.type === "right-panel");
+            if (i !== -1) {
+                segments.push(previousSegments[i]);
```
bwindels: why not just copy over all segments over after right-panel?

src/domain/session/rightpanel/MemberDetailsViewModel.js
```diff
@@ -0,0 +1,81 @@
...
+
+export class MemberDetailsViewModel extends ViewModel {
+    constructor(options) {
+        super(options);
+        this._observableMember = options.observableMember;
+        this._mediaRepository = options.mediaRepository;
+        this._member = this._observableMember.get();
+        this._isEncrypted = options.isEncrypted;
+        this._powerLevelsObservable = options.powerLevelsObservable;
+        this._powerLevel = this._powerLevelsObservable.get().getUserLevel(this._member.userId);
```
bwindels: you could make this a virtual property, by implementing it as a getter

src/domain/session/rightpanel/MemberDetailsViewModel.js
```diff
@@ -0,0 +1,81 @@
...
+
+    _onMemberChange() {
+        this._member = this._observableMember.get();
+        this.emitChange();
+    }
+
+    _onPowerLevelsChange() {
+        const powerLevels = this._powerLevelsObservable.get();
+        this._powerLevel = powerLevels.getUserLevel(this._member.userId);
+        this.emitChange();
```
bwindels: although not really used for anything today, it's a custom to mention the name of the property that should be checked again (here `role`)

src/domain/session/rightpanel/MemberDetailsViewModel.js
```diff
@@ -0,0 +1,81 @@
...
+
+    get name() { return this._member.name; }
+    get userId() { return this._member.userId; }
+
+    get type() { return "member-details"; }
+    get shouldShowBackButton() { return true; }
+    get previousSegmentName() { return "members"; }
+    
+    get role() {
+        if (this._powerLevel >= 100) { return "Admin"; }
```
bwindels: For user-visible strings, you can also use this.i18n here

src/platform/web/ui/session/rightpanel/MemberDetailsView.js
```diff
@@ -0,0 +1,54 @@
...
+                t.div({className: "MemberDetailsView_name"}, t.h2(vm => vm.name)),
+                t.div({className: "MemberDetailsView_id"}, vm.userId),
+                this._createSection(t, vm.i18n`ROLE`, vm => vm.role),
```
bwindels: Perhaps better to upper-case with css text-transform

src/platform/web/ui/session/rightpanel/MemberDetailsView.js
bwindels: So in Element Web, the avatar is sized to 2x30vh: https://github.com/matrix-org/matrix-react-sdk/blob/e78a1adb6f1af2ea425b0bae9034fb7344a4b2e8/src/components/views/right_panel/UserInfo.tsx#L1454-L1456 but the size is dynamic there depending on the right panel width. Prior to that it was 800x800, so if there is a synapse bug here, it might have been undetected so far. If you can accurately describe the problem you see with thumbnails, might be good to file an issue on https://github.com/matrix-org/synapse/issues/ and refer to that in the comment or commit for why 128 is not good enough. Also, make sure `window.devicePixelRatio` is 1 on the device/browser where you are seeing the problem, otherwise this is expected. On that subject, we should multiply the number here with `this._platform.devicePixelRatio` in any case.

src/platform/web/ui/session/rightpanel/MemberDetailsView.js
```diff
@@ -0,0 +1,54 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import {AvatarView} from "../../AvatarView.js";
+import {TemplateView} from "../../general/TemplateView.js";
+
+export class MemberDetailsView extends TemplateView {
+    render(t, vm) {
+        const encryptionMessages = {
```
bwindels: just use ?: here rather than creating an extra object?

bwindels (commented):
first round of review, looking good!

src/domain/navigation/index.js
bwindels: ah, you wouldn't want to copy the member segment over as we can't assume that the other room will have the same member...

bwindels: will this actually just copy over the members segment but not the member segment?

MidhunSureshR: >why not just copy over all segments over after right-panel?

Good idea! We currently only have one segment after the "right-panel" segment but in future we may have more.

src/domain/navigation/index.js
```diff
@@ -132,10 +132,11 @@ export function parseUrlPath(urlPath, currentNavPath, defaultSessionId) {
...
             segments.push(new Segment("room", roomId));
-            if (currentNavPath.get("details")?.value) {
-                pushRightPanelSegment(segments, "details");
-            } else if (currentNavPath.get("members")?.value) {
-                pushRightPanelSegment(segments, "members");
+            // Add right-panel segments from previous path
+            const previousSegments = currentNavPath.segments;
+            const i = previousSegments.findIndex(s => s.type === "right-panel");
+            if (i !== -1) {
+                segments.push(...previousSegments.slice(i));
```
bwindels, line 139: I don't think we want to copy the `member` segment though ... perhaps only copy until you find a segment which value isn't `true`?

src/domain/session/rightpanel/MemberDetailsViewModel.js
```diff
@@ -0,0 +1,89 @@
...
+    get isEncrypted() {
+        return this._isEncrypted;
+    }
+
+    get powerLevel() {
+        if (!this._powerLevel) {
+            const powerLevels = this._powerLevelsObservable.get();
+            this._powerLevel = powerLevels.getUserLevel(this._member.userId);
+        }
+        return this._powerLevel;
```
bwindels: good start, but I meant more that we don't need to cache this even, as it's all derived from things we already have:
```suggestion
        return this._powerLevelsObservable.get()?.getUserLevel(this._member.userId);
```

src/domain/navigation/index.js
bwindels: We're going to address this in a follow-up PR

src/domain/session/rightpanel/MemberDetailsViewModel.js
```diff
@@ -0,0 +1,83 @@
...
+
+export class MemberDetailsViewModel extends ViewModel {
+    constructor(options) {
+        super(options);
+        this._observableMember = options.observableMember;
+        this._mediaRepository = options.mediaRepository;
+        this._member = this._observableMember.get();
+        this._isEncrypted = options.isEncrypted;
+        this._powerLevelsObservable = options.powerLevelsObservable;
+        this._powerLevel = null;
```
bwindels: can also remove this now

src/domain/session/rightpanel/MemberDetailsViewModel.js
MidhunSureshR: Oops! Fixed now.

src/domain/session/rightpanel/RightPanelViewModel.js
```diff
@@ -1,42 +1,70 @@
...
+            }
+        );
     }
 
-    _hookUpdaterToSegment(segment, viewmodel, argCreator) {
+    _hookUpdaterToSegment(segment, viewmodel, argCreator, failCallback) {
         const observable = this.navigation.observe(segment);
-        const updater = this._setupUpdater(segment, viewmodel, argCreator);
+        const updater = this._setupUpdater(segment, viewmodel, argCreator, failCallback);
         this.track(observable.subscribe(() => updater()));
```
bwindels: this can be just fwiw
```suggestion
        this.track(observable.subscribe(updater));
```

src/platform/web/ui/session/rightpanel/MemberDetailsView.js
bwindels: the spec says homeservers should only support certain resolutions for each scale mode, with the tl;dr for now being that we have to choose between blurry avatars or avatars that don't respect the aspect-ratio of the design. We're going with blurry.

src/platform/web/ui/session/rightpanel/RightPanelView.js
```diff
@@ -2,12 +2,14 @@ import {TemplateView} from "../../general/TemplateView.js";
 import {RoomDetailsView} from "./RoomDetailsView.js";
 import {MemberListView} from "./MemberListView.js";
 import {LoadingView} from "../../general/LoadingView.js";
+import {MemberDetailsView} from "./MemberDetailsView.js";
 
 export class RightPanelView extends TemplateView {
     render(t) {
         const viewFromType = {
             "room-details": RoomDetailsView,
-            "member-list": MemberListView
```
bwindels, line 10: why prefer this over a switch statement?

bwindels (approved):
some minor comments, but basically looks good to go

bwindels (approved):
Thanks! This is good to go!

PR #418, opened 2021-07-20
Sending replies

DanilaFe:
Opening this primarily to keep track of what's left to be done. So far there's:

- [X] Add a reply button to messages in the timeline that puts us in reply mode
- [X] Make sure replies to pending events work (this requires expanding the code that resolves remote echo to still work with the shape of reply events, which use `m.relates_to`, but not the same way as annotations).
- [ ] Proper message formatting. From the spec:
    > Clients that do support rich replies MUST provide the fallback format on replies, and MUST strip the fallback before rendering the reply. Rich replies MUST have a format of org.matrix.custom.html and therefore a formatted_body alongside the body and appropriate msgtype.

   That's a lot of "MUST"! We don't even send HTML messages yet.
  - [x] Properly _insert_ reply fallback, in both plain `body` and `formatted_body`.
  - [x] Properly _strip_ reply fallback
  - [ ] Render replies using another mechanism.
- [x] Indicate in the UI that we're replying to a message. This may be a bit problematic since the rendering of regular message tiles is not quite what we want (hides avatars, shows replies, etc).
- [ ] Test that everything still works as intended. There are some changes to `SendQueue`, for instance.

### DIFF
--- a/src/domain/session/room/ComposerViewModel.js
+++ b/src/domain/session/room/ComposerViewModel.js
@@ -47,8 +47,8 @@ export class ComposerViewModel extends ViewModel {
         return this._roomVM.isEncrypted;
     }
 
-    sendMessage(message) {
-        const success = this._roomVM._sendMessage(message, this._replyVM);
+    async sendMessage(message) {
+        const success = await this._roomVM._sendMessage(message, this._replyVM);
         if (success) {
             this._isEmpty = true;
             this.emitChange("canSend");
--- a/src/domain/session/room/ComposerViewModel.js
+++ b/src/domain/session/room/ComposerViewModel.js
@@ -5,13 +5,23 @@ export class ComposerViewModel extends ViewModel {
         super();
         this._roomVM = roomVM;
         this._isEmpty = true;
+        this._replyId = null;
         this._replyVM = null;
     }
 
-    setReplyingTo(tile) {
-        const changed = this._replyVM !== tile;
-        this._replyVM = tile;
+    setReplyingTo(entry) {
+        const newId = entry?.id || null;
+        const changed = this._replyId !== newId;
         if (changed) {
+            this._replyId = newId;
+            if (this._replyVM) {
+                this.untrack(this._replyVM);
+                this._replyVM.dispose();
+            }
+            this._replyVM = entry && this._roomVM._createTile(entry);
+            if (this._replyVM) {
+                this.track(this._replyVM);
+            }
             this.emitChange("replyViewModel");
         }
     }
--- a/src/domain/session/room/ComposerViewModel.js
+++ b/src/domain/session/room/ComposerViewModel.js
@@ -1,0 +1,71 @@
+import {ViewModel} from "../../ViewModel.js";
+
+export class ComposerViewModel extends ViewModel {
+    constructor(roomVM) {
+        super();
+        this._roomVM = roomVM;
+        this._isEmpty = true;
+        this._replyVM = null;
+    }
+
+    setReplyingTo(tile) {
+        const changed = this._replyVM !== tile;
+        this._replyVM = tile;
+        if (changed) {
+            this.emitChange("replyViewModel");
+        }
+    }
+
+    clearReplyingTo() {
+        this.setReplyingTo(null);
+    }
+
+    get replyViewModel() {
+        return this._replyVM;
+    }
+
+    get isEncrypted() {
+        return this._roomVM.isEncrypted;
+    }
+
+    sendMessage(message) {
+        const success = this._roomVM._sendMessage(message, this._replyVM);
+        if (success) {
+            this._isEmpty = true;
+            this.emitChange("canSend");
+            this.clearReplyingTo();
+        }
+        return success;
+    }
+
+    sendPicture() {
+        this._roomVM._pickAndSendPicture();
+    }
+
+    sendFile() {
+        this._roomVM._pickAndSendFile();
+    }
+
+    sendVideo() {
+        this._roomVM._pickAndSendVideo();
+    }
+
+    get canSend() {
+        return !this._isEmpty;
+    }
+
+    async setInput(text) {
+        const wasEmpty = this._isEmpty;
+        this._isEmpty = text.length === 0;
+        if (wasEmpty && !this._isEmpty) {
+            this._roomVM._room.ensureMessageKeyIsShared();
+        }
+        if (wasEmpty !== this._isEmpty) {
+            this.emitChange("canSend");
+        }
+    }
+
+    get kind() {
+        return "composer";
+    }
+}
--- a/src/domain/session/room/ComposerViewModel.js
+++ b/src/domain/session/room/ComposerViewModel.js
@@ -1,3 +1,19 @@
+/*
+Copyright 2021 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 import {ViewModel} from "../../ViewModel.js";
 
 export class ComposerViewModel extends ViewModel {
--- a/src/domain/session/room/ComposerViewModel.js
+++ b/src/domain/session/room/ComposerViewModel.js
@@ -21,15 +21,12 @@ export class ComposerViewModel extends ViewModel {
         super();
         this._roomVM = roomVM;
         this._isEmpty = true;
-        this._replyId = null;
         this._replyVM = null;
     }
 
     setReplyingTo(entry) {
-        const newId = entry?.id || null;
-        const changed = this._replyId !== newId;
+        const changed = this._replyVM?.internalId !== entry?.asEventKey().toString();
         if (changed) {
-            this._replyId = newId;
             if (this._replyVM) {
                 this.untrack(this._replyVM);
                 this._replyVM.dispose();
--- a/src/domain/session/room/ComposerViewModel.js
+++ b/src/domain/session/room/ComposerViewModel.js
@@ -27,13 +27,9 @@ export class ComposerViewModel extends ViewModel {
     setReplyingTo(entry) {
         const changed = this._replyVM?.internalId !== entry?.asEventKey().toString();
         if (changed) {
-            if (this._replyVM) {
-                this.untrack(this._replyVM);
-                this._replyVM.dispose();
-            }
-            this._replyVM = entry && this._roomVM._createTile(entry);
-            if (this._replyVM) {
-                this.track(this._replyVM);
+            this._replyVM = this.disposeTracked(this._replyVM);
+            if (entry) {
+                this._replyVM = this.track(this._roomVM._createTile(entry));
             }
             this.emitChange("replyViewModel");
         }
--- a/src/domain/session/room/RoomViewModel.js
+++ b/src/domain/session/room/RoomViewModel.js
@@ -155,7 +155,7 @@ export class RoomViewModel extends ViewModel {
         this._room.join();
     }
     
-    async _sendMessage(message, replyingTo) {
+    async _sendMessage(message, reply) {
         if (!this._room.isArchived && message) {
             try {
                 let msgtype = "m.text";
@@ -164,8 +164,8 @@ export class RoomViewModel extends ViewModel {
                     msgtype = "m.emote";
                 }
                 const content = {msgtype, body: message};
-                if (replyingTo) {
-                    content["m.relates_to"] = replyingTo.reply();
+                if (reply) {
+                    content["m.relates_to"] = reply;
                 }
                 await this._room.sendEvent("m.room.message", content);
             } catch (err) {
@@ -302,7 +302,7 @@ export class RoomViewModel extends ViewModel {
 
     startReply(entry) {
         if (!this._room.isArchived) {
-            this._composerVM.startReply(entry);
+            this._composerVM.setReplyingTo(entry);
         }
     }
 }
@@ -311,11 +311,24 @@ class ComposerViewModel extends ViewModel {
     constructor(roomVM) {
         super();
         this._roomVM = roomVM;
-        this._replyVM = new ReplyViewModel(roomVM);
+        this._isEmpty = true;
+        this._replyVM = null;
     }
 
-    startReply(entry) {
-        this._replyVM.setReplyingTo(entry);
+    setReplyingTo(tile) {
+        const changed = this._replyVM !== tile;
+        this._replyVM = tile;
+        if (changed) {
+            this.emitChange("replyViewModel");
+        }
+    }
+
+    clearReplyingTo() {
+        this.setReplyingTo(null);
+    }
+
+    get replyViewModel() {
+        return this._replyVM;
     }
 
     get isEncrypted() {
@@ -323,11 +336,11 @@ class ComposerViewModel extends ViewModel {
     }
 
     sendMessage(message) {
-        const success = this._roomVM._sendMessage(message, this._replyVM.replyingTo);
+        const success = this._roomVM._sendMessage(message, this._replyVM?.reply());
         if (success) {
             this._isEmpty = true;
             this.emitChange("canSend");
-            this._replyVM.clearReply();
+            this.clearReplyingTo();
         }
         return success;
     }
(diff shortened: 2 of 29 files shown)


src/matrix/room/timeline/entries/BaseEventEntry.js
```diff
@@ -151,6 +155,58 @@ export class BaseEventEntry extends BaseEntry {
         return createAnnotation(this.id, key);
     }
 
+    _fallbackBlurb() {
```
bwindels: perhaps some of the reply code here (and createReply in relations?) can go into a separate `reply.js` module here in the same folder? This class might be growing a bit large ...

src/matrix/room/timeline/entries/BaseEventEntry.js
```diff
@@ -151,6 +155,58 @@ export class BaseEventEntry extends BaseEntry {
...
+        switch (this.content.msgtype) {
+            case "m.file":
+                return "sent a file.";
+            case "m.image":
+                return "sent an image.";
+            case "m.video":
+                return "sent a video.";
+            case "m.audio":
+                return "sent an audio file.";
+        }
```
bwindels: There are quite a few getter methods and properties here. Perhaps name this function _fallbackReplyBody, pass in a `formatted` flag, and on `default`, inline what you now have in _formattedBody if formatted, otherwise inline what you now have in `_plainBody`.

src/matrix/room/timeline/entries/BaseEventEntry.js
```diff
@@ -151,6 +155,58 @@ export class BaseEventEntry extends BaseEntry {
...
+            <br />
+            ${body}
+          </blockquote>
+        </mx-reply>`;
+    }
+
+    _replyPlainFallback() {
+        const body = this._fallbackBlurb() || this._plainBody || "";
+        const bodyLines = body.split("\n");
+        bodyLines[0] = `> <${this.sender}> ${bodyLines[0]}`
```
bwindels: don't you need _fallbackPrefix here as well?

src/matrix/room/timeline/entries/BaseEventEntry.js
```diff
@@ -151,6 +155,58 @@ export class BaseEventEntry extends BaseEntry {
...
+    }
+
+    _replyFormattedFallback() {
+        const body = this._fallbackBlurb() || this._formattedBody || "";
+        const prefix = this._fallbackPrefix();
+        return `<mx-reply>
+          <blockquote>
+            In reply to
+            ${prefix}<a href="https://matrix.to/#/${this.sender}">${this.displayName || this.sender}</a>
+            <br />
```
bwindels: m.emote messages should not have a newline here AIUI

src/platform/web/ui/session/room/timeline/BaseMessageView.js
```diff
@@ -24,34 +24,38 @@ import {Menu} from "../../../general/Menu.js";
...
     }
 
     render(t, vm) {
-        const li = t.li({className: {
+        const li = t.el(this._tagName, {className: {
             "Timeline_message": true,
             own: vm.isOwn,
             unsent: vm.isUnsent,
             unverified: vm.isUnverified,
+            disabled: this._disabled,
```
bwindels: what is this for? it's only a valid attribute on form elements fwiw ...

src/platform/web/ui/session/room/timeline/BaseMessageView.js
```diff
@@ -24,34 +24,38 @@ import {Menu} from "../../../general/Menu.js";
 import {ReactionsView} from "./ReactionsView.js";
 
 export class BaseMessageView extends TemplateView {
-    constructor(value) {
+    constructor(value, disabled = false, tagName = "li") {
         super(value);
         this._menuPopup = null;
+        this._tagName = tagName;
+        // TODO An enum could be nice to make code easier to read at call sites.
+        this._disabled = disabled;
```
bwindels: perhaps a more specific name like `interactive` would be better here

src/domain/session/room/RoomViewModel.js
```diff
@@ -294,24 +299,48 @@ export class RoomViewModel extends ViewModel {
...
     }
+
+    startReply(entry) {
+        if (!this._room.isArchived) {
+            this._composerVM.setReplyingTo(entry);
+        }
+    }
 }
 
 class ComposerViewModel extends ViewModel {
```
bwindels: perhaps it's a good time to extract this view model to a separate file

src/domain/session/room/RoomViewModel.js
```diff
@@ -294,24 +299,48 @@ export class RoomViewModel extends ViewModel {
...
     constructor(roomVM) {
         super();
         this._roomVM = roomVM;
         this._isEmpty = true;
+        this._replyVM = null;
+    }
+
+    setReplyingTo(tile) {
+        const changed = this._replyVM !== tile;
```
bwindels: In order to not have to hold on to the tile until you hit send (and risk the tile being disposed in the meantime as you pointed out), I'd organize the code a bit differently here:
 - you observe the event by calling BaseRoom.observeEvent, which returns an observable of an EventEntry. The only problem here might be that this currently only works for already sent events...
 - create a copy of the tile by calling the tilesCreator with the entry. Creating a copy will also ensure that isContinuation is false, so you don't have to special case this in the view.
 - you subscribe to the observable and call updateEntry on the tile with the updated entry, this way the preview should also update. You'll need to check the result (see SimpleTile) and call emitChange if it returns an update and replace the tile and emit an update on the this view model if a replacement is returned.

src/domain/session/room/timeline/TimelineViewModel.js
```diff
@@ -38,9 +38,9 @@ import {ViewModel} from "../../../ViewModel.js";
 export class TimelineViewModel extends ViewModel {
     constructor(options) {
         super(options);
-        const {room, timeline} = options;
+        const {room, timeline, roomVM} = options;
```
bwindels: hmm, wonder if there is a way to not pass both the roomVM and room in here ... need to think about this.

bwindels (commented):
Great work, this seems to work quite well! Just a bit of cleanup comments to go through still.

src/platform/web/ui/session/room/timeline/BaseMessageView.js
DanilaFe: I hope I'm doing the right thing here, but the intention is to add a `disabled` class (this is inside the `className` object, isn't it?)

src/matrix/room/timeline/entries/BaseEventEntry.js
DanilaFe: `default` doesn't seem appropriate here since even the blurbs like "sent a file" need to be wrapped in `<mx-reply>` etc. as well. This is why `_replyFormattedFallback` calls out to `fallbackBlurb` and then wraps it in the HTML.

src/matrix/room/timeline/entries/BaseEventEntry.js
DanilaFe: Indeed!

src/matrix/room/timeline/entries/BaseEventEntry.js
```diff
@@ -155,55 +156,26 @@ export class BaseEventEntry extends BaseEntry {
...
-        const bodyLines = body.split("\n");
-        bodyLines[0] = `> <${this.sender}> ${bodyLines[0]}`
-        return bodyLines.join("\n> ");
-    }
-
     reply(msgtype, body) {
         // TODO check for absense of sender / body / msgtype / etc?
-        const newBody = this._replyPlainFallback() + '\n\n' + body;
-        const newFormattedBody = this._replyFormattedFallback() + htmlEscape(body);
+        let blurb = fallbackBlurb(this.content.msgtype);
```
bwindels: I assumed "blurb" was a leftover from doubting what to call it, but now it seems intentional :) Perhaps this has a meaning in English I am unaware of? If not, I would suggest something more specific like `fallbackTextForNonTextualMessage`

src/matrix/room/timeline/entries/reply.js
```diff
@@ -0,0 +1,33 @@
+
```
bwindels, line 3: Matrix foundation copyright header please, you can copy it from other files.

src/matrix/room/timeline/entries/BaseEventEntry.js
bwindels: Also, this can be `const`?

src/matrix/room/timeline/entries/BaseEventEntry.js
```diff
@@ -155,55 +156,26 @@ export class BaseEventEntry extends BaseEntry {
...
+        const prefix = fallbackPrefix(this.content.msgtype);
+        const sender = this.sender;
+        const name = this.displayName || sender;
```
bwindels: Given that all the properties used on `this` here are public, shall we just put all of this in `reply.js` and pass in the entry in that function? Just to not have HTML stuff in here ...

src/platform/web/ui/session/room/timeline/BaseMessageView.js
bwindels: ahh, yes, I misread and thought this were just attributes on the element.

src/matrix/room/sending/PendingEvent.js
```diff
@@ -28,6 +28,10 @@ export const SendStatus = createEnum(
     "Error",
 );
 
+const preservedContentFields = {
+    "m.room.message": [ "m.relates_to" ]
```
bwindels: We can do this independent of the event type I think, just like [element](https://github.com/matrix-org/matrix-js-sdk/blob/a69d9a9f20b680e3710e0f43cca68a2ce31ba030/src/crypto/index.ts#L2753) does. Also, to be correct, we should remove it from the content to be encrypted.

src/matrix/room/timeline/entries/reply.js
```diff
@@ -31,3 +50,25 @@ export function createReply(targetId, msgtype, body, formattedBody) {
     };
 }
 
+export function reply(entry, msgtype, body) {
```
bwindels: call this `createReplyContent` to make clear this does not actually send the reply? (and inline or rename the other `createReply`?)

src/matrix/room/sending/PendingEvent.js
```diff
@@ -96,7 +98,25 @@ export class PendingEvent {
         this._emitUpdate("status");
     }
 
+    get cleanedContent() {
```
bwindels: how about a name more specific to encryption? e.g. `contentForEncryption`?

src/matrix/room/sending/PendingEvent.js
```diff
@@ -28,6 +28,8 @@ export const SendStatus = createEnum(
     "Error",
 );
 
+const preservedContentFields = [ "m.relates_to" ];
```
bwindels: Add a comment here or name it clarifying this is specifically for encryption?

src/domain/session/room/ComposerViewModel.js
```diff
@@ -0,0 +1,81 @@
...
+        this._replyVM = null;
+    }
+
+    setReplyingTo(entry) {
+        const newId = entry?.id || null;
+        const changed = this._replyId !== newId;
+        if (changed) {
+            this._replyId = newId;
+            if (this._replyVM) {
+                this.untrack(this._replyVM);
```
bwindels: you can use this._replyVM = this.disposeTracked(this._replyVM) here

src/domain/session/room/ComposerViewModel.js
```diff
@@ -0,0 +1,81 @@
+import {ViewModel} from "../../ViewModel.js";
```
bwindels, line 17: copyright header please :pray:

src/domain/session/room/ComposerViewModel.js
```diff
@@ -0,0 +1,81 @@
...
+        super();
+        this._roomVM = roomVM;
+        this._isEmpty = true;
+        this._replyId = null;
+        this._replyVM = null;
+    }
+
+    setReplyingTo(entry) {
+        const newId = entry?.id || null;
+        const changed = this._replyId !== newId;
```
bwindels: you could use `this._replyVM.internalId` (see SimpleTile) here and compare it with `entry.asEventKey().toString()`, so you can avoid the `_replyId` field altogether.

src/domain/session/room/ComposerViewModel.js
```diff
@@ -0,0 +1,81 @@
...
+    get replyViewModel() {
+        return this._replyVM;
+    }
+
+    get isEncrypted() {
+        return this._roomVM.isEncrypted;
+    }
+
+    sendMessage(message) {
+        const success = this._roomVM._sendMessage(message, this._replyVM);
```
bwindels: oops, this was already missing an await before your PR. Can you add it please? :pray:

src/domain/session/room/timeline/TilesCollection.js
```diff
@@ -181,6 +181,8 @@ export class TilesCollection extends BaseObservableList {
     }
 
     _replaceTile(tileIdx, existingTile, newTile, updateParams) {
+        // TODO What happens with a tile that's being replied to? Can we have
```
bwindels: can remove this comment now

src/domain/session/room/timeline/deserialize.js
```diff
@@ -389,7 +398,7 @@ export function tests() {
     };
 
     function test(assert, input, output) {
-        assert.deepEqual(parseHTMLBody(platform, null, input), new MessageBody(input, output));
+        assert.deepEqual(parseHTMLBody(platform, null, true, input), new MessageBody(input, output));
```
bwindels: can we perhaps have a test for filtering out the nested tags of mx-reply when this flag is false?

src/domain/session/room/timeline/tiles/SimpleTile.js
```diff
@@ -129,6 +129,10 @@ export class SimpleTile extends ViewModel {
         return this._options.room;
```
bwindels: can we make RoomViewModel._room public, change this to _roomVM.room and not pass the room in and have one option less?

src/matrix/room/ObservedEventMap.js
```diff
@@ -25,7 +25,7 @@ export class ObservedEventMap {
     observe(eventId, eventEntry = null) {
         let observable = this._map.get(eventId);
         if (!observable) {
-            observable = new ObservedEvent(this, eventEntry);
```
bwindels, line 28: what were these changes about? are they still desirable now that we didn't end up going with observeEvent?

src/platform/web/ui/session/room/timeline/BaseMessageView.js
```diff
@@ -24,34 +24,38 @@ import {Menu} from "../../../general/Menu.js";
...
+            this._interactive ? t.button({className: "Timeline_messageOptions"}, "⋯") : [],
         ]);
         // given that there can be many tiles, we don't add
         // unneeded DOM nodes in case of a continuation, and we add it
         // with a side-effect binding to not have to create sub views,
         // as the avatar or sender doesn't need any bindings or event handlers.
         // don't use `t` from within the side-effect callback
         t.mapSideEffect(vm => vm.isContinuation, (isContinuation, wasContinuation) => {
-            if (isContinuation && wasContinuation === false) {
+            if (isContinuation && this._interactive && wasContinuation === false) {
```
bwindels: now that we clone the tile, I thing isContinuation should always be false, so I don't think we need this anymore.

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -462,16 +462,55 @@ a {
...
+    display: block;
+    border: none;
+    text-indent: 200%;
+    white-space: nowrap;
+    overflow: hidden;
+    background-color: transparent;
+    background-image: url('icons/clear.svg');
+    background-repeat: no-repeat;
+    background-position: center;
+    background-size: 18px;
```
bwindels, line 502: can has `cursor: pointer` here?

bwindels (commented):
Excellent work, thanks for seeing it through! Just a bunch more cleanup comments, but this is basically good to go!

src/matrix/room/timeline/entries/reply.js
DanilaFe: Sounds good! I think, however, that we should keep `BaseEventEntry`'s `reply` as `reply`, to match `annotate`.

src/matrix/room/ObservedEventMap.js
DanilaFe: I believe that there's a bug here. If you try to unsubscribe from an observed event whose content is null (which is possible, for instance, if we can't retrieve an event from storage, or just haven't yet), you end up de-referencing a null when trying to remove the observedevent from the map (null.id).

src/platform/web/ui/css/themes/element/theme.css
DanilaFe: Indeed; I think the send and attach file buttons are missing this as well.

src/domain/session/room/ComposerViewModel.js
DanilaFe: Does this mean making `sendMessage` and `_trySend` async, as well? They depend on the results of `_sendMessage`.

src/matrix/room/ObservedEventMap.js
bwindels: gotcha, good catch!

bwindels (approved):
lgtm!

PR #427, opened 2021-07-30
[SSO] - [PR 1] - Rewrite password login to use PasswordLoginMethod

MidhunSureshR:


src/matrix/login/PasswordLoginMethod.js
```diff
@@ -0,0 +1,31 @@
...
+
+export class PasswordLoginMethod extends LoginMethod {
+    constructor(options) {
+        super(options);
+        this.username = options.username;
+        this.password = options.password;
+    }
+
+    async login(hsApi, deviceName) {
+        return this._platform.logger.run("passwordLogin", async log =>
```
bwindels: probably better to pass down the log item already in progress in startWithLogin, rather than to start a new log operation. The `log` argument is usually passed down as the last argument btw. Passing it down makes the /password request being logged as a child of the ongoing operation, I can show you what that looks like in the log viewer.

bwindels (commented):
Looks great, just need to pass down logging

bwindels:
Obsoleted by #453

PR #429, opened 2021-07-31
[SSO] - [PR 3] - UI for SSO Login

MidhunSureshR:
- [x] Show "Login with SSO" ~~link~~ button only if hs supports it
- [x] ~~Show CompleteSSOView on RootView when sso segment is observed~~
- [x] Split LoginView into PasswordLoginView & SSOLoginView
- [x] Style SSO button
- [x] Style CompleteSSOView


src/domain/LoginViewModel.js
```diff
@@ -17,6 +17,14 @@ limitations under the License.
 import {ViewModel} from "./ViewModel.js";
 import {SessionLoadViewModel} from "./SessionLoadViewModel.js";
 
+function normalizeHomeserver(homeServer) {
```
bwindels: why move this here rather than doing it in session container?

bwindels:
Obsoleted by #453

PR #432, opened 2021-08-04
Member Panel - PR 3 - Update powerlevels from state event

MidhunSureshR:


src/matrix/room/Room.js
```diff
@@ -173,6 +174,7 @@ export class Room extends BaseRoom {
         if (Array.isArray(roomResponse.timeline?.events)) {
             removedPendingEvents = await this._sendQueue.removeRemoteEchos(roomResponse.timeline.events, txn, log);
         }
+        this._updatePowerLevels(roomResponse);
```
bwindels: This (e.g. updating in-memory data) should happen in afterSync, rather than writeSync, which is only about trying to persist the room response and returning data to incrementally update in-memory data with in afterSync. This is so that in-memory data structures are not altered until the storage transaction has successfully committed, so in case something goes wrong, we can safely retry without having data being out of sync between storage and in-memory. afterSync receives the return value from writeSync for this purpose. So here you'll actually want to return the new state events from the room response, and then call `_updatePowerLevels` with it in afterSync.

To find out about new state events, you'll need to look both at `roomResponse.state.events` (any state events since the last sync and the beginning of the returned timeline chunk in this sync response) and `roomResponse.timeline.events` (where state events are mixed with non-state events, and you can just filter events that have a state_key)

src/matrix/room/Room.js
```diff
@@ -264,9 +268,14 @@ export class Room extends BaseRoom {
         }
     }
 
-    _updatePowerLevels(roomResponse) {
-        const powerLevelEvent = roomResponse.timeline.events.find(event => event.type === POWERLEVELS_EVENT_TYPE);
-        if (powerLevelEvent && this._powerLevels) {
+    _getPowerLevelsEvent(roomResponse) {
+        const isPowerlevelEvent = event => event.type === POWERLEVELS_EVENT_TYPE;
+        const powerLevelEvent = roomResponse.state?.events.find(isPowerlevelEvent) ?? roomResponse.timeline?.events.find(isPowerlevelEvent);
```
bwindels: wonder if you need `?.find` here?
Also, state events in timeline.events come after the state events in state.events, so you should first look in timeline.events and only if none found look in state.events to find the most recent power level event (as both arrays could have it)

MidhunSureshR: Hmm.. I don't know what I did but I did get something similar to "Sync Error: roomResponse.state is undefined". The same for roomResponse.timeline.

MidhunSureshR: Can't really reproduce anymore so I'll remove the `?` for now. Will keep testing though.

bwindels: no, I actually thought you may need an extra `?` when calling find, e.g:
```js
const powerLevelEvent = roomResponse.state?.events?.find(isPowerlevelEvent) ?? roomResponse.timeline?.events?.find(isPowerlevelEvent); 
```
but I misunderstood how `?` works and this does not seem to be needed, in node:
```
> let foo = {};
undefined
> foo.events?.bar.foo.find()
undefined
> 
```
So you initially had it correct with just a `?` after state and timeline
```js
const powerLevelEvent = roomResponse.state?.events.find(isPowerlevelEvent) ?? roomResponse.timeline?.events.find(isPowerlevelEvent); 
```

src/matrix/room/Room.js
```diff
@@ -250,6 +259,32 @@ export class Room extends BaseRoom {
...
+        for (const [userId, memberChange] of memberChanges) {
+            const observableMember = this._observedMembers.get(userId);
+            if (observableMember) {
+                observableMember.set(memberChange.member);
+            }
+        }
+    }
+
+    _getPowerLevelsEvent(roomResponse) {
+        const isPowerlevelEvent = event => event.type === POWERLEVELS_EVENT_TYPE;
```
bwindels: can we also check here there is a `state_key` on the event, to make sure people can't sneak non-state events in the power levels

src/matrix/room/Room.js
```diff
@@ -23,6 +23,7 @@ import {WrappedError} from "../error.js"
 import {Heroes} from "./members/Heroes.js";
 import {AttachmentUpload} from "./AttachmentUpload.js";
 import {DecryptionSource} from "../e2ee/common.js";
+import {PowerLevels, EVENT_TYPE as POWERLEVELS_EVENT_TYPE } from "./timeline/PowerLevels.js";
```
bwindels: can you move this file one level up now that it's not exclusively used for the timeline anymore?

bwindels (approved):
looks great, just one small cleanup but basically ready to go

bwindels:
did you test btw that this updates the member panel role when changing the power level?

PR #448, opened 2021-08-06
Snowpack MVP

DanilaFe:
Experimental PR to add Snowpack support. This should make adding dependencies less painful (no more modifying `post-install.js` to add packages!) and give us TypeScript support. In this particular PR, there are minimum changes that prove the viability of Snowpack.

* Replace DOMPurify's `lib` import with a regular import, and remove its symlinking step from `post-install.js`.
* Convert `BaseEntry` to TypeScript
* Switch `yarn start` to serve via Snowpack. 
* Add a basic TypeScript configuration, so that we can run `yarn tsc` to typecheck our files.

After this, we can still run Hydrogen in the browser (though impunity does not, as yet, support TypeScript). Open questions/remarks:
* We occasionally get a `404` for `sw.js` while serving. Should `sw.js` also be directly mounted on `/`?
* The old `serve-local.js` script doesn't work anymore, but is kept as is for now.
* Even though I updated `build.mjs` to copy `index.html` from its new location, the build still fails due to `dompurify` and TypeScript.


bwindels:
> We occasionally get a 404 for sw.js while serving. Should sw.js also be directly mounted on /?

So for local dev, we don't use a service worker. For prod builds, sw.js should be at the same level as index.html.

bwindels:
> Even though I updated build.mjs to copy index.html from its new location, the build still fails due to dompurify and TypeScript.

Do you think this is fixable in this PR, by having build.mjs run snowpack/typescript first and then run rollup on the result of that?

DanilaFe:
> > Even though I updated build.mjs to copy index.html from its new location, the build still fails due to dompurify and TypeScript.
> 
> Do you think this is fixable in this PR, by having build.mjs run snowpack/typescript first and then run rollup on the result of that?

I got this working by invoking snowpack from the script and reading files from `build/src` instead of `src`.

DanilaFe:
> > We occasionally get a 404 for sw.js while serving. Should sw.js also be directly mounted on /?
> 
> So for local dev, we don't use a service worker. For prod builds, sw.js should be at the same level as index.html.

So is 404 in development on `sw.js` the expected behavior, then? That's what currently happens (though Hydrogen still works)

bwindels:
What was the conclusion again on using `.ts` extensions in the imports? I remember there was a problem with it, but when changing one of the imports of `BaseEntry`, it still seems to work ... probably missing something.

bwindels:
> > > We occasionally get a 404 for sw.js while serving. Should sw.js also be directly mounted on /?
> > 
> > 
> > So for local dev, we don't use a service worker. For prod builds, sw.js should be at the same level as index.html.
> 
> So is 404 in development on `sw.js` the expected behavior, then? That's what currently happens (though Hydrogen still works)

Oh, could this be because /sw.js is symlinked to the actual service worker file?

scripts/build.mjs
```diff
@@ -56,14 +59,19 @@ parameters
     .option("--override-css <main css file>", "pass in an alternative main css file")
 parameters.parse(process.argv);
 
+function srcPath(src) {
```
bwindels, line 71: can we name this `modulePath` or something referring more the fact that this is only for javascript files used as an es module (e.g. not the service worker atm). Also, can you add a comment saying as much?

bwindels (approved):
Looking great, thank you! Just one minor thing, but looks good to go otherwise.

scripts/build.mjs
DanilaFe: What about these lines of code over here? Seems like the service worker is also included via `srcPath`.

https://github.com/vector-im/hydrogen-web/blob/b6d0d5424191bae9d7df2acb52062d6fbf022380/scripts/build.mjs#L109-L110

Should this not be the case?

bwindels: Have discussed OOB, this doesn't matter too much either way, until we want to use TS in the sw, so is fine to stay as is.

PR #450, opened 2021-08-06
Add rendering for horizontal rules

DanilaFe:
Before this PR, messages with `<hr>` tags would not be rendered and cause a crash.


bwindels (approved):
thanks!

PR #453, opened 2021-08-15
[SSO] - [PR 4] - SSO/Token login functionality

MidhunSureshR:


src/domain/login/SSOLoginViewModel.js
```diff
@@ -0,0 +1,79 @@
...
+        this._loadViewModelSubscription = this.track(this._loadViewModel.disposableOn("change", () => {
+            if (!this._loadViewModel.loading) {
+                this._loadViewModelSubscription = this.disposeTracked(this._loadViewModelSubscription);
+            }
+            this.emitChange("isBusy");
+        }));
+    }
+
+    async startSSOLogin() {
+        await this.platform.settingsStorage.setString("homeserver", this._homeserver);
```
bwindels: can we scope the name of this key a bit more, something like `sso_ongoing_login_homeserver`?

src/domain/login/SSOLoginViewModel.js
```diff
@@ -0,0 +1,79 @@
...
+        const link = `${this._homeserver}/_matrix/client/r0/login/sso/redirect?redirectUrl=${this.urlCreator.createSSOCallbackURL()}`;
```
bwindels: This creation of the sso url should be able to go in `SSOLoginHelper` class (see other comment), passing in the redirectUrl as a parameter?

src/domain/login/SSOLoginViewModel.js
```diff
@@ -0,0 +1,79 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import {ViewModel} from "../ViewModel.js";
+import {SessionLoadViewModel} from "../SessionLoadViewModel.js";
+
+export class SSOLoginViewModel extends ViewModel{
```
bwindels: I'd suggest to split up this view model into a `StartSSOLoginViewModel` (which receives the loginOptions.sso) and `CompleteSSOLoginViewModel` (which receives the loginOptions.token and the login token) . These two things don't seem to have much in common.

src/domain/login/SSOLoginViewModel.js
```diff
@@ -0,0 +1,79 @@
...
+        const homeserver = await this.platform.settingsStorage.getString("homeserver");
+        const loginOptions = await this._sessionContainer.queryLogin(homeserver);
+        this._loadViewModelSubscription = this.disposeTracked(this._loadViewModelSubscription);
+        if (this._loadViewModel) {
+            this._loadViewModel = this.disposeTracked(this._loadViewModel);
+        }
+        this._loadViewModel = this.track(new SessionLoadViewModel(this.childOptions({
+            createAndStartSessionContainer: async () => {
+                if (loginOptions.sso) {
+                    this._sessionContainer.startWithLogin(loginOptions.sso(this._loginToken));
```
bwindels: so this would actually be `loginOptions.token(this._loginToken)`

src/matrix/SessionContainer.js
```diff
@@ -97,25 +91,51 @@ export class SessionContainer {
...
+        implements LoginMethod
+        */
+        const flows = options.flows;
+        const result = {};
+        for (const flow of flows) {
+            if (flow.type === "m.login.password") {
+                result.password = (username, password) => new PasswordLoginMethod({homeServer, username, password});
+            }
+            else if (flow.type === "m.login.sso" && flows.find(flow => flow.type === "m.login.token")) {
+                result.sso = loginToken => new TokenLoginMethod({homeServer, loginToken});
```
bwindels: I think here we should rather put a class to help create the sso redirect url. SSO piggy-backs on token login, and the only specific part is redirecting to the homeserver url. This should be handled by an `SSOLoginHelper` class (rather than `SSOLoginMethod` as it wouldn't have the `login` method given CompleteSSOLoginViewModel would use the token login method.), and `SSOLoginHelper` is also where we can add the support of the multiple SSO providers.

src/matrix/SessionContainer.js
```diff
@@ -42,14 +44,6 @@ export const LoginFailure = createEnum(
     "Unknown",
 );
 
-function normalizeHomeserver(homeServer) {
```
bwindels, line 45: Why did you move this to the view model? Ideally, validation is done at the lowest possible level, which is the session container here.

src/domain/RootViewModel.js
```diff
@@ -35,12 +35,15 @@ export class RootViewModel extends ViewModel {
...
         this.track(this.navigation.observe("login").subscribe(() => this._applyNavigation()));
         this.track(this.navigation.observe("session").subscribe(() => this._applyNavigation()));
+        this.track(this.navigation.observe("sso").subscribe(() => this._applyNavigation()));
         this._applyNavigation(true);
     }
 
     async _applyNavigation(shouldRestoreLastUrl) {
         const isLogin = this.navigation.observe("login").get();
         const sessionId = this.navigation.observe("session").get();
+        // TODO: why not observe?
```
bwindels: good point, not sure why we're doing `observe().get()` rather than just `get()` above ...

src/matrix/net/HomeServerApi.js
```diff
@@ -130,6 +130,10 @@ export class HomeServerApi {
             {}, {}, options);
     }
 
+    queryLogin() {
```
bwindels: I'd suggest `getLoginFlows` here

src/matrix/SessionContainer.js
```diff
@@ -97,25 +91,51 @@ export class SessionContainer {
...
+                result.sso = loginToken => new TokenLoginMethod({homeServer, loginToken});
+            }
+            else if (flow.type === "m.login.token") {
+                result.token = loginToken => new TokenLoginMethod({homeServer, loginToken});
+            }
+        }
+        return result;
+    }
+
+    async queryLogin(homeServer) {
```
bwindels, line 124: we can normalize the homeserver here (and later on we can also add .well-known lookup here)

src/domain/navigation/index.js
```diff
@@ -223,6 +228,12 @@ export function tests() {
             const urlPath = stringifyPath(path);
             assert.equal(urlPath, "/session/1/rooms/a,b,c/1/details");
         },
+        "Parse loginToken query parameter into SSO segment": assert => {
+            const segments = parseUrlPath("/?loginToken=a1232aSD123");
```
bwindels: wouldn't location.search start with a `?` rather than a `/`? just to test it in the way it is used ...

src/domain/login/SSOLoginViewModel.js
```diff
@@ -0,0 +1,79 @@
...
+        }
+        const homeserver = await this.platform.settingsStorage.getString("homeserver");
+        const loginOptions = await this._sessionContainer.queryLogin(homeserver);
+        this._loadViewModelSubscription = this.disposeTracked(this._loadViewModelSubscription);
+        if (this._loadViewModel) {
+            this._loadViewModel = this.disposeTracked(this._loadViewModel);
+        }
+        this._loadViewModel = this.track(new SessionLoadViewModel(this.childOptions({
+            createAndStartSessionContainer: async () => {
+                if (loginOptions.sso) {
```
bwindels: probably best to check the correct login method (which here will be token rather than sso as mentioned below) before creating the SessionLoadViewModel or the UI will just be stuck at a white screen.

src/domain/login/LoginViewModel.js
```diff
@@ -0,0 +1,105 @@
...
+            const defaultHomeServer = normalizeHomeserver(this._defaultHomeServer);
+            await this.queryLogin(defaultHomeServer);
+            this._showPasswordLogin();
+            this._showSSOLogin(defaultHomeServer);
+        }
+    }
+
+    _showPasswordLogin() {
+        this._passwordLoginViewModel = new PasswordLoginViewModel(this.childOptions({defaultHomeServer: this._defaultHomeServer}));
+        const observable = this._passwordLoginViewModel.homeserverObservable;
```
bwindels: It's probably worth it to extract the homeserver field from the password login view (model) and put it in the login view (model). Some servers won't support password login at all, but you still need to be able to change the homeserver.

src/domain/RootViewModel.js
```diff
@@ -94,7 +103,7 @@ export class RootViewModel extends ViewModel {
         }
     }
 
-    _showLogin() {
+    _showLogin(options) {
```
bwindels: can we make this loginToken, to be more explicit, and also to make the typescript conversion easier.

src/domain/login/LoginViewModel.js
```diff
@@ -0,0 +1,105 @@
...
+        this._start();
+    }
+
+    get passwordLoginViewModel() { return this._passwordLoginViewModel; }
+    get ssoLoginViewModel() { return this._ssoLoginViewModel; }
+    get loadViewModel() {return this._loadViewModel; }
+
+    async _start() {
+        if (this._loginToken) {
+            this._ssoLoginViewModel = this.track(new SSOLoginViewModel(this.childOptions({loginToken: this._loginToken})));
```
bwindels: no dynamic properties, please. Ideally declare properties in the constructor, or in a method called by the constructor, but not conditionally like here.

bwindels: Here, you'd actually have two members: one for StartSSOLoginViewModel, and one for CompleteSSOLoginViewModel

bwindels (commented):
This looks good, and works well from a first try. Did my first review, didn't get all the way through yet though.

src/domain/login/CompleteSSOLoginViewModel.js
```diff
@@ -70,10 +64,4 @@ export class SSOLoginViewModel extends ViewModel{
             this.emitChange("isBusy");
```
bwindels: `isBusy` is not a property on this VM

src/domain/SessionLoadViewModel.js
```diff
@@ -38,7 +38,7 @@ export class SessionLoadViewModel extends ViewModel {
         try {
             this._loading = true;
             this.emitChange("loading");
-            this._sessionContainer = this._createAndStartSessionContainer();
+            this._sessionContainer = await this._createAndStartSessionContainer();
```
bwindels: We should probably already make `loadLabel` return something here while awaiting... perhaps a constructor argument called `initialLabel` which we can then set to `Finishing your SSO login...`.

src/domain/login/LoginViewModel.js
```diff
@@ -0,0 +1,125 @@
...
+        this._createSessionContainer = createSessionContainer;
+        this._ready = ready;
+        this._defaultHomeServer = defaultHomeServer;
+        this._loginToken = loginToken;
+        this._sessionContainer = this._createSessionContainer();
+        this._loginOptions = null;
+        this._passwordLoginViewModel = null;
+        this._startSSOLoginViewModel = null;
+        this._completeSSOLoginViewModel = null;
+        this._homeserver = null;
```
bwindels: can't we initialize this to the default homeserver rather than have a separate field?

src/domain/login/LoginViewModel.js
```diff
@@ -0,0 +1,125 @@
...
+        this._start(this._defaultHomeServer);
+    }
+
+    get passwordLoginViewModel() { return this._passwordLoginViewModel; }
+    get startSSOLoginViewModel() { return this._startSSOLoginViewModel; }
+    get completeSSOLoginViewModel(){ return this._completeSSOLoginViewModel; }
+    get defaultHomeServer() { return this._defaultHomeServer; }
+    get errorMessage() { return this._errorMessage; }
+
+    async _start(homeserver) {
```
bwindels: call this something more specific like `_loadServerOptions` given it is not only used after the object is created?

src/domain/login/LoginViewModel.js
```diff
@@ -0,0 +1,125 @@
...
+        this._startSSOLoginViewModel = this.track(new StartSSOLoginViewModel(this.childOptions()));
+        this.emitChange("startSSOLoginViewModel");
+    }
+
+    _showError(message) {
+        this._errorMessage = message;
+        this.emitChange("errorMessage");
+    }
+
+    async queryLogin(homeserver) {
```
bwindels: Inline this method as it doesn't really do anything and is only used once?

src/platform/web/ui/login/LoginView.js
```diff
@@ -16,61 +16,43 @@ limitations under the License.
...
-                            type: "submit"
-                        }, vm.i18n`Log In`),
-                    ]),
-                ]),
-                // use t.mapView rather than t.if to create a new view when the view model changes too
-                t.p(hydrogenGithubLink(t))
-            ])
+            t.h1([vm.i18n`Sign In`]),
+            t.mapView(vm => vm.completeSSOLoginViewModel, vm => vm ? new CompleteSSOView(vm) : null),
+            t.if(vm => !vm.completeSSOLoginViewModel,
```
bwindels: I'd prefer to keep this logic out of the view and rather add a `showHomeserver` field to the viewmodel

src/matrix/login/SSOLoginHelper.js
```diff
@@ -0,0 +1,25 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+export class SSOLoginHelper{
+    constructor(homeserver) {
+        this._homeserver = homeserver;
+    }
+
+    ssoEndpointLink(redirectURL) {
```
bwindels: `createSSORedirectURL` and call `redirectURL` `returnURL` instead (even though the url param is called the former)?

src/domain/login/PasswordLoginViewModel.js
```diff
@@ -41,22 +40,23 @@ export class LoginViewModel extends ViewModel {
         }
     }
 
-    async login(username, password, homeserver) {
+    async login(username, password) {
+        const homeserver = this._homeserver;
+        if (!this._loginOptions.password) {
+            const path = this.navigation.pathFrom([this.navigation.segment("session")]);
```
bwindels: This is already checked by the `LoginViewModel`?

src/matrix/SessionContainer.js
```diff
@@ -97,25 +100,52 @@ export class SessionContainer {
         });
     }
 
-    async startWithLogin(homeServer, username, password) {
+    parseLoginOptions(options, homeServer) {
```
bwindels: _ for private

src/domain/login/StartSSOLoginViewModel.js
```diff
@@ -0,0 +1,32 @@
...
+export class StartSSOLoginViewModel extends ViewModel{
+    constructor(options) {
+        super(options);
+        const {loginOptions, homeserver} = options;
+        this._sso = loginOptions.sso;
+        this._homeserver = homeserver;
+    }
+
+    async startSSOLogin() {
+        await this.platform.settingsStorage.setString("sso_ongoing_login_homeserver", this._homeserver);
```
bwindels: you can use `this._sso.homeServer` here so you don't need to pass in the homeserver separately.

src/domain/login/LoginViewModel.js
```diff
@@ -0,0 +1,125 @@
...
+                } 
+            }
+            else {
+                this._showError("Could not query login methods supported by the homeserver");
+            }
+        }
+    }
+
+    _showPasswordLogin() {
+        this._passwordLoginViewModel = this.track(new PasswordLoginViewModel(this.childOptions()));
```
bwindels: I personally wouldn't override childOptions here and just pass in the login option need, like so:
```js
this._passwordLoginViewModel = this.track(new PasswordLoginViewModel(this.childOptions({
  createPasswordLogin: this._loginOptions.password})));
```
Same for SSO below

src/domain/login/LoginViewModel.js
```diff
@@ -0,0 +1,125 @@
...
+        this.emitChange("disposeViewModels");
+    }
+
+    updateHomeServer(newHomeserver) {
+        this._homeserver = newHomeserver;
+        this._disposeViewModels();
+        this._start(newHomeserver);
+    }
+
+    childOptions(options = {}) {
```
bwindels: As said above, I think the code could be simpler without overriding childOptions here, with the added benefit of just passing everything needed to the child view models.

src/platform/web/ui/login/PasswordLoginView.js
```diff
@@ -0,0 +1,61 @@
...
+                }
+            }, [
+                t.div({ className: "form-row" }, [t.label({ for: "username" }, vm.i18n`Username`), username]),
+                t.div({ className: "form-row" }, [t.label({ for: "password" }, vm.i18n`Password`), password]),
+                t.mapView(vm => vm.loadViewModel, loadViewModel => loadViewModel ? new SessionLoadStatusView(loadViewModel) : null),
+                t.div({ className: "button-row" }, [
+                    t.a({
+                        className: "button-action secondary",
+                        href: vm.cancelUrl
+                    }, [vm.i18n`Go Back`]),
```
bwindels: Perhaps the go back button should also go into the login view (model), otherwise it looks a bit weird with the sso button below ... perhaps as a chevron on the top left corner?

src/platform/web/ui/login/LoginView.js
```diff
@@ -16,61 +16,43 @@ limitations under the License.
...
-        });
         const homeserver = t.input({
-            id: "homeserver",
-            type: "text",
-            placeholder: vm.i18n`Your matrix homeserver`,
-            value: vm.defaultHomeServer, 
-            disabled
-        });
-        
+                id: "homeserver",
```
bwindels: Perhaps you can only construct this element in the if binding below where it is used?

src/domain/login/LoginViewModel.js
```diff
@@ -0,0 +1,125 @@
...
+        return {
+            ...super.childOptions(options),
+            ready: sessionContainer => {
```
bwindels: so, as discussed, I think it would be worthwhile to move the SessionLoadViewModel up to avoid too many nested view models and code duplication.

I would pass to the child view models a
`attemptLogin(method: LoginMethod): Promise<LoginFailure>` callback that they can await and get a LoginFailure back. This way:
 - child view models can show login errors on their own UI as they see fit
 - child views can still disable their input while attempting the login
 - child views don't have to deal with the SessionLoadViewModel
 - the login view model can also observe the progress of the login and for example disable the homeserver field

So, concretely, once the child view model calls `attemptLogin`, the login view model passes it to startWithLogin on the session container it already created, and wait for the the loadStatus to go from `Login` to something else. If it's a login failure, return that, otherwise return null (meaning the login succeeded). Furthermore, if the login succeeded, it should create a SessionLoadViewModel with the session container passed in. SessionLoadViewModel can now be just about loading the session, and we don't handle the Login(Failure) in there anymore.
Once login succeeded, and we start loading the session, the login options and homeserver field should also be hidden, with only the session load status visible.

Note that `StartSSOLoginViewModel` won't need `attemptLogin`, but `CompleteSSOLoginViewModel` will. The latter will now also also disappear once the token login succeeded, preventing both the "Finishing your sso login" and the label from the SessionLoadViewModel being shown at the same time.

bwindels (commented):
Thanks for the changes! Here's another preliminary review of those, with some proposed changes we already discussed OOB. Thanks again!

src/domain/login/LoginViewModel.js
bwindels: actually feel free to pass in just the whole loginOptions object (and the attemptLogin calback as described below)

src/platform/web/ui/login/LoginView.js
```diff
@@ -16,61 +16,49 @@ limitations under the License.
...
-        const homeserver = t.input({
-            id: "homeserver",
-            type: "text",
-            placeholder: vm.i18n`Your matrix homeserver`,
-            value: vm.defaultHomeServer, 
-            disabled
-        });
-        
         return t.div({className: "PreSessionScreen"}, [
+            t.a({className: "button-utility LoginView_back", href: vm.cancelUrl}),
```
MidhunSureshR: This should probably be a button so that we can disable it during a login.

src/domain/login/PasswordLoginViewModel.js
```diff
@@ -0,0 +1,60 @@
...
+        this._showError = showError;
+        this._homeserver = homeserver;
+        this._isBusy = false;
+    }
+
+    get isBusy() { return this._isBusy; }
+
+    _toggleBusy(state) {
+        this._isBusy = state;
+        this.emitChange("isBusy");
```
MidhunSureshR, line 35: It might be better if the `LoginViewModel` indicates to all child view(models) that a login is being attempted. This would make it easy, for example, to disable the sso login button when a password login is being attempted.

src/domain/login/LoginViewModel.js
```diff
@@ -0,0 +1,187 @@
...
+        const loadStatus = this._sessionContainer.loadStatus;
+        const handle = loadStatus.waitFor(status => status !== LoadStatus.Login);
+        await handle.promise;
+        this._toggleBusy(false);
+        const status = loadStatus.get();
+        if (status === LoadStatus.LoginFailed) {
+            return this._sessionContainer.loginFailure;
+        }
+        this._hideHomeserver = true;
+        this._disposeViewModels();
```
MidhunSureshR, line 138: Is it okay to dispose the viewmodels when said viewmodels are awaiting on this method?

src/domain/login/LoginViewModel.js
```diff
@@ -25,22 +25,21 @@ export class LoginViewModel extends ViewModel {
...
         this._errorMessage = "";
-        this._start(this._defaultHomeServer);
+        this._start(this._homeserver);
     }
 
     get passwordLoginViewModel() { return this._passwordLoginViewModel; }
     get startSSOLoginViewModel() { return this._startSSOLoginViewModel; }
     get completeSSOLoginViewModel(){ return this._completeSSOLoginViewModel; }
-    get defaultHomeServer() { return this._defaultHomeServer; }
+    get defaultHomeServer() { return this._homeserver; }
```
bwindels: I think we can also get rid of this getter and just use `vm.homeserver` in the view?

src/domain/login/CompleteSSOLoginViewModel.js
```diff
@@ -46,25 +46,22 @@ export class CompleteSSOLoginViewModel extends ViewModel {
...
             return;
         }
-        this._loadViewModelSubscription = this.disposeTracked(this._loadViewModelSubscription);
-        if (this._loadViewModel) {
-            this._loadViewModel = this.disposeTracked(this._loadViewModel);
+        const status = await this._attemptLogin(loginOptions.token(this._loginToken));
+        let error = "";
+        switch (status) {
+            case LoginFailure.Credentials:
+                error = `Your login-token is invalid.`;
```
bwindels: this.i18n`` please for future-proofness

src/domain/login/CompleteSSOLoginViewModel.js
```diff
@@ -46,25 +46,22 @@ export class CompleteSSOLoginViewModel extends ViewModel {
...
+            case LoginFailure.Connection:
+                error = `Can't connect to ${this._homeserver}.`;
+                break;
+            case LoginFailure.Unknown:
+                error = `Something went wrong while checking your login-token.`;
+                break;
+        }
+        if (error) {
+            this._showError(error);
+            this._sessionContainer.resetStatus();
```
bwindels: can we do this internally in the session container when calling `startWith...`? Then we also don't need to inject the sessionContainer here.

src/domain/login/LoginViewModel.js
```diff
@@ -80,14 +89,56 @@ export class LoginViewModel extends ViewModel {
...
+        }
+        this._hideHomeserver = true;
+        this._disposeViewModels();
+        this._createLoadViewModel();
+        return null;
+    }
+
+    _createLoadViewModel() {
+        this._loadViewModelSubscription = this.disposeTracked(this._loadViewModelSubscription);
+        if (this._loadViewModel) {
```
bwindels: you don't need the if here, disposeTracked ignores calls with null or undefined

src/domain/login/LoginViewModel.js
```diff
@@ -80,14 +89,56 @@ export class LoginViewModel extends ViewModel {
...
+        }
+        this._loadViewModel = this.track(new SessionLoadViewModel(this.childOptions()));
+        this._loadViewModel.start();
+        this.emitChange("loadViewModel");
+        this._loadViewModelSubscription = this.track(
+            this._loadViewModel.disposableOn("change", () => {
+                if (!this._loadViewModel.loading) {
+                    this._loadViewModelSubscription = this.disposeTracked(this._loadViewModelSubscription);
+                }
+                this.emitChange("isBusy");
```
bwindels: Don't you need to set _isBusy to false here first? This matters when the loading fails for some reason, that you don't want to keep everything disabled.

src/domain/login/PasswordLoginViewModel.js
bwindels: Yeah, perhaps we can add a setBusy to PasswordLoginViewModel and StartSSOLoginViewModel called by the LoginViewModel? That way, we also don't need to toggle busy while calling attemptLogin.

src/domain/login/PasswordLoginViewModel.js
```diff
@@ -15,51 +15,47 @@ limitations under the License.
...
-        this._loadViewModelSubscription = this.disposeTracked(this._loadViewModelSubscription);
-        if (this._loadViewModel) {
-            this._loadViewModel = this.disposeTracked(this._loadViewModel);
+        this._toggleBusy(true);
+        const status = await this._attemptLogin(this._loginOptions.password(username, password));
+        this._toggleBusy(false);
+        let error = "";
+        switch (status) {
+            case LoginFailure.Credentials:
+                error = `Your credentials don't seem to be correct.`;
```
bwindels: Can we go back to the previous text here?

src/matrix/SessionContainer.js
```diff
@@ -349,4 +353,10 @@ export class SessionContainer {
             this._sessionId = null;
         }
     }
+
+    resetStatus() {
+        this._status.set(LoadStatus.NotLoading);
```
bwindels, line 362: as said elsewhere, can make this private and call this from startWithLogin/ExistingSession?

src/domain/login/CompleteSSOLoginViewModel.js
bwindels: also, no dash here?

src/domain/login/PasswordLoginViewModel.js
```diff
@@ -20,8 +20,7 @@ import {LoginFailure} from "../../matrix/SessionContainer.js";
 export class PasswordLoginViewModel extends ViewModel {
     constructor(options) {
         super(options);
-        const {ready, loginOptions, sessionContainer, homeserver, attemptLogin, showError} = options;
-        this._ready = ready;
+        const {loginOptions, sessionContainer, homeserver, attemptLogin, showError} = options;
```
bwindels: I would the view (model) for every login method have it's own errors shown:
image

So the error is closer to the invalid input

src/platform/web/ui/login/LoginView.js
bwindels: It would probably be a feature to be able to cancel login like this, but SessionContainer.dispose currently doesn't abort any login requests in progress, so it would probably throw if you go back and a long running /login request finishes. So yeah, fine to disable for now.

src/platform/web/ui/login/LoginView.js
```diff
@@ -16,61 +16,49 @@ limitations under the License.
...
+                        value: vm.defaultHomeServer,
+                        disabled: vm => vm.isBusy,
+                        onChange: event => vm.updateHomeServer(event.target.value),
+                    })
+                ]
+            )),
+            t.mapView(vm => vm.passwordLoginViewModel, vm => vm ? new PasswordLoginView(vm): null),
+            t.if(vm => vm.passwordLoginViewModel && vm.startSSOLoginViewModel, t => t.p({className: "LoginView_separator"}, vm.i18n`or`)),
+            t.mapView(vm => vm.startSSOLoginViewModel, vm => vm ? new StartSSOLoginView(vm) : null),
+            t.if(vm => vm.errorMessage, (t, vm) => t.p({className: "LoginView_error"}, vm.i18n(vm.errorMessage))),
```
bwindels: You can add the error class for redness and boldness:
```suggestion
            t.if(vm => vm.errorMessage, (t, vm) => t.p({className: "LoginView_error error"}, vm.i18n(vm.errorMessage))),
```

src/platform/web/ui/login/LoginView.js
```diff
@@ -16,61 +16,49 @@ limitations under the License.
...
+                        id: "homeserver",
+                        type: "text",
+                        placeholder: vm.i18n`Your matrix homeserver`,
+                        value: vm.defaultHomeServer,
+                        disabled: vm => vm.isBusy,
+                        onChange: event => vm.updateHomeServer(event.target.value),
+                    })
+                ]
+            )),
+            t.mapView(vm => vm.passwordLoginViewModel, vm => vm ? new PasswordLoginView(vm): null),
```
bwindels, line 52: Can we have a spinner with a message while it's querying the login methods on the homeserver?

src/platform/web/ui/login/LoginView.js
bwindels: Perhaps put this at the top as well, before the homeserver field?

src/domain/RootViewModel.js
bwindels: feel free to remove this comment and convert the others above to just use get without observe

src/domain/SessionLoadViewModel.js
```diff
@@ -111,20 +110,6 @@ export class SessionLoadViewModel extends ViewModel {
...
-                case LoadStatus.LoginFailed:
-                    switch (sc.loginFailure) {
-                        case LoginFailure.LoginFailure:
-                            return `Your username and/or password don't seem to be correct.`;
-                        case LoginFailure.Connection:
-                            return `Can't connect to ${this._homeserver}.`;
-                        case LoginFailure.Unknown:
-                            return `Something went wrong while checking your login and password.`;
-                    }
-                    break;
```
bwindels, line 127: can we perhaps have a comment here that the statuses related to login are handled in the login view model

src/domain/login/PasswordLoginViewModel.js
```diff
@@ -0,0 +1,60 @@
...
+        this._toggleBusy(true);
+        const status = await this._attemptLogin(this._loginOptions.password(username, password));
+        this._toggleBusy(false);
+        let error = "";
+        switch (status) {
+            case LoginFailure.Credentials:
+                error = `Your credentials don't seem to be correct.`;
+                break;
+            case LoginFailure.Connection:
+                error = `Can't connect to ${this._homeserver}.`;
```
bwindels: You don't need to inject the homeserver separately I think. When you create the password login option, you can get the homeserver from there.

src/domain/login/LoginViewModel.js
```diff
@@ -0,0 +1,187 @@
...
+        this._sessionContainer.startWithLogin(loginMethod);
+        const loadStatus = this._sessionContainer.loadStatus;
+        const handle = loadStatus.waitFor(status => status !== LoadStatus.Login);
+        await handle.promise;
+        this._toggleBusy(false);
+        const status = loadStatus.get();
+        if (status === LoadStatus.LoginFailed) {
+            return this._sessionContainer.loginFailure;
+        }
+        this._hideHomeserver = true;
```
bwindels, line 136: emitChange after this?

src/domain/login/PasswordLoginViewModel.js
```diff
@@ -0,0 +1,60 @@
...
+                break;
+            case LoginFailure.Unknown:
+                error = `Something went wrong while checking your credentials.`;
+                break;
+        }
+        if (error) {
+            this._showError(error);
+            this._sessionContainer.resetStatus();
```
bwindels: I think if this is done implicitly by startWithLogin as suggested elsewhere, you don't need to inject the sessionContainer here anymore.

src/domain/RootViewModel.js
```diff
@@ -65,7 +68,13 @@ export class RootViewModel extends ViewModel {
                     this._showSessionLoader(sessionId);
                 }
             }
-        } else {
+        } else if (ssoSegment) {
+            this.urlCreator.normalizeUrl();
+            if (this.activeSection !== "login") {
+                this._showLogin({loginToken: ssoSegment.value});
```
bwindels: I think we discussed making this just the loginToken rather than an object before. Any reason to keep it this way?

src/domain/login/LoginViewModel.js
bwindels: Hmm, yeah, it's not ideal but as the callee should not do anything if there is no error, I think it's probably fine in practice. I also don't see a better way.

bwindels (commented):
Thanks for the changes! We're getting very close, just another set of comments. Thanks for bearing with me :sweat_smile:

src/domain/login/CompleteSSOLoginViewModel.js
```diff
@@ -48,13 +48,13 @@ export class CompleteSSOLoginViewModel extends ViewModel {
         let error = "";
         switch (status) {
             case LoginFailure.Credentials:
-                error = `Your login-token is invalid.`;
+                error = this.i18n`Your logintoken is invalid.`;
```
bwindels: Ah, heh, I just meant two words (login token) :)

src/matrix/SessionContainer.js
```diff
@@ -129,9 +129,11 @@ export class SessionContainer {
     }
 
     async startWithLogin(loginMethod) {
-        if (this._status.get() !== LoadStatus.NotLoading) {
+        const currentStatus = this._status.get();
+        if (currentStatus !== LoadStatus.LoginFailed && currentStatus !== LoadStatus.NotLoading) {
```
bwindels: Should we allow LoadStatus.Error here as well?

src/domain/login/LoginViewModel.js
```diff
@@ -117,11 +117,15 @@ export class LoginViewModel extends ViewModel {
 
     async attemptLogin(loginMethod) {
         this._toggleBusy(true);
+        this._passwordLoginViewModel?.toggleBusy(true);
```
bwindels: Why not put these two lines inside the method `_toggleBusy`? Also, to me, `toggle` rather than `set` implies it wouldn't take an argument but rather just invert the current state.

src/domain/login/CompleteSSOLoginViewModel.js
```diff
@@ -24,15 +24,22 @@ export class CompleteSSOLoginViewModel extends ViewModel {
...
+        this._errorMessage = "";
         this.performSSOLoginCompletion();
     }
 
+    get errorMessage() { return this._errorMessage; }
+
+    _showError(message) {
+        this._errorMessage = message;
+        this.emitChange("errorMessage");
+        this._errorMessage = "";
```
bwindels: Curious why you clear it again after emitting? If any other emit in the viewmodel triggers a binding update, the error would just disappear at an unrelated timed. I'd expect to clear it only when you try to login again ...

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -161,6 +161,11 @@ a.button-action {
     background-image: url('icons/disable-grid.svg');
 }
 
+.error {
+    color: #f00;
```
bwindels: This is already present on line 633 in this same file.

src/domain/login/LoginViewModel.js
```diff
@@ -49,11 +49,15 @@ export class LoginViewModel extends ViewModel {
...
     get errorMessage() { return this._errorMessage; }
     get showHomeserver() { return !this._hideHomeserver; }
-    get cancelUrl() { return this.urlCreator.urlForSegment("session"); }
     get loadViewModel() {return this._loadViewModel; }
     get isBusy() { return this._isBusy; }
     get isFetchingLoginOptions() { return this._isFetchingLoginOptions; }
 
+    goBack() {
+        const path = this.navigation.pathFrom([this.navigation.segment("session")]);
+        this.navigation.applyPath(path);
```
bwindels: you could just do `this.navigation.push("session")` here fwiw.

bwindels (commented):
Thanks for the changes, looking great! Just a few final nits, and we're there. Have tried it an it works great :+1:

src/matrix/SessionContainer.js
MidhunSureshR: yes 👍

src/domain/login/CompleteSSOLoginViewModel.js
MidhunSureshR: Yeah that wasn't very smart of me 🙂

Fixed for LoginViewModel and PasswordLoginViewModel but for CompleteSSOLoginViewModel I've removed `this._errorMessage = ""` since there's no UI currently for retrying a failed login using loginToken.

src/domain/login/CompleteSSOLoginViewModel.js
```diff
@@ -0,0 +1,70 @@
...
+        this._errorMessage = message;
+        this.emitChange("errorMessage");
+    }
+
+    async performSSOLoginCompletion() {
+        if (!this._loginToken) {
+            return;
+        }
+        const homeserver = await this.platform.settingsStorage.getString("sso_ongoing_login_homeserver");
+        const loginOptions = await this._sessionContainer.queryLogin(homeserver);
```
bwindels: Hmm, this should probably have a try catch around it catching any error and calling showError

src/domain/login/CompleteSSOLoginViewModel.js
```diff
@@ -0,0 +1,70 @@
...
+        if (!loginOptions.token) {
+            const path = this.navigation.pathFrom([this.navigation.segment("session")]);
+            this.navigation.applyPath(path);
```
bwindels: could also do just `this.navigation.push("session")` here

bwindels (approved):
This is good to go!

PR #456, opened 2021-08-17
Snowpack  + Typescript conversion (Part 1)

DanilaFe:


src/matrix/storage/common.ts
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-export const STORE_NAMES = Object.freeze([
+export const STORE_NAMES: Readonly<string[]> = Object.freeze([
```
bwindels: Could this easily be a [string `enum`](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums)?

src/matrix/storage/idb/utils.ts
```diff
@@ -112,33 +121,41 @@ export function txnAsPromise(txn) {
...
-            const cursor = event.target.result;
+            const cursor = (event.target as IDBRequest<I>).result;
             if (!cursor) {
                 resolve(false);
+                // @ts-ignore
                 needsSyncPromise && Promise._flush && Promise._flush();
                 return; // end of results
             }
-            const result = processValue(cursor.value, cursor.key, cursor);
+            const result = processValue(cursor["value"], cursor.key, cursor);
```
bwindels, line 173: Did you consider this? Find it a bit more clear why we need to jump through hoops ...
```suggestion
            const result = processValue((cursor as IDBCursorWithValue).value, cursor.key, cursor);
```

src/matrix/storage/idb/utils.ts
```diff
@@ -112,33 +121,41 @@ export function txnAsPromise(txn) {
...
+            // @ts-ignore
             needsSyncPromise && Promise._flush && Promise._flush();
         });
     });
 }
 
-export function iterateCursor(cursorRequest, processValue) {
+type CursorIterator<T, I extends IDBCursor> = I extends IDBCursorWithValue ?
+    (value: T, key: IDBValidKey, cursor: IDBCursorWithValue) => { done: boolean, jumpTo?: IDBValidKey } :
+    (value: undefined, key: IDBValidKey, cursor: IDBCursor) => { done: boolean, jumpTo?: IDBValidKey }
```
bwindels: wow, what a whopper :)

src/matrix/storage/idb/utils.ts
```diff
@@ -206,12 +206,12 @@ export async function findStoreValue<T>(db: IDBDatabase, storeName: string, toCu
     const store = tx.objectStore(storeName);
     const cursor = await reqAsPromise(toCursor(store));
     let match;
-    // @ts-ignore
     const matched = await iterateCursor<T>(cursor, (value) => {
         if (matchesValue(value)) {
             match = value;
-            return true;
```
bwindels, line 198: heh, good spot. `findStoreValue` actually doesn't seem to be used anywhere, so this should have no effect otherwise.

src/matrix/storage/idb/error.ts
```diff
@@ -34,22 +37,22 @@ export class IDBError extends StorageError {
         if (cause) {
             fullMessage += cause.message;
         }
-        super(fullMessage, cause);
+        super(fullMessage, cause || undefined);
```
bwindels: hmm, wonder why this is needed?

src/matrix/storage/idb/QueryTarget.ts
```diff
@@ -14,14 +14,29 @@ See the License for the specific language governing permissions and
...
+import {iterateCursor, DONE, NOT_DONE, reqAsPromise} from "./utils";
 
-export class QueryTarget {
-    constructor(target) {
+type Reducer<A,B> = (acc: B, val: A) => B
+
+export type IDBQuery = IDBValidKey | IDBKeyRange | undefined | null
+
+interface QueryTargetInterface<T> {
+    openCursor: (range?: IDBQuery, direction?: IDBCursorDirection | undefined) => IDBRequest<IDBCursorWithValue | null>;
```
bwindels: any reason to use the lambda syntax here rather than the method syntax?

bwindels (commented):
didn't get to the end of the PR, here are my comments so far

src/matrix/storage/idb/error.ts
DanilaFe: Because `cause` can be `null`, but the argument expects something that can be `undefined`. This was the least invasive way of getting it to typecheck, I think.

src/matrix/storage/idb/QueryTarget.ts
DanilaFe: Because I didn't know about it 🥲

src/matrix/storage/idb/utils.ts
DanilaFe: I think that this may actually hurt intuition; it's not that we _know_ that the value is an `IDBCursorWithValue`, but that
we are okay with getting back an `undefined` if `cursor` is an `IDBCursor`.

src/matrix/storage/idb/error.ts
```diff
@@ -18,7 +18,10 @@ limitations under the License.
 import { StorageError } from "../common";
 
 export class IDBError extends StorageError {
-    constructor(message, source, cause) {
+    storeName: string;
+    databaseName: string;
+
+    constructor(message: string, source, cause: DOMException | null) {
```
bwindels: I think source can be typed as `IDBObjectStore` here.

src/matrix/storage/idb/error.ts
bwindels: hmm, we should probably decide on one single way to have optional arguments (if not everywhere, at least for (error) constructors), so we don't need conversions like this. Ideally we would use `param?: Type` but a lot of code checks for `null` I think, so perhaps we start out with `param: Type | null = null` which will convert undefined to null and also make the parameter optional. See this [playground snippet](https://www.typescriptlang.org/play?#code/C4TwDgpgBA8mwEsD2A7AhgGwDwBUB8UAvFDlAD5QoCuGGA3AFAMBmVKAxoqlM0kgBRoAXJSoBbAEYQATgBooEkdUkzyo2kXUYAlEvFTpUAN4MoZqO1QBnJBggA6DEgDm-Cdsbmo0iMCrSUBTRpQXlQSCRmBSJCYgAiZQM4qAB+aJE0DwYAXyZWDi5AiWDBPRU5BTKDXVFy41NzHz8AqDQoACoFRlyWNk5kQN4kAEZS2ur6ryb-Qb5QqDYAEwhmBBQIRaye-P7uIYAmMcSZbUnG3xmeObR5alotvL7Cq6QAZiP9E7OzaZahwQeLD4owArFkDvwAGzgvjvADsHiAA) as well. After the initial migration, we could then go from null to undefined to be able to use `param?: Type` in one sweep. Wdyt?

src/matrix/storage/idb/QueryTarget.ts
```diff
@@ -33,95 +48,99 @@ export class QueryTarget {
...
+                        keyPath = [keyPath];
+                    }
+                    return keyPath.reduce((obj, key) => obj[key], value);
                 }
             });
         }
     }
 
-    reduce(range, reducer, initialValue) {
+    reduce<B>(range: IDBQuery, reducer: Reducer<T,B>, initialValue: B): Promise<boolean> {
```
bwindels, line 75: heh, interesting ... the return type checks out but does not seem what you would expect from reduce (`B`). I imagine this is a bug, but as reduce and reduceReverse are not used anywhere, it didn't surface. Fine to leave like this until we use it though (or delete it).

src/matrix/storage/idb/Store.ts
```diff
@@ -0,0 +1,175 @@
...
+
+class QueryTargetWrapper<T> {
+    private _qt: IDBIndex | IDBObjectStore;
+
+    constructor(qt: IDBIndex | IDBObjectStore) {
+        this._qt = qt;
+    }
+
+    get keyPath(): string | string[] {
+        if (this._qt["objectStore"]) {
```
bwindels: So had a look, and the idiomatic way to narrow the type seems to be [`in` narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing) here, which preserves type-safety without casts. Also see [this playground example](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAQwM4GUpwE4FMAUAjlAFyICSAIgELlgAmuAHogD4U0DyARgFa7RMOXAEoyVaj36CseRAG8AUIhWIYwRPgBEcPgKhC8WtUmIiFy1VbxQQ2U1AB0u6Qdm4A3JZUBfb4hs7By8-RVBIWAQUDHcAJiJSDloGZjYkqX1DUXEuPRlhCyt1TWIAbR08t2EtAF1zJSsrYmdKrK8rP2tcW3tEYhDFIA).

src/matrix/storage/idb/utils.ts
bwindels: So had a look, and the idiomatic way to narrow the type seems to be [`in` narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing) here, which preserves type-safety without casts. Also see [this playground example](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAQwM4GUpwE4FMAUAjlAFyICSAIgELlgAmuAHogD4U0DyARgFa7RMOXAEoyVaj36CseRAG8AUIhWIYwRPgBEcPgKhC8WtUmIiFy1VbxQQ2U1AB0u6Qdm4A3JZUBfb4hs7By8-RVBIWAQUDHcAJiJSDloGZjYkqX1DUXEuPRlhCyt1TWIAbR08t2EtAF1zJSsrYmdKrK8rP2tcW3tEYhDFIA).

src/matrix/storage/idb/Store.ts
```diff
@@ -0,0 +1,175 @@
...
+            return (this._qt as IDBIndex).objectStore.keyPath;
+        } else {
+            return this._qt.keyPath;
+        }
+    }
+
+    get _qtStore(): IDBObjectStore {
+        return this._qt as IDBObjectStore;
```
bwindels: I doubt we use `add` or `put` on an IDBIndex currently, but perhaps we can actually do the right thing here like so:
```ts
get _qtStore(): IDBObjectStore {
    if ("objectStore" in this._qt) {
        return this._qt.objectStore;
    }
    return this._qt;
}
```

bwindels (commented):
Thanks for the changes, have finished my review and left some feedback on the ongoing comments.

src/matrix/storage/idb/error.ts
DanilaFe: Unfortunately not, since `IDBRequestError` feeds in an `IDBRequest` here. I figured that the source may be pretty much anything in the long run, and decided against typing it as `IDBObjectStore | IDBRequest` (though that seems possible if you think it's appropriate)

src/matrix/storage/idb/utils.ts
DanilaFe: Wouldn't this introduce a condition / branch here where there previously wasn't one? Especially since the function call does not at all depend on whether `value` is present. If that's fine, my initial thought was `"value" in cursor ? cursor.value : undefined`, but that actually doesn't compile. 

It's not just a problem with ternary conditionals: seems like narrowing doesn't work in this case:
2021-08-20-093609_1195x140_scrot

src/matrix/storage/idb/Store.ts
DanilaFe: Is that equivalent to the current behavior? I _think_ that currently the code just tries to call add/subtract and throws an exception if not, so unsafe casting seemed to be a similar approach in spirit. But if we don't use `add` or `put` on an `IDBIndex` (if we did, we'd be seeing exceptions :smile: ) , then it's probably safe to change the behavior of these functions.

src/matrix/storage/idb/utils.ts
DanilaFe: This is likely because `cursor` does not have a union type. I personally still think that this (admittedly convoluted) implementation of `iterateCursor` is nicer than the alternative:

We could remove the `I` generic here and just accept a function argument that accepts `T | undefined`. The problem is that in this case every call site of `iterateCursor` needs to provide a function / lambda where the argument may be `undefined`. This in turn requires lots of extra tiptoeing around non-existent `undefined`s. 

The current approach (and I'll add a comment for this since the type of `CursorIterator` is indeed a whopper) makes for very clean usage of the API: whenever we pass in an `IDBRequest<IDBCursor>`, we have to pass in a lambda that accepts `undefined` and `IDBCursor`; on the other hand, whenever we pass in an `IDBRequest<IDBCursorWithValue>`, we have to pass in a lambda that accepts `T` (and only `T`) and an `IDBCursorWithValue`. There's thus no need for unnecessary `undefined` checks.

DanilaFe: I wrote up a comment explaining this in 50b7a8a3fd09e879777be4d0989c0378fb8baee7.

src/matrix/storage/idb/error.ts
DanilaFe: I didn't know about this interaction with `undefined` and default arguments. It seems like this is indeed the best approach (if only because it cleans up the ugly `|| undefined` casts); I'll switch to using `| null` in `StorageError` and friends. I don't see anywhere that accesses `.cause` though, or checks for null in this instance.

src/matrix/storage/common.ts
DanilaFe: I think so, though as of this PR some parts that use `STORE_NAMES` and the hypothetical `StoreNames` `enum` are in JavaScript. I was able to achieve what appears to be the right behavior (without breaking any tests or experiencing any issues in running H) by changing `Storage.storeNames` to be a reference to `StoreNames` (and having type `typeof StoreNames`). I will try to see how this affects the "future" PRs.

src/matrix/storage/idb/error.ts
bwindels: Looks like `IDBRequestError` passes in `request.source` though? This should be of type `IDBIndex | IDBCursor | IDBObjectStore`. So, there is currently a bug where it sometimes says `<unknown store>.<unknown db>` in the error message, I'm guessing that either I messed up and I'm passing the wrong thing in somewhere, or browsers don't always set source.name and source.transaction.db.name. I'm guessing the former. I mainly added the `?` operator out of caution; to ensure the constructor does not crash and hide the real error. It would be good to add some type-safety here nonetheless, while keeping the caution.

While you're at it, at least passing in `IDBCursor` explains the bug, as it does not have a name, but it does have an `objectStore` property. Similarly, `IDBIndex` has an `objectStore` property, but also a `name`, which would be good to include in the error message in the case of an index, like `hydrogen_393920320.timelineEvents.byEventId` or something. So I'd have this constructor accept  `IDBIndex | IDBCursor | IDBObjectStore` for the `source` and then with instanceof narrowing determine in db and object store name for each case in a better way than currently to avoid the aforementioned bug.

src/matrix/storage/idb/Store.ts
bwindels: Seems safe to me indeed

src/matrix/storage/idb/Store.ts
```diff
@@ -35,7 +35,7 @@ class QueryTargetWrapper<T> {
     }
 
     get keyPath(): string | string[] {
-        if (this._qt["objectStore"]) {
+        if ("objectStore" in this._qt) {
             return (this._qt as IDBIndex).objectStore.keyPath;
```
bwindels: we should be able to drop this cast now

bwindels (commented):
thanks for the changes, I'll let you look at the remaining open comments when convenient

bwindels (approved):

PR #458, opened 2021-08-18
Opening member-details panel should not open/close grid view

### LINKED ISSUE #457: Opening member details panel opens/closes grid view

MidhunSureshR:
Fixes #457 

### DIFF
--- a/src/domain/navigation/index.js
+++ b/src/domain/navigation/index.js
@@ -132,9 +132,9 @@ export function parseUrlPath(urlPath, currentNavPath, defaultSessionId) {
                 segments.push(roomsSegmentWithRoom(rooms, roomId, currentNavPath));
             }
             segments.push(new Segment("room", roomId));
-            const partIndex = parts.findIndex(part => part === "open-room");
-            const partsAfterRoom = parts.slice(partIndex + 1);
-            if (partsAfterRoom.length === 1) {
+            const openRoomPartIndex = parts.findIndex(part => part === "open-room");
+            const hasOnlyRoomIdAfterPart = openRoomPartIndex >= parts.length - 2;
+            if (hasOnlyRoomIdAfterPart) {
                 // Copy right-panel segments from previous path only if there are no other parts after open-room
                 // fixes memberlist -> member details closing/opening grid view
                 const previousSegments = currentNavPath.segments;
--- a/src/domain/navigation/index.js
+++ b/src/domain/navigation/index.js
@@ -327,6 +327,28 @@ export function tests() {
             assert.equal(segments[4].type, "details");
             assert.equal(segments[4].value, true);
         },
+        "open-room action should only copy over previous segments if there are no parts after open-room": assert => {
+            const nav = new Navigation(allowsChild);
+            const path = nav.pathFrom([
+                new Segment("session", 1),
+                new Segment("rooms", ["a", "b", "c"]),
+                new Segment("room", "b"),
+                new Segment("right-panel", true),
+                new Segment("members", true)
+            ]);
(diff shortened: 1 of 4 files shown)


src/domain/navigation/index.js
```diff
@@ -132,11 +132,14 @@ export function parseUrlPath(urlPath, currentNavPath, defaultSessionId) {
                 segments.push(roomsSegmentWithRoom(rooms, roomId, currentNavPath));
             }
             segments.push(new Segment("room", roomId));
-            // Add right-panel segments from previous path
-            const previousSegments = currentNavPath.segments;
-            const i = previousSegments.findIndex(s => s.type === "right-panel");
-            if (i !== -1) {
-                segments.push(...previousSegments.slice(i));
+            if (!urlPath.includes("/member/")) {
```
bwindels: Hah, I completely misunderstood what you are doing here for some reason likely more related to me than to your code, to then look into the code wondering if we couldn't reuse the `open-room` action and come to exactly the same conclusion as you to then come back here and suggest that and go ... aaaaah, that's what's happening here :smile: 

Having said that, I wonder if we can make the check a bit more generic here by just checking if  there is only one element (the room id) after the `open-room` part in `parts`. Wdyt?

src/domain/session/rightpanel/MemberTileViewModel.js
```diff
@@ -47,7 +48,7 @@ export class MemberTileViewModel extends ViewModel {
     }
 
     get detailsUrl() {
-        return `${this.urlCreator.urlUntilSegment("room")}/member/${this._member.userId}`;
+        return `${this.urlCreator.openRoomActionUrl(this._roomId)}/member/${this._member.userId}`;
```
bwindels: you could also get the current room id from `this.navigation.path.get("room").value` so you don't need to inject it ...

bwindels (commented):
Took me a while to come to the same conclusion as you, be great approach :smile: perhaps we can make the check in the parseUrl function a bit more generic though.

src/domain/navigation/index.js
MidhunSureshR: Sounds good to me :+1:

src/domain/session/rightpanel/MemberTileViewModel.js
MidhunSureshR: Oops!🤦‍♂️

MidhunSureshR:
~~I should also write a test for the new logic in `index.js`...~~ Done

src/domain/navigation/index.js
```diff
@@ -132,8 +132,10 @@ export function parseUrlPath(urlPath, currentNavPath, defaultSessionId) {
                 segments.push(roomsSegmentWithRoom(rooms, roomId, currentNavPath));
             }
             segments.push(new Segment("room", roomId));
-            if (!urlPath.includes("/member/")) {
-                // Add right-panel segments from previous path if /member is not in url
+            const partIndex = parts.findIndex(part => part === "open-room");
+            const partsAfterRoom = parts.slice(partIndex + 1);
+            if (partsAfterRoom.length === 1) {
```
bwindels: Thanks for the changes! A bit of a nit, but I don't think we need to create an array here with slice, as we won't be using it. Can also do just:
```js
            const openRoomPartIndex = parts.findIndex(part => part === "open-room");
            // -2 for the room id after open-room and the open-room part itself
            const hasOnlyRoomIdAfterPart = openRoomPartIndex >= parts.length - 2;
            if (hasOnlyRoomIdAfterPart) {
```

MidhunSureshR: Done :heavy_check_mark:

bwindels (approved):
Thanks!

PR #471, opened 2021-08-26
Fix missing new joins preventing key shares

### LINKED ISSUE #470: New joins can be missed, preventing keys to be shared with them in e2ee rooms
If the homeserver sends a new join both in the state and timeline section of the sync response, Hydrogen will consider the event in the state section is not a membership change, and will rely on it being present in the timeline for it to trigger a membership change (if it's not and the room response is not limited, this must be a lazy-loaded member and we shouldn't consider it a join).

However, since the membership event in the state section is written to storage nonetheless (to gather as many member events as possible and avoid pulling in full member list), by the time we process the timeline membership event, it looks like we already know about this member and we don't consider it a new join either.

We'll only add the roomId to roomIds list on the userIdentity we share e2ee rooms with when a memberChange is considered a new join, and this is used to determine who we share keys with. So failing to consider it a new join will fail to share keys with that member.

We'll need a storage migration to fixup the problem in storage where it has occured before, comparing userIdentities with roomMembers.

bwindels:
Fixes #470

### DIFF
--- a/src/matrix/room/Room.js
+++ b/src/matrix/room/Room.js
@@ -120,7 +120,8 @@ export class Room extends BaseRoom {
             txn.roomMembers.removeAllForRoom(this.id);
         }
         const {entries: newEntries, updatedEntries, newLiveKey, memberChanges} =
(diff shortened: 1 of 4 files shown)


bwindels:
Closing this in favor of #472

PR #473, opened 2021-08-27
Snowpack + Typescript conversion (Part 2)

DanilaFe:


src/matrix/storage/idb/stores/RoomMemberStore.ts
```diff
@@ -15,44 +15,59 @@ See the License for the specific language governing permissions and
...
+// TODO: Move to RoomMember when that's TypeScript.
+export interface MemberData {
+    roomId: string;
+    userId: string;
+    avatarUrl: string;
+    displayName: string;
+    membership: "join" | "leave" | "invite" | "ban";
+}
+
+type MemberStorageEntry = MemberData & { key: string }
```
bwindels, line 39: :smile: nice, basically a shortcut for extending an interface with another.

src/matrix/storage/idb/stores/RoomMemberStore.ts
```diff
@@ -15,44 +15,59 @@ See the License for the specific language governing permissions and
...
+    get(roomId: string, userId: string): Promise<MemberStorageEntry | null> {
         return this._roomMembersStore.get(encodeKey(roomId, userId));
-	}
+    }
 
-	async set(member) {
-        member.key = encodeKey(member.roomId, member.userId);
-        return this._roomMembersStore.put(member);
-	}
+    async set(member: MemberData): Promise<IDBValidKey> {
```
bwindels: hmm, ideally we wouldn't leak IDB impl details like `IDBValidKey` from the storage layer. I don't think the returned key is used anywhere, so we can probably avoid the issue altogether by just returning undefined.

src/matrix/storage/idb/stores/SessionStore.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export class SessionStore {
+    private _sessionStore: Store<SessionEntry>
+
+    constructor(sessionStore: Store<SessionEntry>) {
+        this._sessionStore = sessionStore;
+    }
+
+    async get(key: IDBValidKey): Promise<any> {
```
bwindels:
```suggestion
    async get(key: string): Promise<any> {
```

src/matrix/storage/idb/stores/SessionStore.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    set(key: string, value: any): Promise<IDBValidKey> {
+        return this._sessionStore.put({key, value});
+    }
+
+    add(key: string, value: any): Promise<IDBValidKey> {
+        return this._sessionStore.add({key, value});
+    }
+
+    remove(key: IDBValidKey): Promise<undefined> {
```
bwindels:
```suggestion
    remove(key: string): Promise<undefined> {
```

src/matrix/storage/idb/stores/TimelineEventStore.ts
```diff
@@ -18,22 +18,49 @@ import {EventKey} from "../../../room/timeline/EventKey.js";
...
 
-function decodeEventIdKey(eventIdKey) {
+function decodeEventIdKey(eventIdKey: string): { roomId: string, eventId: string } {
     const [roomId, eventId] = eventIdKey.split("|");
     return {roomId, eventId};
 }
 
 class Range {
-    constructor(IDBKeyRange, only, lower, upper, lowerOpen, upperOpen) {
+    private _IDBKeyRange: any; // TODO what's the appropriate representation here?
```
bwindels: [This thing](https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L9136). Perhaps `typeof IDBKeyRange`?

src/matrix/storage/types.ts
```diff
@@ -14,20 +14,15 @@ See the License for the specific language governing permissions and
...
-    }
-
-    set(invite) {
-        return this._inviteStore.put(invite);
-    }
-
-    remove(roomId) {
-        this._inviteStore.delete(roomId);
-    }
+export interface RoomEvent {
```
bwindels: We tend to call these timeline events, as state events also appear in a room.

src/matrix/storage/idb/stores/TimelineEventStore.ts
```diff
@@ -231,38 +261,38 @@ export class TimelineEventStore {
...
+     *  @return a promise resolving to undefined if the operation was successful, or a StorageError if not.
      *  @throws {StorageError} ...
      */
-    insert(entry) {
+    insert(entry: StorageEntry): Promise<IDBValidKey> {
         entry.key = encodeKey(entry.roomId, entry.fragmentId, entry.eventIndex);
         entry.eventIdKey = encodeEventIdKey(entry.roomId, entry.event.event_id);
         // TODO: map error? or in idb/store?
-        this._timelineStore.add(entry);
+        return this._timelineStore.add(entry);
```
bwindels: So in addition to the earlier comment about not leaking `IDBValidKey` here, the lack of return is also sort off intentional. We don't have any store where keys are automatically created by indexeddb, so we never need the result of a `put` or `add`. Hence, we (I think) never await the promise, but rather make sure the transaction succeeds. The idea is that if there are a lot of `put`/`add`s, this creates a lot of `Promise` instances that are never awaited, so by not exposing it from the storage layer, I wanted to keep the option open of not even calling `reqAsPromise` on modification requests. I forgot the reason I haven't done so already.

bwindels: In short, perhaps don't add the returns for now and just make the return type `void`.

bwindels: I see now the jsdoc comment contradicts the above, but I just forgot to update it I assume.

bwindels (commented):
looks great, just a few small things

DanilaFe:
This should take care of everything you've pointed out here :smile:

bwindels (approved):

bwindels:
Checks didn't run for some reason on the last commit, but have run them locally and they all pass :heavy_check_mark:

PR #475, opened 2021-08-27
Snowpack + Typescript conversion (Part 4)

DanilaFe:


src/matrix/storage/idb/Transaction.ts
```diff
@@ -36,106 +36,111 @@ import {OperationStore} from "./stores/OperationStore";
 import {AccountDataStore} from "./stores/AccountDataStore";
 
 export class Transaction {
-    constructor(txn, allowedStoreNames, IDBKeyRange) {
+    private _txn: IDBTransaction;
+    private _allowedStoreNames: string[];
+    private _stores: { [storeName : string] : any };
+
+    constructor(txn: IDBTransaction, allowedStoreNames: string[], IDBKeyRange) {
```
bwindels: use the enum for the allowed store names? Also see earlier comment for type of IDBKeyRange

src/matrix/storage/idb/Storage.ts
```diff
@@ -44,13 +49,14 @@ export class Storage {
             if (this._hasWebkitEarlyCloseTxnBug) {
                 await reqAsPromise(txn.objectStore(storeNames[0]).get(WEBKITEARLYCLOSETXNBUG_BOGUS_KEY));
             }
+            // @ts-ignore
```
bwindels: why is this needed? to cast the enums?

src/matrix/storage/idb/schema.ts
```diff
@@ -86,13 +92,14 @@ function createE2EEStores(db) {
...
 
 // v5
-async function migrateEncryptionFlag(db, txn) {
+async function migrateEncryptionFlag(db: IDBDatabase, txn: IDBTransaction): Promise<void> {
     // migrate room summary isEncrypted -> encryption prop
     const roomSummary = txn.objectStore("roomSummary");
     const roomState = txn.objectStore("roomState");
-    const summaries = [];
-    await iterateCursor(roomSummary.openCursor(), summary => {
+    const summaries: any[] = [];
```
bwindels, line 103: reason to not use `SummaryData` here is because we're dealing with the `isEncrypted` that isn't supported anymore?

src/matrix/storage/idb/export.ts
```diff
@@ -14,25 +14,24 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { iterateCursor, txnAsPromise } from "./utils";
+import { iterateCursor, NOT_DONE, txnAsPromise } from "./utils";
 import { STORE_NAMES } from "../common";
 
-export async function exportSession(db) {
-    const NOT_DONE = {done: false};
+export async function exportSession(db: IDBDatabase): Promise<{ [storeName : string] : any }> {
```
bwindels: can be `any[]` I suppose?

bwindels (commented):
looking great, just a few comments!

src/matrix/storage/idb/schema.ts
DanilaFe: I mentioned in the comment that I would just leave the "old" entries unspecified (doesn't seem all that valuable). In this case it's possible to add the `& { isDisabled: boolean }` annotation, and work that out.

In addition, `SummaryData` is currently from a JS module, and this implicily typed as `any`.

bwindels (approved):

PR #481, opened 2021-08-31
Check for null keys in settings view model

jryans:
The keys to format in the settings view model might be null. This avoids crashing the UI by testing for this case.


bwindels (approved):
Hah, if you're using it without e2ee, of course :+1: I'd do the check in `fingerprintKey` though, I'll make the change, no worries. Thanks!

bwindels:
Ah, it's a fork, I can't make the change.

jryans:
@bwindels Thanks for the review, should be updated to match the approach you mentioned.

bwindels:
Thanks!

PR #487, opened 2021-09-06
improve menu on small screens

mcnesium:
original: 
 

* wrap to prevent the menu from popping out of small screens:

* menu distance to border equal on left and right
* reduce unused screen real estate inside the menu

* put reply and delete buttons next to each other to prevent delete button from popping out of the screen

### DIFF
--- a/src/platform/web/ui/css/themes/element/theme.css
+++ b/src/platform/web/ui/css/themes/element/theme.css
@@ -726,6 +726,9 @@ button.link {
     list-style: none;
     margin: 0;
     width: calc(100% - 48px);
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: space-between;
 }
 
 .menu li{
@@ -737,7 +740,7 @@ button.link {
     border: none;
     background-color: transparent;
     text-align: left;
-    padding: 8px 32px 8px 8px;
+    padding: 8px;
     font-size: 1.5rem;
     height: 24px;
     cursor: pointer;
@@ -749,8 +752,9 @@ button.link {
 
 .menu .quick-reactions {
     display: flex;
-    padding: 8px;
+    padding: 8px 8px 0;
     flex-wrap: wrap;
+    flex-basis: 100%;
 }
 
 .menu .quick-reactions button {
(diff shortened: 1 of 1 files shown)


src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -726,6 +726,9 @@ button.link {
     list-style: none;
     margin: 0;
     width: calc(100% - 48px);
+    display: flex;
```
bwindels: not sure why the delete button should be too wide? I'd rather not wrap the text buttons and keep one per line.

bwindels (changes requested):
not sure about the last commit, but other than that, looks like a good improvement! thanks!

src/platform/web/ui/css/themes/element/theme.css
mcnesium: keeping them one per line will make the delete button sit outside of the screen, when opening the menu on the last posted line:

bwindels: This should be fixed now when you rebase your branch on master

bwindels: (See #509)

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -725,6 +725,10 @@ button.link {
     background-color: white;
     list-style: none;
     margin: 0;
+    width: calc(100% - 48px);
```
bwindels, line 728: hmm, wouldn't this give very wide menus on bigger screens. If anything, this should be `max-width`? Do we need it at all though?

mcnesium: Ok that's true, did not think about wide screens.

Leaving out a width in general but having the `flex-wrap: wrap` from [here](https://github.com/vector-im/hydrogen-web/pull/487/commits/6fd2cc24f88ff3a34b189793962fd6d2c5f76b7f) would make the menu really narrow:

Screenshot from 2021-10-01 13-02-38

I guess what we want is a `min-width` instead. What do you think:

Screenshot from 2021-10-01 13-06-47

tandy-1000: I quite prefer the narrow menu.. looks easier to use one handed

bwindels:
Closing this as it's been a while and its not mergable as is because of the above concerns.

PR #488, opened 2021-09-06
Convert ListView and TimelineView to typescript

bwindels:


src/platform/web/ui/general/ListView.ts
```diff
@@ -0,0 +1,191 @@
...
+            this.loadList();
+        }
+    }
+
+    mount(): HTMLElement {
+        const attr: {[name: string]: any} = {};
+        if (this._className) {
+            attr.className = this._className;
+        }
+        this._root = el(this._tagName, attr);
```
DanilaFe, line 82: What if we do this:
```suggestion
        const root = this._root = el(this._tagName, attr);
```

src/platform/web/ui/general/ListView.ts
```diff
@@ -0,0 +1,191 @@
...
+        this.loadList();
+        if (this._onItemClick) {
+            this._root!.addEventListener("click", this);
```
DanilaFe, line 85: Then, this becomes:
```suggestion
            root.addEventListener("click", this);
```

src/platform/web/ui/general/ListView.ts
```diff
@@ -0,0 +1,191 @@
...
+        }
+        return this._root!;
```
DanilaFe, line 87: And this,
```suggestion
        return root;
```
In short, no `!` needed?

src/platform/web/ui/general/types.ts
```diff
@@ -13,15 +13,14 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-    let node;
-    try {
-        node = view.mount(mountArgs);
-    } catch (err) {
-        node = errorToDOM(err);
-    }
-    return node;
-}
\ No newline at end of file
+export interface UIView {
```
DanilaFe, line 21: Should we perhaps stick to the convention of starting our interfaces with `I`? it seems to be in use everywhere else.

bwindels: Have done so in https://github.com/vector-im/hydrogen-web/commit/a6bcfac597a22ae3a26f1d528def45a9a0d36497 in #486 as I have made changes there to the usage of the interface.

src/platform/web/ui/general/ListView.ts
bwindels: Have done so in https://github.com/vector-im/hydrogen-web/commit/b71a26b04a6368921279fd52d26e0a0b21d47e73 in #486

PR #491, opened 2021-09-08
Add a `/context` API call.

DanilaFe:
Adding mostly to keep track of what's left to be done:

1. [x] New fragments shouldn't be created unless a new overlap is detected
2. [x] Throw if return from the `/context` endpoint doesn't have end/begin tokens to avoid creating "live" fragments.
3. [ ] Switch other code to new `FragmentIdComparer.add` API (from `append`)


src/matrix/room/BaseRoom.js
```diff
@@ -266,6 +266,51 @@ export class BaseRoom extends EventEmitter {
...
+                fragmentIdComparer: this._fragmentIdComparer,
+                ownUserId: this._user.id,
+            });
+            const gapWriter = new GapWriter({
+                roomId: this._roomId,
+                storage: this._storage,
+                fragmentIdComparer: this._fragmentIdComparer,
+                relationWriter
+            });
+            const callbackResult = await callback(txn, gapWriter);
```
bwindels, line 290:
```suggestion
            {extraGapFillChanges, gapResult} = await callback(txn, gapWriter);
```

src/matrix/room/Room.js
```diff
@@ -110,6 +110,18 @@ export class Room extends BaseRoom {
         }
     }
 
+    async fetchContext(eventId, log = null) {
```
bwindels: This method should probably:
 - return the event entry for given event id? this way it can be used for reply preview rendering, I think?
 - first check if the event is not already in storage

bwindels: Also wondering how this intersects with `observeEvent` (which already does `_readEventById`). Perhaps this should be integrated with that method?

bwindels:
We'll also need a way to open a timeline at a given event id (for permalinks). That shouldn't happen in this PR, but perhaps take that into account when structuring the code here.

src/matrix/room/BaseRoom.js
DanilaFe: I thought so too, but no cigar:
2021-09-09-093453_1788x127_scrot

src/matrix/room/Room.js
DanilaFe: Since we have decided out of band to make `fetchContext` private and call it from `observeEvent`, we can probably forego checking if the entry is already in storage (`observeEvent` would already make that check)

DanilaFe:
This.... seems to work?

src/matrix/room/BaseRoom.js
bwindels: must be a TS thing then, it should be valid JS.

bwindels:
I'm going to close this, as we've discussed out of band that unfortunately, we can't store /context responses without potentially swallowing history given /sync returns things in stream order and /context in DAG order.

PR #494, opened 2021-09-14
Unit tests for GapWriter, using a new timeline mock utility

### LINKED ISSUE #393: History is swallowed before point of joining room
Caused by #160 mitigation. 

I guess Hydrogen could be a bit more forgiving in how it handles duplicate events in /messages

It assumes that the first duplicate event is on a different fragment and that your pagination is colliding into that fragment and hence they need to be linked up and the rest of the events will also overlap with the fragment. If that fragment is the same as the fragment who's gap your filling, it screams #160 and prevents it linking up with itself.

Perhaps I could check overlap for every event in the /messages response and filter out the ones on the same fragment.

I also wonder if federation makes this more likely to happen on the server side. IIRC federation can cause events through /sync and /messages to be in different order. I've filed this before as https://github.com/matrix-org/synapse/issues/7164 but it's expected (and somewhat underspecced) behaviour.

bwindels:


src/matrix/storage/idb/Store.ts
```diff
@@ -131,7 +131,7 @@ export class Store<T> extends QueryTarget<T> {
     private _transaction: Transaction;
 
     constructor(idbStore: IDBObjectStore, transaction: Transaction) {
-        super(new QueryTargetWrapper<T>(idbStore));
```
bwindels, line 135: can we bump the `_transaction` property to `QueryTarget` and keep the `IDBKeyRange` getter and add one for the factory? That way we don't create too many new fields...

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,268 @@ export class GapWriter {
...
+        assert.equal(fragment1.nextId, fragment2.id);
+        assert.equal(fragment2.previousId, fragment1.id);
+        assert.notEqual(fragment2.previousToken, null);
+        assert.notEqual(fragment1.nextToken, null);
+    }
+
+    return {
+        "Backfilling an empty fragment": async assert => {
+            const { txn, fragmentIdComparer, gapWriter, eventCreator } = await setup();
+            const emptyFragment = await createFragment(0, txn, fragmentIdComparer, { previousToken: startToken });
```
bwindels: can you extract the fragment id here in a variable and reuse it when you check for the results below in the store? Now the relation is implicit, by using EventKey.minKey which also happens to use fragmentId `0`, making the test work, but it's not apparent without knowing all of this.

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,268 @@ export class GapWriter {
...
+            const newEntry = FragmentBoundaryEntry.start(emptyFragment, fragmentIdComparer);
+
+            const response = eventCreator.createMessagesResponse();
+            await gapWriter.writeFragmentFill(newEntry, response, txn, null);
+
+            const allEvents = await txn.timelineEvents.eventsAfter(roomId, EventKey.minKey, 100 /* fetch all */);
+            for (let i = 0; i < response.chunk.length; i++) {
+                const responseEvent = response.chunk[response.chunk.length -i - 1];
+                const storedEvent = allEvents[i];
```
bwindels: Perhaps we check here that the fragmentId and eventIndex has been assigned as expected? This seems like a thing that could go wrong in case of an empty fragment.

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,268 @@ export class GapWriter {
...
+            const initialEntries = await prefillFragment(txn, eventCreator, existingFragment, 10);
+            const response = eventCreator.createMessagesResponse();
+            response.chunk.push(initialEntries[initialEntries.length-1]); /* Expect overlap */
+            await gapWriter.writeFragmentFill(newEntry, response, txn, null);
+
+            const allEvents = await txn.timelineEvents._timelineStore.selectAll();
+            let i = 0;
+            for (const initialEntry of initialEntries) {
+                const storedEvent = allEvents[i++];
+                assert.deepEqual(initialEntry, storedEvent.event);
```
bwindels: can we check the fragmentId and eventIndex here as well?

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,268 @@ export class GapWriter {
...
+            const { txn, fragmentIdComparer, gapWriter, eventCreator } = await setup();
+            const existingFragment = await createFragment(0, txn, fragmentIdComparer, { nextId: 1, nextToken: endToken });
+            const liveFragment = await createFragment(1, txn, fragmentIdComparer, { previousId: 0, previousToken: startToken });
+            const newEntry = FragmentBoundaryEntry.start(liveFragment, fragmentIdComparer);
+
+            const initialEntries = await prefillFragment(txn, eventCreator, existingFragment, 10);
+            const response = eventCreator.createMessagesResponse();
+            await gapWriter.writeFragmentFill(newEntry, response, txn, null);
+
+            const allEvents = await txn.timelineEvents._timelineStore.selectAll();
```
bwindels: can we just do a separate query here for both fragments so we don't need to poke in the internals of the store?

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,268 @@ export class GapWriter {
...
+import {createMockStorage} from "../../../../mocks/Storage.js";
+import {FragmentBoundaryEntry} from "../entries/FragmentBoundaryEntry.js";
+import {createEvent, withTextBody, withContent, withSender} from "../../../../mocks/event.js";
+import {NullLogger} from "../../../../logging/NullLogger.js";
+
+export function tests() {
+    const alice = "alice@hs.tdl";
+    const bob = "bob@hs.tdl";
+    const roomId = "!room:hs.tdl";
+    const startToken = "begin_token";
```
bwindels: I wonder if using static tokens here might hide bugs if we don't persist the right token, or don't overwrite it. How about using the index as the token?

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,268 @@ export class GapWriter {
...
+            for (const initialEntry of initialEntries) {
+                const storedEvent = allEvents[i++];
+                assert.deepEqual(initialEntry, storedEvent.event);
+            }
+            for (let j = 0; j < response.chunk.length - 1; j++) {
+                const responseEvent = response.chunk[response.chunk.length - j - 1];
+                const storedEvent = allEvents[i + j];
+                assert.deepEqual(responseEvent, storedEvent.event);
+            }
+            await assertWeakLink(assert, txn, 0, 1);
```
bwindels: if we don't use static values for the pagination tokens, we can check the actual value of the token rather than it being truthy.

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,171 @@ export class GapWriter {
...
+        const entries = await txn.timelineEvents.eventsAfter(roomId, new EventKey(fragmentId, KeyLimits.minStorageKey));
+        return entries.map(e => e.event);
+    }
+
+    async function fetchFragment(mocks, fragmentId) {
+        const {txn} = mocks;
+        return txn.timelineFragments.get(roomId, fragmentId);
+    }
+
+    function assertDeepLink(assert, fragment1, fragment2) {
```
bwindels: Deep and shallow aren't really used elsewhere to described whether a gap has been filled. How about `assertFilledLink` and `assertGapLink`?

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,171 @@ export class GapWriter {
...
+            assert.deepEqual(secondEvents.map(e => e.event_id), eventIds(10, 25));
+        },
+        "Backfilling a fragment that is expected to link up, but doesn't yet": async assert => {
+            const mocks = await setup();
+            const { timelineMock } = mocks;
+            timelineMock.append(10);
+            const {syncResponse, fragmentEntry: firstFragmentEntry} = await syncAndWrite(mocks);
+            timelineMock.append(20);
+            const {fragmentEntry: secondFragmentEntry} = await syncAndWrite(mocks, syncResponse);
+            await backfillAndWrite(mocks, secondFragmentEntry);
```
bwindels: perhaps make the amount of events an explicit parameter here so you don't have to assume it's the default 10 to make sense of the assertions below?

src/mocks/TimelineMock.ts
```diff
@@ -0,0 +1,257 @@
...
+        const index = eventOrdering.findIndex(e => e.event_id === token);
+        if (index === -1) {
+            // We didn't find this event token at all. What are we
+            // even looking at?
+            throw new Error("Invalid token passed to TimelineMock");
+        }
+        return index + step;
+    }
+
+    messages(begin: string, end: string | undefined, direction: "f" | "b", limit: number = 10) {
```
bwindels, line 86: is `end` ever used? It doesn't seem like something that comes from the protocol either?

src/mocks/TimelineMock.ts
```diff
@@ -0,0 +1,257 @@
...
+    context(eventId: string, limit: number = 10) {
+        if (limit <= 0) {
+            throw new Error("Cannot fetch zero or less events!");
+        }
+        let eventIndex = this._dagOrder.findIndex(e => e.event_id === eventId);
+        if (eventIndex === -1) {
+            throw new Error("Fetching context for unknown event");
+        }
+        const event = this._dagOrder[eventIndex];
+        limit -= 1;
```
bwindels: fwiw, synapse doesn't seem to include this in the limit count. If you request /context with the default 10, it will return 5 before and 5 after it seems, and the event itself of course. Indeed not obvious from the spec, but best to mirror synapse behaviour here?

src/mocks/TimelineMock.ts
```diff
@@ -0,0 +1,257 @@
...
+            assert.deepEqual(events2, expectedEvents2);
+        },
+        "A sync that receives no events has the same next batch as it started with": assert => {
+            const timeline = new TimelineMock(SENDER);
+            timeline.append(10);
+            const sync1 = timeline.sync();
+            const sync2 = timeline.sync(sync1.next_batch);
+            assert.equal(sync1.next_batch, sync2.next_batch);
+        }, 
+        "An event inserted in the midle still shows up in a sync": assert => {
```
bwindels:
```suggestion
        "An event inserted in the middle still shows up in a sync": assert => {
```

src/mocks/TimelineMock.ts
```diff
@@ -0,0 +1,257 @@
...
+            const timeline = new TimelineMock(SENDER);
+            timeline.append(30);
+            const sync1 = timeline.sync(undefined, 10);
+            const sync2 = timeline.sync(sync1.next_batch, 10)
+            assert.deepEqual(sync2.timeline.events, []);
+            assert.equal(sync2.timeline.limited, false);
+            timeline.insertAfter(TIMELINE_START_TOKEN, 1);
+            const sync3 = timeline.sync(sync2.next_batch, 10)
+            const events = sync3.timeline.events.map(e => e.event_id);
+            assert.deepEqual(events, eventIds(30, 31));
```
bwindels: I would have expected just event `31` in the last sync?

src/mocks/TimelineMock.ts
```diff
@@ -0,0 +1,257 @@
...
+        },
+        "An event inserted in the midle does not show up in a message fetch": assert => {
```
bwindels:
```suggestion
        "An event inserted in the start does not show up in a non-overlapping message fetch": assert => {
```

src/mocks/TimelineMock.ts
```diff
@@ -0,0 +1,257 @@
...
+        },
+        "Two consecutive message fetches detect newly inserted event": assert => {
+            const timeline = new TimelineMock(SENDER);
+            timeline.append(30);
+            const messages1 = timeline.messages(eventId(20), undefined, "b", 10);
+            const events1 = messages1.chunk.map(e => e.event_id).reverse();
+            timeline.insertAfter(eventId(9), 1);
+            const messages2 = timeline.messages(eventId(10), undefined, "b", 10);
+            const events2 = messages2.chunk.map(e => e.event_id).reverse();
+            assert.deepEqual(events1, eventIds(10, 20));
```
bwindels: move this below the definition of `events1` might make it easier to read.

src/mocks/TimelineMock.ts
```diff
@@ -0,0 +1,257 @@
...
+            assert.deepEqual(events2, eventIds(0, 10));
+        },
+        "Two consecutive message fetches detect newly inserted event": assert => {
+            const timeline = new TimelineMock(SENDER);
+            timeline.append(30);
+            const messages1 = timeline.messages(eventId(20), undefined, "b", 10);
+            const events1 = messages1.chunk.map(e => e.event_id).reverse();
+            timeline.insertAfter(eventId(9), 1);
```
bwindels, line 224: not sure how the /messages before this step is relevant to the test?

src/mocks/TimelineMock.ts
bwindels: oh, the lower bound is not included, this might be clearer:
```suggestion
            assert.deepEqual(events, [eventId(31)]);
```

src/mocks/TimelineMock.ts
```diff
@@ -0,0 +1,257 @@
+import {createEvent, withTextBody, withSender} from "./event.js";
+import {TimelineEvent} from "../matrix/storage/types";
+
+export const TIMELINE_START_TOKEN = "timeline_start";
+
+export function eventId(i: number): string {
+    return `$event${i}`;
+}
+
+export function eventIds(from: number, to: number): string[] {
```
bwindels, line 11: the lower event id is not included here, right? This made it a bit confusing the read the unit tests ... any particular reason for this? I'd at least add a comment here with that fact + rationale ...

src/matrix/storage/idb/Transaction.ts
```diff
@@ -56,13 +56,20 @@ export class Transaction {
     private _writeErrors: WriteErrorInfo[];
 
     constructor(txn: IDBTransaction, allowedStoreNames: StoreNames[], storage: Storage) {
+        this._txn = txn;
```
bwindels: did you mean to duplicate this here?

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,171 @@ export class GapWriter {
...
+
+            const firstFragment = await fetchFragment(mocks, firstFragmentEntry.fragmentId);
+            const secondFragment = await fetchFragment(mocks, secondFragmentEntry.fragmentId);
+            assertShallowLink(assert, firstFragment, secondFragment)
```
bwindels: fwiw, the shallow link here is actually made by the second `syncAndWrite` and not `backfillAndWrite`

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,171 @@ export class GapWriter {
...
+
+            const firstFragment = await fetchFragment(mocks, firstFragmentEntry.fragmentId);
+            const secondFragment = await fetchFragment(mocks, secondFragmentEntry.fragmentId);
+            assertDeepLink(assert, firstFragment, secondFragment)
+            const firstEvents = await allFragmentEvents(mocks, firstFragmentEntry.fragmentId);
+            assert.deepEqual(firstEvents.map(e => e.event_id), eventIds(0, 10));
+            const secondEvents = await allFragmentEvents(mocks, secondFragmentEntry.fragmentId);
+            assert.deepEqual(secondEvents.map(e => e.event_id), eventIds(10, 25));
+        },
+        "Backfilling a fragment that is expected to link up, but doesn't yet": async assert => {
```
bwindels:
```suggestion
        "Backfilling a fragment that is expected close a gap, but doesn't yet": async assert => {
```

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -252,3 +254,171 @@ export class GapWriter {
...
+    }
+
+    function assertShallowLink(assert, fragment1, fragment2) {
+        assert.equal(fragment1.nextId, fragment2.id);
+        assert.equal(fragment2.previousId, fragment1.id);
+        assert.notEqual(fragment2.previousToken, null);
+    }
+
+    return {
+        "Backfilling after one sync": async assert => {
```
bwindels, line 346: in all the tests we should probably `await txn.complete()` to ensure that indexeddb doesn't throw because of violated constraints, ...

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -47,28 +47,30 @@ export class GapWriter {
                 }
                 nonOverlappingEvents.push(...remainingEvents.slice(0, duplicateEventIndex));
                 if (!expectedOverlappingEventId || duplicateEventId === expectedOverlappingEventId) {
-                    // TODO: check here that the neighbourEvent is at the correct edge of it's fragment
-                    // get neighbour fragment to link it up later on
-                    const neighbourEvent = await txn.timelineEvents.getByEventId(this._roomId, duplicateEventId);
-                    if (neighbourEvent.fragmentId === fragmentEntry.fragmentId) {
-                        log.log("hit #160, prevent fragment linking to itself", log.level.Warn);
-                    } else {
+                    // Only link fragment if this is the first overlapping fragment we discover.
```
bwindels: I think if we change `findFirstOccurringEventId` to return all overlapping event ids (which should only be marginally more expensive in terms of query time), we can remove the loop here. This should make the code easier to understand (because it took me a good 10 reads to understand how the fix works, e.g. we do multiple iterations now because we don't trim remainingEvents anymore) and also make it more performant in extreme cases as you only do one query.

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -47,28 +47,30 @@ export class GapWriter {
...
+                    // TODO is this sufficient? Might we get "out of order" fragments from events?
+                    if (!neighbourFragmentEntry) {
```
bwindels: I suspect it might not be sufficient ... in theory all the overlapping events can be distributed amongst many fragments.
We should probably take the most occuring fragment to link up with, and at the same time don't change any existing links? Perhaps also not connect to a fragment that we're already connected to (indrectly) in the other direction? This is complicated ... I need to think more about this.

bwindels:
Excellent work on the unit tests! Left some comments, and have some questions about the actual fix, we can discuss "in person" later on.

src/matrix/room/timeline/persistence/GapWriter.js
DanilaFe: Indeed; the idea is to ensure that gap closing does _not_ occur in this case, which is a test of GapWriter.

src/matrix/storage/idb/Transaction.ts
DanilaFe: Not quite sure what you mean by this comment, sorry :worr

src/mocks/TimelineMock.ts
DanilaFe: Lower is included, upper is excluded; same idea as strings and and their indexing.

src/mocks/TimelineMock.ts
DanilaFe: The potentially-broken part is the encountering of overlapping events which causes the new event to be swallowed. A single `/messages` call in isolation will always create a new fragment, and there's little need to be concerned about swallowing; but if linking / amending a fragment is exercised (as it should be in this test), two calls seem advisable.

src/mocks/TimelineMock.ts
DanilaFe: It comes from [the `to` argument to `/messages`](https://matrix.org/docs/spec/client_server/latest#get-matrix-client-r0-rooms-roomid-messages); seemed appropriate to include it here.

src/matrix/storage/idb/Transaction.ts
bwindels: these 3 lines that were added seem identical to the 3 lines below this fragment.

bwindels:
Good to go, thanks again for this!

PR #505, opened 2021-09-21
Backfill behavior changes

### LINKED ISSUE #393: History is swallowed before point of joining room

DanilaFe:
Fixes #393 

I still want to take a look at what to do (if anything) with the expected overlapping event ID.

### DIFF
--- a/src/matrix/room/timeline/persistence/GapWriter.js
+++ b/src/matrix/room/timeline/persistence/GapWriter.js
@@ -310,6 +310,10 @@ export function tests() {
         return txn.timelineFragments.get(roomId, fragmentId);
     }
 
+    async function updatedFragmentEntry(mock, fragmentEntry) {
+        return fragmentEntry.withUpdatedFragment(await fetchFragment(mock, fragmentEntry.fragmentId));
+    }
+
     function assertFilledLink(assert, fragment1, fragment2) {
         assert.equal(fragment1.nextId, fragment2.id);
         assert.equal(fragment2.previousId, fragment1.id);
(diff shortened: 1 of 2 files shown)


DanilaFe:
Some questions:
* Does encountering the "expected overlap ID" change our decision (i.e., close gap even if not all events overlap)
* Can we accidentally advance the token past the edge of the proper fragment, at which point the gap will never be filled?
* Is picking the first overlapping event for the fragment sufficient? Do we want all fragments to match? Is there another heuristic?

bwindels:
Working on an alternative solution in #508

bwindels:
> Some questions:

All very good questions

> 
>     * Does encountering the "expected overlap ID" change our decision (i.e., close gap even if not all events overlap)

Yeah, this seems reasonable. We might still miss events that were inserted in the neighbour fragment but that is not a problem we can solve while persisting history, and seems like a fine tradeoff. I've put this on the table for sync v3 fwiw,

> 
>     * Can we accidentally advance the token past the edge of the proper fragment, at which point the gap will never be filled?

This is a very valid point and made me think if #508 is not a better approach. It would be easy to go too far if the neighbour fragment is small (e.g. 10 events).
> 
>     * Is picking the first overlapping event for the fragment sufficient? Do we want all fragments to match? Is there another heuristic?

Indeed what I went with in the other PR in the end.

PR #506, opened 2021-09-21
Drop events that have been synced before

### LINKED ISSUE #504: Sync is bricked when server syncs same event multiple times
Even though this server behaviour is incorrect, we should just ignore the event in question

bwindels:
Fixes #504


bwindels:
will merge tomorrow after I have looked at it with fresh :eyes:

Issue #482, opened 2021-09-01
In the element theme, the font-size in :root should be scoped to hydrogen


svieujot:
To avoid bleeding outsize of the hydrogen container when hydrogen is embedded in a web page.

https://github.com/vector-im/hydrogen-web/blob/25d0cbe9a2120169db2e00942d216b31afbcf021/src/platform/web/ui/css/themes/element/theme.css#L21

bwindels:
We could move this line to index.html so it is not included for embedded cases, but I wonder if this will effectively require to move away from `rem` units in CSS, as they couple the root font-size to the scale factor of the hydrogen. They don't seem ideal for the embedded use case, as they prevent other sections on the page from using `rem` units as well.  Instead, we could use:
 - `em` units, but there scaling will depend on nesting, which is undesirable.
 - `px`, but then we can't offer a scale factor.
 - use css variables for either the scale factor or all font-sizes used, but would need to see if postcss can inline these (for IE).
 - same, but use preprocessor variables (in combination with snowpack)

PR #514, opened 2021-09-26
Fix issue #482 - make font-size compatible with embedded scenario.

svieujot:


src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -18,10 +18,6 @@ limitations under the License.
 @import url('inter.css');
 @import url('timeline.css');
 
-:root {
-    font-size: 10px;
```
bwindels, line 22: As noted on the ticket, I don't think we can't remove this until Hydrogen doesn't use `rem` anymore as a unit in CSS. The PR will currently look like this, with fonts too large:

image

PR #521, opened 2021-09-30
Start migrating Observable code to TypeScript.

DanilaFe:
The most controversial thing here is relaxing the `protected` constraint on `ListView` and friends, I think. It seems required, though, because the class needs to be compatible with the `IListObserver` interface.

Some confidence checks:
* [x] `yarn test` still works.
* [x] `yarn tsc` still works with no errors.
* [x] Hydrogen still runs.


src/observable/BaseObservable.ts
```diff
@@ -14,20 +14,18 @@ See the License for the specific language governing permissions and
...
+    onSubscribeFirst(): void {
 
     }
 
-    onUnsubscribeLast() {
+    onUnsubscribeLast(): void {
 
     }
 
-    subscribe(handler) {
```
bwindels, line 30: can we put the `SubscriptionHandle` type in here and return `undefined` so we can keep doing `this._subscription = this._subscription!()` in `ListView`? Or was there some other reason you needed to change that?

src/observable/BaseObservable.ts
```diff
@@ -48,28 +46,26 @@ export class BaseObservable {
...
 export function tests() {
-    class Collection extends BaseObservable {
-        constructor() {
-            super();
-            this.firstSubscribeCalls = 0;
-            this.firstUnsubscribeCalls = 0;
-        }
-        onSubscribeFirst() {  this.firstSubscribeCalls += 1; }
+    class Collection extends BaseObservable<{}> {
+        firstSubscribeCalls: number= 0;
```
bwindels, line 65: missing space after number

src/platform/web/ui/session/room/TimelineView.ts
```diff
@@ -227,17 +232,17 @@ class TilesListView extends ListView<SimpleTile, TileView> {
         this.onChanged();
     }
 
-    protected onAdd(idx: number, value: SimpleTile) {
+    onAdd(idx: number, value: SimpleTile) {
```
bwindels, line 235: can we use `override` here and on the others?

src/observable/list/BaseMappedList.ts
```diff
@@ -18,60 +18,68 @@ limitations under the License.
...
-    constructor(sourceList, mapper, updater, removeCallback) {
+export type Mapper<F,T> = (value: F) => T
+export type Updater<F,T> = (mappedValue: T, params: any, value: F) => void;
+
+export class BaseMappedList<F,T> extends BaseObservableList<T> {
+    protected _sourceList: BaseObservableList<F>;
+    protected _sourceUnsubscribe: (() => void) | null = null;
+    _mapper: Mapper<F,T>;
+    _updater: Updater<F,T>;
+    _removeCallback: ((value: T) => void) | null;
```
bwindels, line 29: use `_removeCallback?` rather than `| null`?

bwindels (commented):
This looks great, thank you so much! Just a few nitpicks and this can be merged.

bwindels:
> The most controversial thing here is relaxing the protected constraint on ListView and friends

This is fine, if we want to hide this we can use some sort of inner object/class but that's not needed for now.

PR #546, opened 2021-10-01
Create MVVM SDK build powered by vite

bwindels:


bwindels:
Abandoning this PR in favor of #548 as what we're trying to do here (convert the build and develop server to vite AND split up the code into separate modules) is too much to chew on.

Issue #529, opened 2021-09-30
Create MVVM SDK


bwindels:
- switch to vite
 - generate library bundle exporting all views + view models + sessioncontainer/client bootstrap
 - generate api docs
 - fix #482
 - fix #467

Issue #292, opened 2021-03-24
Allow SDK usage from service worker


bwindels:
The service worker is growing in size, so everything the browser polls for an update, it downloads the whole file. Also, the code is growing in complexity, and it is becoming hard to navigate the single large file. All the replaced variables should be defined in the file, along with an import, and the bundle with the real code should go in a file that has a content hash on it.

~~We want to keep the ability to run the service worker without transpilation in both latest FF and Chromium, so we need to use `importScripts`~~ (can't do this anyway if we want to run sync from service worker).

We will end up using multiple files in the main import, and we would want those bundled up, so we would need to build a bundle for the service worker code that is worker-compatible (e.g. exports it symbols through a global variable) and then with the build script replace the name we  do `importScripts` of in the service worker.

 - service-worker/
   - service-worker.js
   - main.js
   - push.js
   - cache.js

being bundled up into `sw.js` and `sw-bundle-34985934.js`. Most of the work will be in the build.mjs script, the rest in reorganizing the service worker code into multiple files.

The olm worker currently isn't very big but might eventually also benefit from this.

bwindels:
I guess instead of exports we'll need to assign to the global scope, .e.g. `self`?

bwindels:
How could this tie in with running sync from the service worker (separate issue, not part of this)? The only way I can think off would be to:
 - not use ES modules anymore in production
 - have a bundle for the sdk and a bundle for the rest, that set a global variable
 - the page loads them with script tags
 - the service worker `importScripts` the sdk, and then finds it at the same global variable.
 - This means again the service worker can't be run locally for development ... so perhaps we should just bite the bullet and say you can only run the service worker locally in a browser that [supports imports in a worker](https://caniuse.com/mdn-javascript_statements_import_worker_support) (e.g. chrome at the time of writing). We'd have to transpile the service worker code to use `importScripts` to load its bundle though.

bwindels:
So, we'll need a rollup plugin where we can define files where imports should be replaced with importScripts (in our case `sw.js`). We'll also set it up that we have a `main.js` for the sw, and we turn that into a bundle. 
so we'll have:
```js
//... service worker constants...
import {startServiceWorker} from "./main.js";
startServiceWorker();
```
which will be turned into
```js
//... service worker constants...

// defines __sw_main as global variable
importScripts("sw-main-13924839.js");
const {startServiceWorker} = __sw_main;
startServiceWorker();
```

[...]

PR #548, opened 2021-10-05
Convert develop server and build system to using Vite

### LINKED ISSUE #230: Empty page with "SyntaxError: expected expression, got '.'" on Firefox 60esr and Chromium 73
**Describe the bug**
Hydrogen is broken on Firefox 60esr and all versions before 74 and Chromium 73 and all versions before 80, the page is only blank.

**To Reproduce**
1. Go to https://hydrogen.element.io/ with Firefox 60esr (any version before 74, see below) or Chromium 73 (any version before 80, see below)
2. You are only greeted with an empty blank page
3. In the web console, the following error is raised
	  ```
	  SyntaxError: expected expression, got '.' 
	  hydrogen-1718798642.js:131:13
	  ```
**Expected behavior**
Page is properly displayed without issue

**Desktop:**
 - OS: GNU/Linux Debian 9.13 x64
 - Browser Firefox
 - Version 60.9.0esr (64 bits)

**Desktop:**
 - OS: GNU/Linux Debian 9.13 x64
 - Browser Chromium
 - Version 73.0.3683.75 (64 bits)

**Additional context**
Identical issue as https://github.com/matrix-org/matrix.to/issues/193 (fixed in https://github.com/matrix-org/matrix.to/commit/a4642540b4de9446628c2f9b0fa39c916c932769 that I tried to apply without success) but for Hydrogen, the same applies here.

bwindels:
- [x] change file location to better match vite expectations
 - [x] remove usage of post-install script
 - [x] make unit tests run again
 - [x] make type checks pass again
 - [x] produce service worker with correct cache assets on build
 - [x] generate PWA manifest on build
 - [x] add version to <del>index.html</del> the bundle
 - [x] build with same postcss config
 - <del> build bundle for IE11</del> punting this for now
 - <del> make themes work again</del> removing the very basic theme support so far to bring it back in #531 
 - [x] make build hash and app version available to bundle
 - <del>add license header again to bundle</del> given the dependencies have multiple licenses and are bundled into one `vendor` bundle, not sure how to indicate these licenses. For now, I'll remove [LibreJS](https://www.gnu.org/software/librejs/) support that was introduced in #301, and have reopened #209.
 - [x] remove old dependencies and redundant scripts
 - [x] inline config in index.html
 - [x] enable & transpile service worker
 - [x] enable olm worker again
 - <del>transpile web worker</del> punting this as the built-in vite plugin prevents us from using `importScripts` for the olm worker, which we use to load olm. We'll need to be careful to not cause #230 in the worker here, and not transpiling it will again mean no IE11 support for now. The long term plan is to use our own plugin, see #292 

Related to https://github.com/vector-im/hydrogen-web/issues/529
Fixes https://github.com/vector-im/hydrogen-web/issues/230


bwindels:
This seems good to go :crossed_fingers:

PR #554, opened 2021-10-20
Only keep a limited amount of olm InboundGroupSession objects in memory to prevent OOM

### LINKED ISSUE #551: Decrypting a sync response with a large number (64?) of inbound megolm sessions used can make olm OOM
This has the same root cause as #150 but is for decryption rather than encryption.

bwindels:
Fixes #551


bwindels:
This should be good to go. Have tested new keys arriving and fetching keys from key backup.

PR #556, opened 2021-10-24
update caniuse-lite lockfile entry

yu-re-ka:
Previously the entry didn't contain a sha1 hash sum, probably because it
was manually edited in 7e05e4e109eb141bd8a688612f661bdf58deb986


bwindels:
Thanks for this, have done the same in PR https://github.com/vector-im/hydrogen-web/pull/577

PR #559, opened 2021-10-26
Continue filling gaps that return only non-rendered events in the first backfill

### LINKED ISSUE #517: Automatic gap filling stops if backfill doesn't contain any rendered items
This can happen if there are only reactions in the response (or #515 will also cause it) It will just show "Not loading" where the gap is in the timeline. It should fetch more events, but probably cap it at a reasonable limit (e.g. 5 fetches of 10?)

MidhunSureshR:
Signed-off-by: RMidhunSuresh <hi@midhun.dev>

Fixes #517

### DIFF
--- a/src/domain/session/room/timeline/TimelineViewModel.js
+++ b/src/domain/session/room/timeline/TimelineViewModel.js
@@ -49,8 +49,7 @@ export class TimelineViewModel extends ViewModel {
         this._showJumpDown = false;
     }
 
-    async watchForGapFill(gapPromise, tileRange, gapTile) {
-        console.log("watchForGapFill called");
+    async watchForGapFill(gapPromise, gapTile) {
         let hasSeenUpdate = false;
         const func = (idx, tile) => {
             if (tile.shape !== "gap") {
@@ -63,13 +62,13 @@ export class TimelineViewModel extends ViewModel {
             onRemove: (idx, tile) => func(idx, tile)
         };
         this.tiles.subscribe(subscription);
-        await gapPromise;
+        const gapResult = await gapPromise;
+        if (!gapResult) {
+            return;
+        }
         this.tiles.unsubscribe(subscription);
-        console.log("hasSeenUpdate is ", hasSeenUpdate);
         if (!hasSeenUpdate) {
-            // this.watchForGapFill(gapTile.notifyVisible(), undefined, gapTile);
-            const { startTile, endTile } = tileRange;
-            this.setVisibleTileRange(startTile, endTile);
+            this.watchForGapFill(gapTile.notifyVisible(), gapTile);
         }
     }
 
@@ -99,8 +98,7 @@ export class TimelineViewModel extends ViewModel {
             for (const tile of this._tiles.sliceIterator(startIndex, endIndex + 1)) {
                 const ret = tile.notifyVisible();
                 if (ret && !tile.isAtTop) {
-                    console.log("ret is", ret);
-                    this.watchForGapFill(ret, { startTile, endTile }, tile);
+                    this.watchForGapFill(ret, tile);
                 }
             }
             loadTop = startIndex < 10;
--- a/src/domain/session/room/timeline/TimelineViewModel.js
+++ b/src/domain/session/room/timeline/TimelineViewModel.js
@@ -49,6 +49,30 @@ export class TimelineViewModel extends ViewModel {
         this._showJumpDown = false;
     }
 
+    async watchForGapFill(gapPromise, tileRange, gapTile) {
+        console.log("watchForGapFill called");
+        let hasSeenUpdate = false;
+        const func = (idx, tile) => {
+            if (tile.shape !== "gap") {
+                hasSeenUpdate = true;
+            }
+        }
+        const subscription = {
+            onAdd: (idx, tile) => func(idx, tile),
+            onUpdate: (idx, tile) => func(idx, tile),
+            onRemove: (idx, tile) => func(idx, tile)
+        };
+        this.tiles.subscribe(subscription);
+        await gapPromise;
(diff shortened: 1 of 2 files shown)


MidhunSureshR:
Here's [some code](https://gist.github.com/MidhunSureshR/3335bce7848086d7fd7ad08be8d3fc2a) to reproduce the issue.

src/domain/session/room/timeline/TimelineViewModel.js
```diff
@@ -49,6 +49,29 @@ export class TimelineViewModel extends ViewModel {
...
 
+    async watchForGapFill(gapPromise, gapTile) {
+        let hasSeenUpdate = false;
+        const checkForUpdate = (idx, tile) => {
+            if (tile.shape !== "gap") {
+                hasSeenUpdate = true;
+            }
+        }
+        const subscription = {
+            onAdd: (idx, tile) => checkForUpdate(idx, tile),
```
bwindels, line 72: you'll need to support to onMove and onReset handlers here as well to prevent crashes when those event happen to be emitted (they're not currently on the timeline, but this is the contract for observableList.subscribe()), See BaseObservableList as well.

bwindels: Note that probably neither of those events should call checkForUpdate ...

src/domain/session/room/timeline/TimelineViewModel.js
```diff
@@ -49,6 +49,29 @@ export class TimelineViewModel extends ViewModel {
         this._showJumpDown = false;
     }
 
+    async watchForGapFill(gapPromise, gapTile) {
+        let hasSeenUpdate = false;
+        const checkForUpdate = (idx, tile) => {
+            if (tile.shape !== "gap") {
```
bwindels, line 61: AIUI, we need this to filter out the gap entering loading state and emitting an update from being considered as content from the gap fill being rendered in the timeline. Could you add a comment saying as much (or any other reason why we need this if the above is incorrect or incomplete).

src/domain/session/room/timeline/TimelineViewModel.js
```diff
@@ -49,6 +49,29 @@ export class TimelineViewModel extends ViewModel {
...
+                hasSeenUpdate = true;
+            }
+        }
+        const subscription = {
+            onAdd: (idx, tile) => checkForUpdate(idx, tile),
+            onUpdate: (idx, tile) => checkForUpdate(idx, tile),
+            onRemove: (idx, tile) => checkForUpdate(idx, tile)
+        };
+        this.tiles.subscribe(subscription);
+        const gapResult = await gapPromise;
```
bwindels: can we do a try/finally here so we always unsubscribe even if the promise throws an error?

src/domain/session/room/timeline/TimelineViewModel.js
```diff
@@ -49,6 +49,29 @@ export class TimelineViewModel extends ViewModel {
...
+        this.tiles.unsubscribe(subscription);
+        if (!gapResult) {
```
bwindels, line 86: perhaps add a comment here what the return value is and why we bail out

src/domain/session/room/timeline/TimelineViewModel.js
```diff
@@ -49,6 +49,29 @@ export class TimelineViewModel extends ViewModel {
...
+            return;
+        }
+        if (!hasSeenUpdate) {
+            this.watchForGapFill(gapTile.notifyVisible(), gapTile);
```
bwindels: I wonder if we should cap the recursion here to prevent filling the gap infinitely if somebody put a ton of non-rendered events in a room. Perhaps limit it to 10 backfills?

bwindels (commented):
Thanks for working on this! Looks great, just a few more cleanups.

src/domain/session/room/timeline/TimelineViewModel.js
```diff
@@ -56,12 +56,13 @@ export class TimelineViewModel extends ViewModel {
         }
         let hasSeenUpdate = false;
         const checkForUpdate = (idx, tile) => {
-            if (tile.shape !== "gap") {
+            if (tile.shape !== "gap" && !tile.isOwn) {
```
bwindels: perhaps `isPending` (locally pending being sent) would be more appropriate, as events sent by the syncing user could very well be returned from a backfill.

bwindels (commented):
Thanks for the changes, one more thing and should be ready!

MidhunSureshR:
As you suggested, one possible solution is to piggyback on the existing `updatePreviousSibling()` and `updateNextSibling()` calls so that we store references to the adjacent tiles for every tile (forming a  doubly linked-list of tiles). 
But I noticed that the `GapTile` never seems to get either of these calls (possibly because the GapTile is removed before any new tiles are added).

But I managed to fix the race by ensuring that we fill the gap again if we see an update but the call to `watchForGapFill` from that update has already returned (because we we're already filling the gap).

src/domain/session/room/timeline/TimelineViewModel.js
```diff
@@ -47,6 +47,58 @@ export class TimelineViewModel extends ViewModel {
         this._requestedEndTile = null;
         this._requestScheduled = false;
         this._showJumpDown = false;
+        this._watchMap = new WeakMap();
```
MidhunSureshR, line 50: My understanding is that if we use an object as key here, it should be a weak reference and therefore shouldn't prevent the GC from releasing that object. But logging the contents of `this._watchMap` after the gap has filled ( and tile disposed ) still shows the object there? I did manually trigger garbage collection as well.

MidhunSureshR:
Closing because Bruno suggested a much simpler way to solve this :tada:

PR #566, opened 2021-10-29
Support for multiline messages

MidhunSureshR:
- [x] Convert to <textarea> and style
- [x] Resize composer with text
- [ ] Fix timeline resizing

### DIFF
--- a/src/platform/web/ui/css/themes/element/theme.css
+++ b/src/platform/web/ui/css/themes/element/theme.css
@@ -537,7 +537,6 @@ a {
 }
 
 .MessageComposer_input > textarea {
-    padding: 0 16px;
     border: none;
     border-radius: 24px;
     background: #F6F6F6;
@@ -548,6 +547,8 @@ a {
     flex: 1;
     padding: 14px;
     box-sizing: border-box;
+    overflow: hidden;
+    max-height: 113px;
 }
 
 .MessageComposer_input > button.send {
--- a/src/platform/web/ui/css/themes/element/theme.css
+++ b/src/platform/web/ui/css/themes/element/theme.css
@@ -540,7 +540,6 @@ a {
     border: none;
     border-radius: 24px;
     background: #F6F6F6;
-    height: 48px;
     font-size: 14px;
     font-family: "Inter", sans-serif;
     resize: none;
--- a/src/platform/web/ui/css/themes/element/theme.css
+++ b/src/platform/web/ui/css/themes/element/theme.css
@@ -536,7 +536,7 @@ a {
     margin-left: 12px;
 }
 
-.MessageComposer_input > input {
+.MessageComposer_input > textarea {
     padding: 0 16px;
     border: none;
     border-radius: 24px;
@@ -544,6 +544,10 @@ a {
     height: 48px;
     font-size: 14px;
     font-family: "Inter", sans-serif;
+    resize: none;
+    flex: 1;
+    padding: 14px;
+    box-sizing: border-box;
 }
 
 .MessageComposer_input > button.send {
--- a/src/platform/web/ui/session/room/MessageComposer.js
+++ b/src/platform/web/ui/session/room/MessageComposer.js
@@ -28,11 +28,11 @@ export class MessageComposer extends TemplateView {
     }
 
     render(t, vm) {
-        this._input = t.input({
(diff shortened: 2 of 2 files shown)


src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -536,14 +536,19 @@ a {
...
-    padding: 0 16px;
+.MessageComposer_input > textarea {
     border: none;
     border-radius: 24px;
     background: #F6F6F6;
     height: 48px;
     font-size: 14px;
     font-family: "Inter", sans-serif;
+    resize: none;
+    flex: 1;
```
bwindels, line 546: I think this should be `flex: 0 0 auto`? e.g. it should not grow above it's intrinsic height which we change by setting the height inline? Otherwise when the timeline doesn't have enough stuff to fill the remaining height, it could end up sharing the remaining height with when growing.

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -536,14 +536,19 @@ a {
...
+    padding: 14px;
```
bwindels, line 547: This doesn't confusing the scrollHeight? At least it did for me in my quick demo in that every time you typed a character it would grow the height by the amount of padding.

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -536,14 +536,19 @@ a {
...
+    box-sizing: border-box;
+    overflow: hidden;
+    max-height: 113px;
```
bwindels, line 550: I assume this corresponds emperically to and amount of rows? Could you add a comment how many?

src/platform/web/ui/session/room/MessageComposer.js
```diff
@@ -105,4 +110,13 @@ export class MessageComposer extends TemplateView {
             this._attachmentPopup.showRelativeTo(evt.target, 12);
         }
     }
+
+    _adjustHeight() {
+        window.requestAnimationFrame(() => {
```
bwindels: I think you'll only want to schedule a frame callback if you haven't done so already (and clear your flag of course when the callback does get called).

src/platform/web/ui/session/room/MessageComposer.js
```diff
@@ -105,4 +110,13 @@ export class MessageComposer extends TemplateView {
...
+            this._input.style.height = "auto";
```
MidhunSureshR: This results in a slight change in height that causes the timeline layout to reflow but it does not trigger the ResizeObserver watching the timeline and as a result there is no call to `restoreScrollPosition`.
end result: the timeline is scrolled up a little.
I wonder if this has something to do with the browser render pipeline :thinking:

bwindels:
I also see the composer shrinking by a few pixels when typing the first character in the composer.

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -536,14 +536,19 @@ a {
     margin-left: 12px;
 }
 
-.MessageComposer_input > input {
-    padding: 0 16px;
+.MessageComposer_input > textarea {
     border: none;
     border-radius: 24px;
     background: #F6F6F6;
     height: 48px;
```
bwindels: > I also see the composer shrinking by a few pixels when typing the first character in the composer.

When I remove this line, it doesn't shrink on the first character.

bwindels:
So, after having a look, my findings wrt to preserving the scroll in the timeline while growing the composer:
 - we need to set `auto` for height first so the textarea can shrink. It will cause the textarea to take the height for the `rows=1` property. When not setting height to auto first, the textarea will grow on enter, but not shrink when removing lines
 - quickly switching the height between 1 line (with height=auto) and the scrollHeight doesn't cause the resizeobserver to emit, but it does cause a scroll event in the timeline. I wonder if this scroll event is distuingishable from a real scroll event.
 - in any case, restoring the scroll position when shrinking the timeline viewport seems broken (I think it used to work?). You can easily test this while making the window height smaller. The timeline doesn't stay at the bottom for me.
 - the first scroll event that messes up the scroll position is emitted after typing the first character on a second line. It will set the height to auto, causing the timeline to shrink, to then set it to the scrollHeight. As said, this will cause a scroll event but not trigger the resizeobserver.

bwindels:
I think for a first fix, it might be fine that the composer does not shrink.

PR #571, opened 2021-11-03
Make progress bar work for image uploads in Chrome

### LINKED ISSUE #570: Progress bar for image uploads don't work on Chrome
But they do work on Firefox.

MidhunSureshR:
Fixes #570 

[MDN](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest#monitoring_progress) says:
>Note: You need to add the event listeners before calling open() on the request. Otherwise the progress events will not fire. 

and they [elaborate](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/upload):
>Note: The spec also seems to indicate that event listeners should be attached after open(). However, browsers are buggy on this matter, and often need the listeners to be registered before open() to work.

### DIFF
--- a/src/platform/web/dom/request/xhr.js
+++ b/src/platform/web/dom/request/xhr.js
@@ -37,6 +37,11 @@ class RequestResult {
 
 function createXhr(url, {method, headers, timeout, format, uploadProgress}) {
     const xhr = new XMLHttpRequest();
+
+    if (uploadProgress) {
+        xhr.upload.addEventListener("progress", evt => uploadProgress(evt.loaded));
+    }
+
     xhr.open(method, url);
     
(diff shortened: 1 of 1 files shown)


MidhunSureshR:
Related to this, we should probably not [set Content-Length header](https://github.com/vector-im/hydrogen-web/blob/master/src/matrix/net/HomeServerApi.js#L59) with it being a [forbidden header name](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name)? 
Not that it really matters but Chrome throws:
```
Refused to set unsafe header "Content-Length"
```

bwindels (approved):

PR #573, opened 2021-11-05
Continue filling gaps that return only non-rendered events in the first backfill

### LINKED ISSUE #517: Automatic gap filling stops if backfill doesn't contain any rendered items

MidhunSureshR:
Fixes #517 with a lot more simpler code!

### DIFF
--- a/src/domain/session/room/timeline/tiles/GapTile.js
+++ b/src/domain/session/room/timeline/tiles/GapTile.js
@@ -49,13 +49,15 @@ export class GapTile extends SimpleTile {
     }
 
     async notifyVisible() {
+        // we do (up to 10) backfills while no new tiles have been added to the timeline
+        // because notifyVisible won't be called again until something gets added to the timeline
         let depth = 0;
-        let result;
-        do {
-            result = await this.fill();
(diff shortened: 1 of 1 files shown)


src/domain/session/room/timeline/tiles/GapTile.js
```diff
@@ -42,11 +43,19 @@ export class GapTile extends SimpleTile {
...
 
-    notifyVisible() {
-        this.fill();
+    async notifyVisible() {
+        let depth = 0;
+        let result;
+        do {
+            result = await this.fill();
+            depth = depth + 1;
+        } while (depth < 10 && !this._siblingChanged && result && !this.isDisposed)
```
MidhunSureshR: We check `isDisposed` here so that we don't start an extra gap fill before the entry update makes its way to GapTile.

bwindels (approved):

PR #577, opened 2021-11-05
update caniuse

bwindels:
Run `npx browserslist@latest --update-db` locally to do what #556 does, ensuring we have the right hash


yu-re-ka:
This is not the same as what my PR did. The resolved field is still missing a sha1 hash.

PR #583, opened 2021-11-09
Convert src/logging to typescript

MidhunSureshR:


src/logging/BaseLogger.ts
```diff
@@ -15,23 +15,27 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {LogItem} from "./LogItem.js";
-import {LogLevel, LogFilter} from "./LogFilter.js";
+import {LogItem, LabelOrValues, FilterCreator, LogCallback} from "./LogItem";
```
bwindels: if you just need the type for a given symbol (e.g. you wont call the constructor for a class, ...), it's usually better to do a separate [type import](https://www.typescriptlang.org/docs/handbook/modules.html#importing-types) for all the used types from a file so the bundler has better information to do tree-shaking.

src/logging/BaseLogger.ts
```diff
@@ -127,20 +131,16 @@ export class BaseLogger {
...
-    }
+    abstract _persistItem(item: LogItem, filter?: LogFilter | null, forced?: boolean): void;
 
-    async export() {
-        throw new Error("not implemented");
-    }
+    abstract export(): void;
 
     // expose log level without needing 
     get level() {
```
bwindels: I think this should work here:
```suggestion
    get level(): typeof LogLevel {
```

src/logging/BaseLogger.ts
```diff
@@ -127,20 +131,16 @@ export class BaseLogger {
         this._openItems.clear();
     }
 
-    _persistItem() {
-        throw new Error("not implemented");
-    }
+    abstract _persistItem(item: LogItem, filter?: LogFilter | null, forced?: boolean): void;
```
bwindels: if not too much work, you could only accept `undefined` for the filter, and then you can just write `filter?: LogFilter, ...`. You'll need to adjust the call-sites to pass `undefined` instead of `null`. That's the general direction we've taken so far with conversion, as just accepting `undefined` is more straighforward as accepting both `undefined` and `null`.

src/logging/LogFilter.ts
```diff
@@ -14,24 +14,30 @@ See the License for the specific language governing permissions and
...
+    Error,
+    Fatal,
+    Off
 }
 
+export type LogLevelOrNull = LogLevel | null;
+
 export class LogFilter {
-    constructor(parentFilter) {
+    private _min: LogLevelOrNull = null;
```
bwindels: same here with converging on `undefined` for optional values.

src/logging/LogItem.ts
```diff
@@ -15,13 +15,46 @@ See the License for the specific language governing permissions and
...
+
+type LogItemValues = LogItemWithLabel | LogItemNetwork | LogItemRef;
+
+export type LabelOrValues = string | LogItemValues;
+export type FilterCreator = ((filter: LogFilter, item: LogItem) => LogFilter) | null;
+export type LogCallback = (item: LogItem) => unknown;
 
 export class LogItem {
-    constructor(labelOrValues, logLevel, filterCreator, logger) {
+    public start: number;
```
bwindels: if only written in the constructor, you can add `readonly`.

src/logging/LogItem.ts
```diff
@@ -105,18 +141,31 @@ export class LogItem {
...
+    serialize(filter: LogFilter, parentStartTime: number | null = null, forced: boolean) {
         if (this._filterCreator) {
             try {
                 filter = this._filterCreator(new LogFilter(filter), this);
             } catch (err) {
                 console.error("Error creating log filter", err);
             }
         }
-        let children;
+        interface Item {
```
bwindels: can you put this at the top with the other types?

src/logging/LogItem.ts
```diff
@@ -105,18 +141,31 @@ export class LogItem {
         }
     }
 
-    serialize(filter, parentStartTime = null, forced) {
+    serialize(filter: LogFilter, parentStartTime: number | null = null, forced: boolean) {
```
bwindels: can you add a return type here? In general, all methods that have a return statement should have an explicit return type.

src/logging/LogItem.ts
bwindels: Perhaps you can call this `SerializedItem`?

src/logging/LogItem.ts
```diff
@@ -15,13 +15,46 @@ See the License for the specific language governing permissions and
...
+export type LabelOrValues = string | LogItemValues;
+export type FilterCreator = ((filter: LogFilter, item: LogItem) => LogFilter) | null;
+export type LogCallback = (item: LogItem) => unknown;
 
 export class LogItem {
-    constructor(labelOrValues, logLevel, filterCreator, logger) {
+    public start: number;
+    public logLevel: LogLevel;
+    public error: Error | null;
+    public end: number | null;
```
bwindels: given it's written after the constructor is run, better keep it private with a getter as we can't make it readonly

src/logging/LogItem.ts
```diff
@@ -15,13 +15,46 @@ See the License for the specific language governing permissions and
...
+
+export type LabelOrValues = string | LogItemValues;
+export type FilterCreator = ((filter: LogFilter, item: LogItem) => LogFilter) | null;
+export type LogCallback = (item: LogItem) => unknown;
 
 export class LogItem {
-    constructor(labelOrValues, logLevel, filterCreator, logger) {
+    public start: number;
+    public logLevel: LogLevel;
+    public error: Error | null;
```
bwindels: same as for `end`

src/logging/LogItem.ts
```diff
@@ -15,13 +15,46 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {LogLevel, LogFilter} from "./LogFilter.js";
+import {BaseLogger} from "./BaseLogger";
+import {LogLevel, LogLevelOrNull, LogFilter} from "./LogFilter";
+
+type LogItemWithLabel = {
```
bwindels: It's nice to have specific types, but in this case, they are not really used as we don't process the log values, it's also sort of freeform and might be a bit annoying to keep these definitions up to date. Up to you, but you could just have one generic type like this:
```ts
type LogItemValues {
  l?: string;
  t?: string;
  id?: string | number;
  status?: string | number;
  refId?: number;
  ref?: number;
  [key: string]: any
}
```

bwindels (commented):
great start! left a few comments

src/logging/BaseLogger.ts
MidhunSureshR: I can see that typescript already infers the type here. Should I still explicitly mention the return type?

src/logging/LogItem.ts
MidhunSureshR: Looks like end is written to in [BaseLogger](https://github.com/vector-im/hydrogen-web/blob/master/src/logging/BaseLogger.js#L29)

src/logging/LogFilter.ts
MidhunSureshR: We also use null for optional values elsewhere in BaseLogger. Can I refactor them to use undefined?

src/logging/BaseLogger.ts
```diff
@@ -15,23 +15,30 @@ See the License for the specific language governing permissions and
...
+    protected _platform: Platform;
 
-export class BaseLogger {
     constructor({platform}) {
-        this._openItems = new Set();
         this._platform = platform;
     }
 
-    log(labelOrValues, logLevel = LogLevel.Info) {
+    log(labelOrValues: LabelOrValues, logLevel: LogLevel = LogLevel.Info) {
```
altro3: Maybe, more correct, if you will add return type too (`: void`) like this: `log(labelOrValues: LabelOrValues, logLevel: LogLevel = LogLevel.Info): void` and in other places

src/logging/LogItem.ts
MidhunSureshR: `LogItem.error` is written to in [Sync.js](https://github.com/vector-im/hydrogen-web/blob/master/src/matrix/Sync.js#L133)

src/logging/BaseLogger.ts
MidhunSureshR: I see that I should based on other typescript files in the project 👍

src/logging/BaseLogger.ts
```diff
@@ -15,9 +15,12 @@ See the License for the specific language governing permissions and
...
 */
 
-import {LogItem, LabelOrValues, FilterCreator, LogCallback} from "./LogItem";
-import {LogLevel, LogFilter, LogLevelOrNull} from "./LogFilter";
-import {Platform} from "../platform/web/Platform.js";
+import {LogItem} from "./LogItem";
+import {LogLevel, LogFilter} from "./LogFilter";
+import type {FilterCreator, LabelOrValues, LogCallback} from "./LogItem";
+import type {LogLevelOrNull} from "./LogFilter";
+// todo: should this import be here just for getting the type? should it instead be done when Platform.js --> Platform.ts?
```
bwindels: It's indeed best to do it now already, even if the imported file is still javascript. That way, the conversion of this file is complete already, even though the types used here aren't fully typescripted. If there would be an misuse of the Platform class here, it will shop up once Platform is typescriptified.

src/logging/LogItem.ts
```diff
@@ -180,7 +181,7 @@ export class LogItem {
             return null;
         }
         // in (v)alues, (l)abel and (t)ype are also reserved.
-        const item: Item = {
+        const item: ISerializedItem = {
```
bwindels, line 145: do we need this explicit cast?

src/logging/LogFilter.ts
```diff
@@ -44,7 +44,7 @@ export class LogFilter {
             }
         }
         // neither our children or us have a loglevel high enough, filter out.
-        if (this._min !== null && !Array.isArray(children) && item.logLevel < this._min) {
+        if (this._min && !Array.isArray(children) && item.logLevel < this._min) {
```
bwindels: Perhaps explicitly check `this._min` for undefined as LogLevel is a number/enum (even though the value and `0` is not used IIRC).

src/logging/BaseLogger.ts
MidhunSureshR: Thanks 👍, I figured as much and have done the same elsewhere.

src/logging/LogItem.ts
```diff
@@ -100,27 +100,30 @@ export class LogItem {
         return item.run(callback);
     }
 
-    get duration() {
+    get duration(): number | undefined{
```
bwindels: space missing after undefined :)

src/logging/LogItem.ts
```diff
@@ -91,12 +154,12 @@ export class LogItem {
...
-    log(labelOrValues, logLevel = null) {
-        const item = this.child(labelOrValues, logLevel, null);
-        item._end = item._start;
+    log(labelOrValues: LabelOrValues, logLevel?: LogLevel): void {
+        const item = this.child(labelOrValues, logLevel);
+        item.end = item.start;
     }
 
-    set(key, value) {
+    set(key: string | object, value: unknown): void {
```
bwindels: I think value is optional?
```suggestion
    set(key: string | object, value?: unknown): void {
```

src/logging/BaseLogger.ts
```diff
@@ -15,23 +15,29 @@ See the License for the specific language governing permissions and
...
 */
 
-import {LogItem} from "./LogItem.js";
-import {LogLevel, LogFilter} from "./LogFilter.js";
+import {LogItem} from "./LogItem";
+import {LogLevel, LogFilter} from "./LogFilter";
+import type {FilterCreator, LabelOrValues, LogCallback, ILogItem} from "./LogItem";
+import type {Platform} from "../platform/web/Platform.js";
+
+type RunResult<T> = T | void | Promise<T | void>;
```
bwindels: I think we can just use T for the return type of `run` and friends? See https://www.typescriptlang.org/play?ts=4.3.5#code/PTAEEsGdQQ1BbA9gEwK4BsCmAoTAPAB0QCcAXUAbwF8BubbAY3RkmgBlEBzASVM3krZQw0AVQAjdOAahm4zOgD8ALlCRSxcADtOdEaGKotAHgAqAPgAUchavWadAGlAMY6dOJgMA1qsvg+eFUOHkCASlAAXnNQUzDVU0F9fVIACygAOht0KNkYeXQ9ZOFiTFJUYi0XNw8vb0s0qDCi4SpsNsZELXVYXK1MAHdQEN5+S2bO7vJxXJgBmADYDMMtSwAiVIV0RDXnFgBPLRkA-iiYiiEROYXyfqGABWJEeChMS2IztTLTcHhMRFQpHezgAjAAGCFhCb6UrlSqgNYDEjoZBrOhUCZuTBkSziDKQCT2SwAJmcAGYoXRJj0ZJElit1pt3DtnCcBNEkiJ2s1QCAXBBoAA3RDgZBAA

src/logging/BaseLogger.ts
```diff
@@ -43,28 +49,28 @@ export class BaseLogger {
...
-        if (logLevel === null) {
+    @return {ILogItem} the log item added, useful to pass to LogItem.refDetached */
+    runDetached(labelOrValues: LabelOrValues, callback: LogCallback<unknown>, logLevel?: LogLevel, filterCreator?: FilterCreator): ILogItem {
+        if (!logLevel) {
             logLevel = LogLevel.Info;
         }
-        const item = new LogItem(labelOrValues, logLevel, null, this);
-        this._run(item, callback, logLevel, filterCreator, false /* don't throw, nobody is awaiting */);
+        const item = new LogItem(labelOrValues, logLevel, this);
+        this._run(item, callback, logLevel!, false /* don't throw, nobody is awaiting */, filterCreator);
```
bwindels: try to avoid `!` as it can hide bugs. Here you could rename the argument to _logLevel and replace line 54-56 with:
```js
const logLevel = _logLevel === undefined ? LogLevel.Info : _logLevel;
```

src/logging/BaseLogger.ts
bwindels: For BaseLogger,_run needing a different return type based on shouldThrow flag, see https://www.typescriptlang.org/play?#code/MYGwhgzhAEAyD2BzAkgFwKYFtoG8BQ0h0ATgK4B2APACoB8AFMGCCAEZjADWAXNPWLwQoMmAJTQAvLWjUANNAgALeKRAATaouLwA7r1Rl0o3tQDcBImSp1GzNhx58BcJGizipM+UpXrN2vWgAM2YII14AN3gASzVzIhIKGgYmFnYuXn5BVxEPaTkFZVUNLV1eVnh4EHQwcmMZaAAfaCjY3AsEwgMAT3bO-uhgeHIIVBJ0CFUxiUG7dM56VEVoiFF4gYTooL4fYv9dcXwNjeJ0VFJicnHJkFR14+gAXw7Ox9nUYEU+dGJiQ5eNlsdkU-KUdP8Hv0lgFoD9iPdjs8NkiiM9nng8EMRmNopJoOR0DoXMIsPQ1pjhqNoGA8dEAHRWejRESSaQAVnkBlIRlM0AA9HzqdAVtBUN0AA7ofGkTCsH4U7HQVi0hkUJkszwAIlGxE18hCIDCa35guVIrFkpaMTUeCAA

src/logging/LogFilter.ts
```diff
@@ -14,39 +14,43 @@ See the License for the specific language governing permissions and
...
+    private _min?: LogLevel;
+    private _parentFilter?: LogFilter;
+
+    constructor(parentFilter?: LogFilter) {
         this._parentFilter = parentFilter;
-        this._min = null;
     }
 
-    filter(item, children) {
+    filter(item: ILogItem, children: Array<unknown> | null): boolean {
```
bwindels: Can children be an array of `ISerializedItem`?

bwindels (commented):
looks good, some comments

src/logging/LogItem.ts
MidhunSureshR: I think so because we add `.e`, `.f` and `.c`properties later below. Without the cast ts would complain that e, f and c do not exist on item.

src/logging/BaseLogger.ts
MidhunSureshR: Looks like typescript is able to infer that logLevel cannot be undefined after line 56 without the changes you described.
I think this was only a problem when I had `LogLevelOrNull`.

bwindels (approved):

PR #584, opened 2021-11-15
Convert /utils to typescript

MidhunSureshR:


src/utils/Disposables.ts
```diff
@@ -14,24 +14,25 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-function disposeValue(value) {
+type Func = () => void;
+type Disposable = { dispose: Func; [key: string]: any } | Func;
```
bwindels: why do you need the `[key: string]: any`? We might want to use an IDisposable interface in the future, so could this:
```suggestion
export interface IDisposable {
    dispose(): void;
}

type Disposable = IDisposable | (() => void);
```

src/utils/LockMap.ts
```diff
@@ -14,14 +14,12 @@ See the License for the specific language governing permissions and
...
 */
 
-import {Lock} from "./Lock.js";
+import {Lock} from "./Lock";
 
 export class LockMap {
-    constructor() {
-        this._map = new Map();
-    }
+    private readonly _map: Map<unknown, Lock> = new Map();
```
bwindels: might be best to use a generic parameter for the key type rather than unknown? this way you can't pass different types for a key to `takeLock`...

src/utils/formatSize.ts
```diff
@@ -14,7 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-export function formatSize(size, decimals = 2) {
+
+export function formatSize(size: number, decimals: number = 2): string | undefined {
```
bwindels: we could avoid `| undefined` here if we change `case 3` to `default` I think

src/utils/mergeMap.ts
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-export function mergeMap(src, dst) {
+export function mergeMap<K, V>(src: Map<K, V> | undefined, dst: Map<K, V>): void {
```
bwindels, line 17: why `| undefined`? seems like it would break ...

src/utils/timeout.ts
```diff
@@ -16,9 +16,13 @@ limitations under the License.
 */
 
 import {ConnectionError} from "../matrix/error.js";
+import type {Timeout} from "../platform/web/dom/Clock.js"
 
+type TimeoutCreator = (ms: number) => Timeout;
+// ts-todo: export type RequestResult from fetch.js? we'll need to wait until it's typescript though.
+type Abortable = { abort(): void; [key: string]: any };
```
bwindels: this type already exists in `AbortableOperation`, you could export it there and use it here.

bwindels (commented):
looks great, just a few things!

src/utils/mergeMap.ts
bwindels: second tests say so

bwindels (approved):
thanks for the changes, lgtm!

PR #594, opened 2021-11-24
sync-v3: Support placeholder room tiles

kegsay:
In preparation for supporting [sync-v3](https://github.com/matrix-org/sync-v3), add support for placeholder room tiles and lazy-loading them into room tiles. End users should not notice any changes as nothing in Hydrogen actually makes placeholder tiles yet (only the prototype HTML page).


kegsay:
Closing this as we've decided to do sync v3 work on a long-lived feature branch instead. See https://github.com/vector-im/hydrogen-web/tree/kegan/syncv3

PR #595, opened 2021-11-25
Convert /matrix/push to typescript

MidhunSureshR:


src/matrix/push/Pusher.ts
```diff
@@ -0,0 +1,90 @@
...
+    pushkey: string;
+    data: IPusherData;
+    append?: boolean;
+    profile_tag?: string;
+}
+
+interface IPusherData {
+    format?: string;
+    url?: string;
+    // todo: where did this come from?
```
bwindels: This is the push data that the homeserver will pass to the notification server, sygnal, and is defined here for web push: https://github.com/matrix-org/sygnal/blob/main/docs/applications.md#push-key-and-expected-push-data

bwindels: Also see where the enablePush method in NotificationService.

bwindels (commented):
lgtm, have clarified where the push data comes from. I'll leave it up to you if you want to remove the comment and/or make any other adjustments accordingly.

src/matrix/push/Pusher.ts
MidhunSureshR: Thank you! Have made slight changes.

bwindels (approved):

PR #597, opened 2021-11-25
Convert matrix/ssss to typescript

MidhunSureshR:


src/matrix/ssss/SecretStorage.ts
```diff
@@ -37,7 +43,7 @@ export class SecretStorage {
         }
     }
 
-    async _decryptAESSecret(type, encryptedData) {
+    async _decryptAESSecret(type: string, encryptedData: any): Promise<string> {
```
bwindels: perhaps we can type encrypted data (and cast `accountData?.content?.encrypted?.[this._key.id]` to it above), it's defined here https://github.com/uhoreg/matrix-doc/blob/symmetric_ssss/proposals/2472-symmetric-ssss.md and should look like this:
```json
{
  "type": "m.megolm_backup.v1",
  "content": {
    "encrypted": {
      "the key id": {
        "iv": "...",
        "ciphertext": "...",
        "mac": "..."
      }
    }
  }
}
```

src/matrix/ssss/common.ts
```diff
@@ -14,25 +14,42 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import type {Platform} from "../../platform/web/Platform.js";
+
+export interface IKeyDescription {
```
bwindels: I'd use `type` here, see https://github.com/vector-im/hydrogen-web/blob/master/doc/TS-MIGRATION.md#use-type-rather-than-interface-for-named-parameters-and-pojo-return-values
```suggestion
export type KeyDescriptionData = {
```

src/matrix/ssss/recoveryKey.ts
```diff
@@ -13,17 +13,24 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import {Key} from "./common.js";
+import {Key} from "./common";
+import {KeyDescription} from "./common";
+import type {Platform} from "../../platform/web/Platform.js";
 
-const OLM_RECOVERY_KEY_PREFIX = [0x8B, 0x01];
+export type Olm = {
```
bwindels: Hmm, [libolm has types](https://gitlab.matrix.org/matrix-org/olm/-/blob/master/javascript/index.d.ts), can we do `import type * as Olm from "@matrix-org/olm"`?

bwindels (approved):
Thanks! Mostly looks great, just a few small things.

bwindels (commented):
meant to comment rather than approve before

src/matrix/ssss/SecretStorage.ts
MidhunSureshR: But EncryptedData only needs to be:
```json
 {
        "iv": "...",
        "ciphertext": "...",
        "mac": "..."
}
```
?

src/matrix/ssss/recoveryKey.ts
MidhunSureshR: Can't get this to work: `Cannot use namespace Olm as type`
 But I'm convinced that I'm missing something because the type declaration for `aes-js` also did something similar :thinking:

MidhunSureshR: Maybe I need to do `typeof Olm` as opposed to `Olm`
Edit: I think this is it!

src/matrix/ssss/recoveryKey.ts
```diff
@@ -30,7 +26,7 @@ const OLM_RECOVERY_KEY_PREFIX = [0x8B, 0x01] as const;
  * @param  {string} recoveryKey
  * @return {Key}
  */
-export function keyFromRecoveryKey(keyDescription: KeyDescription, recoveryKey: string, olm: Olm, platform: Platform): Key {
+export function keyFromRecoveryKey(keyDescription: KeyDescription, recoveryKey: string, olm: typeof Olm, platform: Platform): Key {
```
bwindels: why not `Olm`?

bwindels (commented):
Changes look good, just one question, and some conflicts to resolve.

src/matrix/ssss/recoveryKey.ts
bwindels: Ah just saw your comment above about making it work.

src/matrix/ssss/recoveryKey.ts
```diff
@@ -13,17 +13,20 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import {Key} from "./common.js";
+import {Key} from "./common";
+import {KeyDescription} from "./common";
+import type {Platform} from "../../platform/web/Platform.js";
+import type * as Olm from "@matrix-org/olm"
```
bwindels: How about this to avoid the typeof everywhere?
```suggestion
import type * as OlmNamespace from "@matrix-org/olm";
type Olm = typeof OlmNamespace;
```

src/matrix/ssss/index.ts
```diff
@@ -22,7 +22,9 @@ import type {Storage} from "../storage/idb/Storage";
 import type {Transaction} from "../storage/idb/Transaction";
 import type {KeyDescriptionData} from "./common";
 import type {Platform} from "../../platform/web/Platform.js";
-import type * as Olm from "@matrix-org/olm"
+import type * as OlmNamespace from "@matrix-org/olm"
+
+type Olm = typeof OlmNamespace;
```
bwindels, line 27: In the end I'm not sure this actually does anything because when I hover over `Olm` in my editor, it says it's of type `any` :) Oh well, good enough for now though.

bwindels (approved):
Thanks for the changes!

PR #601, opened 2021-11-29
Ensure unwanted data do not end up in logs

MidhunSureshR:


bwindels (commented):
This looks like it would do the trick very well, but it's mixing up concerns a bit, to a level where it would be hard to extract the logging layer from the app. Perhaps we could inject a transform function into the logger (so accessible here through `this._logger.serializeTransformer` or something) where we pass in the serialized object. When we construct the logging in Platform.js, we could then pass in a function and filter out the loginToken there? Wdyt?

bwindels (approved):
perfect, thank you!

PR #608, opened 2021-12-01
Clarify approach to type data objects in doc

MidhunSureshR:


doc/TS-MIGRATION.md
```diff
@@ -2,8 +2,12 @@
...
-`type` and `interface` can be used somewhat interchangebly used, but let's use `type` to describe data and `interface` to describe (polymorphic) behaviour.
+`type` and `interface` can be used somewhat interchangeably, but let's use `type` to describe data and `interface` to describe (polymorphic) behaviour.
 
 Good examples of data are option objects to have named parameters, and POJO (plain old javascript objects) without any methods, just fields.
 
 Also see [this playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBACghgJwgO2AeTMAlge2QZygF4oBvAKCiqmTgFsIAuKfYBLZAcwG5LqATCABs4IAPzNkAVzoAjCAl4BfcuVCQoAYQAWWIfwzY8hEvCSpDuAlABkZPlQDGOITgTNW7LstWOR+QjMUYHtqKGcCNilHYDcAChxMK3xmIIsk4wBKewcoFRVyPzgArV19KAgAD2AUfkDEYNDqCM9o2IQEjIJmHT0DLvxsijCw-ClIDsSjAkzeEebjEIYAuE5oEgADABJSKeSAOloGJSgsQh29433nVwQlDbnqfKA)
+
+## Use `Record<string, any>` to describe a type that accepts any Javascript object.
+
+Prefer this over index signature, `any` or `object`.
```
bwindels: Consider adding the rationale here:
```suggestion
Record<string, any> allows us to avoid passing in primitive types and prevents type errors when accessing properties. As this is a temporary type while converting javascript, it seems best to not add any additional requirements. If any errors occur, they must have already been present, and we should fix it by adding proper types. So prefer `Record<string, any>` over `[key: string]: any`, `any` or `object`.
```

doc/TS-MIGRATION.md
```diff
@@ -2,8 +2,12 @@
...
+`type` and `interface` can be used somewhat interchangeably, but let's use `type` to describe data and `interface` to describe (polymorphic) behaviour.
 
 Good examples of data are option objects to have named parameters, and POJO (plain old javascript objects) without any methods, just fields.
 
 Also see [this playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBACghgJwgO2AeTMAlge2QZygF4oBvAKCiqmTgFsIAuKfYBLZAcwG5LqATCABs4IAPzNkAVzoAjCAl4BfcuVCQoAYQAWWIfwzY8hEvCSpDuAlABkZPlQDGOITgTNW7LstWOR+QjMUYHtqKGcCNilHYDcAChxMK3xmIIsk4wBKewcoFRVyPzgArV19KAgAD2AUfkDEYNDqCM9o2IQEjIJmHT0DLvxsijCw-ClIDsSjAkzeEebjEIYAuE5oEgADABJSKeSAOloGJSgsQh29433nVwQlDbnqfKA)
+
+## Use `Record<string, any>` to describe a type that accepts any Javascript object.
+
+`Record<string, any>` allows us to avoid passing in primitive types and prevents type errors when accessing properties. As this is a temporary type while converting javascript, it seems best to not add any additional requirements. If any errors occur, they must have already been present, and we should fix it by adding proper types. So prefer `Record<string, any>` over `[key: string]: any`, `any` or `object`.
```
bwindels: Hmm, actually Travis gave another idea that I like as well:
for every data type needed, we create a `type` alias with just `{[key: string]: any}` as a preparation to later fill in the type. The advantage is you already have different types for objects with different layouts (although compatible) and adding the types later will require minimal effort. Wdyt?

MidhunSureshR: Sgtm. But there are still rare circumstances where a method/function should accept any object eg: `encodeBody` function in `src/matrix/net/common.ts`.

bwindels: You want to elaborate this in the doc here?

bwindels: @MidhunSureshR so you want to change the guidance here to introduce a new type alias to `{[key: string]: any}` as discussed above?

MidhunSureshR: Sorry, took me a while to get to this. Have made some changes, let me know what you think.

Issue #528, opened 2021-09-30
Typescript conversion


bwindels:
- [x] convert logging/
- [x] convert utils/
- [x] convert matrix/net
- [x] convert matrix/login
- [x] convert matrix/push
- [x] convert matrix/sessioninfo
- [x] convert matrix/ssss
- [x] convert matrix/e2ee/olm
- [ ] convert matrix/e2ee/megolm
- [ ] convert matrix/e2ee
- [ ] convert matrix/room/members
- [ ] convert matrix/room/sending
- [ ] convert matrix/room/timeline/entries/ (and needed parts in matrix/room/timeline)
- [ ] convert matrix/room/timeline/persistence/
- [ ] convert rest of matrix/room/timeline
- [ ] convert matrix/room
- [ ] convert matrix/ (split up Sync types in several files under matrix/sync?)
- [ ] convert domain/navigation
- [x] convert domain/ViewModel
- [x] convert domain/login
- [ ] convert domain/session/room/timeline/linkify/ & MessageBody & deserialize
- [ ] convert domain/session/room/timeline/tiles/
- [ ] convert TilesCollection
- [ ] convert rest of domain/session/room/timeline
- [ ] convert domain/session/room
- [ ] convert domain/session/settings
- [ ] convert domain/session/rightpanel
- [ ] convert domain/session/leftpanel (@ibeckermayer is interested in working on this after LoginViewModel)
- [ ] convert domain/session
- [ ] convert rest of domain
- [ ] convert platform/web/utils
- [ ] convert platform/web/dom
- [ ] convert platform/web/Platform
- [ ] convert platform/web/ui/general
- [ ] convert platform/web/ui/login
- [ ] convert platform/web/ui/session/room/timeline
- [ ] convert platform/web/ui/session/room/
- [ ] convert platform/web/ui/session/leftpanel
- [ ] convert platform/web/ui/session/rightpanel
- [ ] convert platform/web/ui/session/settings
- [ ] convert platform/web/ui/session/leftpanel
- [ ] convert platform/web/ui/session/rightpanel
- [ ] convert platform/web/ui/session
- [ ] convert platform/web/ui
- [ ] convert platform/web/worker?
- [ ] extract interface for `Platform` and add needed stuff in each step below
- [ ] service worker?

shubhangi013:
[...]

PR #610, opened 2021-12-04
migrated 2 files from js to ts

shubhangi013:
MIgrated `/src/platform/web/ui/general/StaticView.js` and `/src/platform/web/ui/general/LoadingView.js` from javascript to typescript.
Solves part of #528
proof that the changes run fine, when run locally :

Screenshot from 2021-12-04 23-47-38


src/platform/web/ui/RootView.js
```diff
@@ -19,7 +19,7 @@ import {LoginView} from "./login/LoginView.js";
 import {SessionLoadView} from "./login/SessionLoadView.js";
 import {SessionPickerView} from "./login/SessionPickerView.js";
 import {TemplateView} from "./general/TemplateView";
-import {StaticView} from "./general/StaticView.js";
+import {StaticView} from "./general/StaticView.ts";
```
bwindels: For typescript, we leave the extension off
```suggestion
import {StaticView} from "./general/StaticView";
```

src/platform/web/ui/general/LoadingView.ts
```diff
@@ -17,7 +17,8 @@ limitations under the License.
 import {TemplateView} from "./TemplateView";
 import {spinner} from "../common.js";
 
-export class LoadingView extends TemplateView {
+export class LoadingView <T> extends TemplateView <T> {
```
bwindels: `T` is the type of the view model here, so you can use `any` as the type.
```suggestion
export class LoadingView extends TemplateView<any> {
```

src/platform/web/ui/general/LoadingView.ts
```diff
@@ -17,7 +17,8 @@ limitations under the License.
...
+    className:string;
```
bwindels: Not sure what this is for?

src/platform/web/ui/general/StaticView.ts
```diff
@@ -15,15 +15,17 @@ See the License for the specific language governing permissions and
...
-    constructor(value, render = undefined) {
+    private _root?: Element;
+    render: any;
+    constructor(value:any , render : any = undefined) {
         if (typeof value === "function" && !render) {
             render = value;
             value = null;
         }
-        this._root = render ? render(tag, value) : this.render(tag, value);
+        this._root= render ? render(tag, value) : this.render(tag, value);
```
bwindels, line 28: Please avoid unrelated cosmetic changes

src/platform/web/ui/general/StaticView.ts
```diff
@@ -15,15 +15,17 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {tag} from "../general/html";
+import {tag} from "./html";
 
 export class StaticView {
-    constructor(value, render = undefined) {
+    private _root?: Element;
+    render: any;
```
bwindels:
```suggestion
    render: (typeof tag, T) => Element;
```

src/platform/web/ui/general/StaticView.ts
```diff
@@ -15,15 +15,17 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {tag} from "../general/html";
+import {tag} from "./html";
 
 export class StaticView {
```
bwindels:
```suggestion
export class StaticView<T> {
```

With T being the type of the view model used.

bwindels (commented):
Thanks for your PR, left some comments

shubhangi013:
> Thanks for your PR, left some comments

Thankyou so much! will do these changes.

src/platform/web/ui/general/StaticView.ts
shubhangi013: > 

This throws in the error `This expression is not callable. Type 'never' has no call signatures.ts(2349)` for line 28. Using typeof also throws in error.

bwindels: Ok, if you can have a look if we can do better than `any` for this callback, they would be great :+1:

shubhangi013: > 

I'll look into it. I had to use `any` in both line 22 and 23 to avoid any error. :sweat_smile: So I can't think of assigning any other type to render.

shubhangi013:
@bwindels I've done the suggested changes.

shubhangi013:
@bwindels  could you review once? :sweat_smile:

bwindels:
I'm sorry, but getting this into shape will take up too much of my time.

PR #611, opened 2021-12-06
Threading fallback - PR 1 - Link events with their related event

MidhunSureshR:
- [x] Create proper event-entries with event data fetched from hs
- [x] Decryption of event-entries created with data from hs
- [x] Display-name/avatar data with event-entries created with data from hs
- [ ] Log operations
## Overview:
- Implement context endpoint needed to fetch an event from hs
- Link every entry with its context-entry (taking care of things like updates)

## Terminologies / changes added to EventEntry:  

For a given `EventEntry` e:
- e.**contextEventId**: is the event-id of another event to which this event is a reply of.
- e.**contextEntry**: is the actual event-entry to which this event is a reply of.
- e.**contextForEntries**: = `{event| event.contextEntry = e}` i.e this is an array of event-entries such that for each event-entry in this array, *e* is its context.  
Used for propagating updates from a given entry to other entries that depend on it to render reply-previews; eg: if an entry is redacted, all other entries which use this entry as context for rendering reply preview must be notified (so that redaction can be rendered in reply previews).

## Terminologies / changes added to Timeline:  
- **contextEntriesNotInTimeline**: Is a Map of entries that were fetched from storage/hs because it was needed as a context for some other entry. They are fetched from storage/hs because they are currently not in the timeline( i.e not in remoteEntries ).

### DIFF
--- a/src/matrix/room/BaseRoom.js
+++ b/src/matrix/room/BaseRoom.js
@@ -501,7 +501,9 @@ export class BaseRoom extends EventEmitter {
                 },
                 clock: this._platform.clock,
                 logger: this._platform.logger,
-                powerLevelsObservable: await this.observePowerLevels()
+                powerLevelsObservable: await this.observePowerLevels(),
+                fetchEventFromStorage: eventId => this._readEventById(eventId),
+                fetchEventFromHomeserver: eventId => this._getEventFromHomeserver(eventId)
             });
             try {
                 if (this._roomEncryption) {
@@ -557,6 +559,12 @@ export class BaseRoom extends EventEmitter {
             }
             return entry;
         }
+    }
+
+    async _getEventFromHomeserver(eventId) {
+        const response = await this._hsApi.event(this._roomId, eventId).response();
+        const entry = new EventEntry({ event: response }, this._fragmentIdComparer);
+        return entry;
     }
 
 
--- a/src/matrix/room/BaseRoom.js
+++ b/src/matrix/room/BaseRoom.js
@@ -571,7 +571,12 @@ export class BaseRoom extends EventEmitter {
             displayName: member.content.displayname,
             avatarUrl: member.content.avatar_url
         };
-        return new NonPersistedEventEntry(entry, this._fragmentIdComparer);
+        const eventEntry = new NonPersistedEventEntry(entry, this._fragmentIdComparer);
+        if (eventEntry.eventType === EVENT_ENCRYPTED_TYPE) {
+            const request = this._decryptEntries(DecryptionSource.Timeline, [eventEntry]);
+            await request.complete();
+        }
+        return eventEntry;
     }
 
 
--- a/src/matrix/room/BaseRoom.js
+++ b/src/matrix/room/BaseRoom.js
@@ -562,9 +562,15 @@ export class BaseRoom extends EventEmitter {
     }
 
     async _getEventFromHomeserver(eventId) {
-        const response = await this._hsApi.event(this._roomId, eventId).response();
-        const entry = new EventEntry({ event: response }, this._fragmentIdComparer);
-        return entry;
+        const response = await this._hsApi.context(this._roomId, eventId, 0).response();
+        const sender = response.event.sender;
+        const member = response.state.find(e => e.type === "m.room.member" && e.user_id === sender);
+        const entry = {
+            event: response.event,
+            displayName: member.content.displayname,
+            avatarUrl: member.content.avatar_url
+        };
+        return new EventEntry(entry, this._fragmentIdComparer);
     }
 
 
--- a/src/matrix/room/BaseRoom.js
+++ b/src/matrix/room/BaseRoom.js
@@ -30,6 +30,7 @@ import {DecryptionSource} from "../e2ee/common.js";
 import {ensureLogItem} from "../../logging/utils";
 import {PowerLevels} from "./PowerLevels.js";
 import {RetainedObservableValue} from "../../observable/ObservableValue";
+import {NonPersistedEventEntry} from "./timeline/entries/NonPersistedEventEntry";
 
 const EVENT_ENCRYPTED_TYPE = "m.room.encrypted";
 
@@ -570,7 +571,7 @@ export class BaseRoom extends EventEmitter {
             displayName: member.content.displayname,
             avatarUrl: member.content.avatar_url
         };
-        return new EventEntry(entry, this._fragmentIdComparer);
+        return new NonPersistedEventEntry(entry, this._fragmentIdComparer);
     }
 
 
--- a/src/matrix/room/BaseRoom.js
+++ b/src/matrix/room/BaseRoom.js
@@ -30,7 +30,7 @@ import {DecryptionSource} from "../e2ee/common.js";
 import {ensureLogItem} from "../../logging/utils";
 import {PowerLevels} from "./PowerLevels.js";
 import {RetainedObservableValue} from "../../observable/ObservableValue";
-import {NonPersistedEventEntry} from "./timeline/entries/NonPersistedEventEntry";
+import {TimelineReader} from "./timeline/persistence/TimelineReader";
 
 const EVENT_ENCRYPTED_TYPE = "m.room.encrypted";
 
@@ -503,8 +503,7 @@ export class BaseRoom extends EventEmitter {
                 clock: this._platform.clock,
                 logger: this._platform.logger,
                 powerLevelsObservable: await this.observePowerLevels(),
-                fetchEventFromStorage: eventId => this._readEventById(eventId),
-                fetchEventFromHomeserver: eventId => this._getEventFromHomeserver(eventId)
+                hsApi: this._hsApi
             });
             try {
                 if (this._roomEncryption) {
@@ -546,39 +545,10 @@ export class BaseRoom extends EventEmitter {
     }
 
     async _readEventById(eventId) {
-        let stores = [this._storage.storeNames.timelineEvents];
-        if (this.isEncrypted) {
-            stores.push(this._storage.storeNames.inboundGroupSessions);
-        }
-        const txn = await this._storage.readTxn(stores);
-        const storageEntry = await txn.timelineEvents.getByEventId(this._roomId, eventId);
-        if (storageEntry) {
-            const entry = new EventEntry(storageEntry, this._fragmentIdComparer);
-            if (entry.eventType === EVENT_ENCRYPTED_TYPE) {
-                const request = this._decryptEntries(DecryptionSource.Timeline, [entry], txn);
-                await request.complete();
-            }
-            return entry;
-        }
+        const reader = new TimelineReader({ roomId: this._roomId, storage: this._storage, fragmentIdComparer: this._fragmentIdComparer });
+        const entry = await reader.readById(eventId);
+        return entry;
     }
-
-    async _getEventFromHomeserver(eventId) {
-        const response = await this._hsApi.context(this._roomId, eventId, 0).response();
-        const sender = response.event.sender;
-        const member = response.state.find(e => e.type === "m.room.member" && e.user_id === sender);
-        const entry = {
-            event: response.event,
-            displayName: member.content.displayname,
-            avatarUrl: member.content.avatar_url
-        };
-        const eventEntry = new NonPersistedEventEntry(entry, this._fragmentIdComparer);
-        if (eventEntry.eventType === EVENT_ENCRYPTED_TYPE) {
-            const request = this._decryptEntries(DecryptionSource.Timeline, [eventEntry]);
-            await request.complete();
-        }
-        return eventEntry;
-    }
-
 
     dispose() {
         this._roomEncryption?.dispose();
--- a/src/matrix/room/timeline/Timeline.js
+++ b/src/matrix/room/timeline/Timeline.js
@@ -259,14 +259,30 @@ export class Timeline {
     /**
      * Update entries based on newly received events.
      * eg: a newly received redacted event may mark an existing event in contextEntriesNotInTimeline as being redacted 
+     * This is only for the events that are not in the timeline but had to fetched from elsewhere to render reply previews.
      */
     _updateFetchedEntries(entries) {
         for (const entry of entries) {
-            const relatedEntry = this._getTrackedEntry(entry.relatedEventId);
-            // todo: can this be called .addRelation instead?
-            if (relatedEntry?.addLocalRelation(entry)) {
-                relatedEntry.contextForEntries?.forEach(e => this._updateEntry(e));
+            const relatedEntry = this._contextEntriesNotInTimeline.get(entry.relatedEventId);
+            if (!relatedEntry) {
+                continue;
             }
+            // update dependents with this new entry indicating that this is an update to contextEntry
+            const newEntry = this._createEntryFromRelatedEntries(entry, relatedEntry);
+            if (newEntry) {
+                Timeline._entryUpdater(relatedEntry, newEntry);
+                relatedEntry.contextForEntries?.forEach(e => {
+                    this._remoteEntries.findAndUpdate(te => te.id === e.id, () => { return { reply: newEntry }; });
+                });
+            }
+        }
+    }
+
+    _createEntryFromRelatedEntries(entry, relatedEntry) {
+        if (entry.isRedaction) {
+            const newEntry = relatedEntry.clone();
+            newEntry.setAsRedacted();
+            return newEntry;
         }
     }
 
--- a/src/matrix/room/timeline/Timeline.js
+++ b/src/matrix/room/timeline/Timeline.js
@@ -280,6 +280,7 @@ export class Timeline {
             if (fetchedEntry) {
                 fetchedEntry.contextForEntries.forEach(e => e.setContextEntry(entry));
                 entry.updateFrom(fetchedEntry);
+                this._contextEntriesNotInTimeline.delete(entry.id);
             }
         }
     }
--- a/src/matrix/room/timeline/Timeline.js
+++ b/src/matrix/room/timeline/Timeline.js
@@ -223,7 +223,7 @@ export class Timeline {
     /** @package */
     replaceEntries(entries) {
         this._addLocalRelationsToNewRemoteEntries(entries);
-        this._updateFetchedEntries(entries);
+        this._updateEntriesNotInTimeline(entries);
         this._loadRelatedEvents(entries);
         for (const entry of entries) {
             try {
@@ -250,18 +250,18 @@ export class Timeline {
     /** @package */
     addEntries(newEntries) {
         this._addLocalRelationsToNewRemoteEntries(newEntries);
-        this._updateFetchedEntries(newEntries);
-        this._moveFetchedEntryToRemoteEntries(newEntries);
+        this._updateEntriesNotInTimeline(newEntries);
+        this._moveEntryToRemoteEntries(newEntries);
         this._remoteEntries.setManySorted(newEntries);
         this._loadRelatedEvents(newEntries);
     }
 
     /**
      * Update entries based on newly received events.
-     * eg: a newly received redacted event may mark an existing event in contextEntriesNotInTimeline as being redacted 
+     * eg: a newly received redaction event may mark an existing event in contextEntriesNotInTimeline as being redacted
      * This is only for the events that are not in the timeline but had to fetched from elsewhere to render reply previews.
      */
-    _updateFetchedEntries(entries) {
+    _updateEntriesNotInTimeline(entries) {
         for (const entry of entries) {
             const relatedEntry = this._contextEntriesNotInTimeline.get(entry.relatedEventId);
             if (!relatedEntry) {
@@ -292,7 +292,7 @@ export class Timeline {
      * If an event we had to fetch from hs/storage is now in the timeline (for eg, due to gap fill),
      * remove the event from _contextEntriesNotInTimeline since it is now in remoteEntries
      */
-    _moveFetchedEntryToRemoteEntries(entries) {
+    _moveEntryToRemoteEntries(entries) {
         for (const entry of entries) {
             const fetchedEntry = this._contextEntriesNotInTimeline.get(entry.id);
             if (fetchedEntry) {
--- a/src/matrix/room/timeline/Timeline.js
+++ b/src/matrix/room/timeline/Timeline.js
@@ -271,7 +271,6 @@ export class Timeline {
             const newEntry = this._createEntryFromRelatedEntries(entry, relatedEntry);
             if (newEntry) {
                 Timeline._entryUpdater(relatedEntry, newEntry);
-                this._contextEntriesNotInTimeline.delete(relatedEntry.id);
                 this._contextEntriesNotInTimeline.set(newEntry.id, newEntry);
                 relatedEntry.contextForEntries?.forEach(e => {
                     this._remoteEntries.findAndUpdate(te => te.id === e.id, () => { return { reply: newEntry }; });
--- a/src/matrix/room/timeline/Timeline.js
+++ b/src/matrix/room/timeline/Timeline.js
@@ -485,10 +485,10 @@ export function tests() {
             // 2. test replaceEntries and addEntries don't fail
             const event1 = withTextBody("hi!", withSender(bob, createEvent("m.room.message", "!abc")));
             const entry1 = new EventEntry({event: event1, fragmentId: 1, eventIndex: 1}, fragmentIdComparer);
-            await timeline.replaceEntries([entry1]);
+            timeline.replaceEntries([entry1]);
             const event2 = withTextBody("hi bob!", withSender(alice, createEvent("m.room.message", "!def")));
             const entry2 = new EventEntry({event: event2, fragmentId: 1, eventIndex: 2}, fragmentIdComparer);
-            await timeline.addEntries([entry2]);
+            timeline.addEntries([entry2]);
             // 3. add local relation (redaction)
             pendingEvents.append(new PendingEvent({data: {
                 roomId,
@@ -512,7 +512,7 @@ export function tests() {
             await timeline.load(new User(bob), "join", new NullLogItem());
             timeline.entries.subscribe(new ListObserver());
             const event = withTextBody("hi bob!", withSender(alice, createEvent("m.room.message", "!abc")));
-            await timeline.addEntries([new EventEntry({event, fragmentId: 1, eventIndex: 2}, fragmentIdComparer)]);
+            timeline.addEntries([new EventEntry({event, fragmentId: 1, eventIndex: 2}, fragmentIdComparer)]);
             let entry = getIndexFromIterable(timeline.entries, 0);
             // 2. add local reaction
             pendingEvents.append(new PendingEvent({data: {
@@ -583,7 +583,7 @@ export function tests() {
             await timeline.load(new User(bob), "join", new NullLogItem());
             timeline.entries.subscribe(new ListObserver());
             // 3. add message to timeline
-            await timeline.addEntries([messageEntry]);
+            timeline.addEntries([messageEntry]);
             const entry = getIndexFromIterable(timeline.entries, 0);
             assert.equal(entry, messageEntry);
             assert.equal(entry.annotations["👋"].count, 1);
@@ -605,7 +605,7 @@ export function tests() {
                 event: withContent(createAnnotation(messageEntry.id, "👋"), createEvent("m.reaction", "!def", bob)),
                 fragmentId: 1, eventIndex: 3, roomId
             }, fragmentIdComparer);
-            await timeline.addEntries([messageEntry, reactionEntry]);
+            timeline.addEntries([messageEntry, reactionEntry]);
             // 3. redact reaction
             pendingEvents.append(new PendingEvent({data: {
                 roomId,
@@ -638,7 +638,7 @@ export function tests() {
             }}));
             await poll(() => timeline.entries.length === 1);
             // 3. add remote reaction target
-            await timeline.addEntries([messageEntry]);
+            timeline.addEntries([messageEntry]);
             await poll(() => timeline.entries.length === 2);
             const entry = getIndexFromIterable(timeline.entries, 0);
             assert.equal(entry, messageEntry);
@@ -672,7 +672,7 @@ export function tests() {
             }}));
             await poll(() => timeline.entries.length === 1);
             // 4. add reaction target
-            await timeline.addEntries([new EventEntry({
+            timeline.addEntries([new EventEntry({
                 event: withTextBody("hi bob!", createEvent("m.room.message", messageId, alice)),
                 fragmentId: 1, eventIndex: 2}, fragmentIdComparer)
             ]);
@@ -695,14 +695,14 @@ export function tests() {
             const timeline = new Timeline({roomId, storage: await createMockStorage(), closeCallback: () => {},
                 fragmentIdComparer, pendingEvents: new ObservableArray(), clock: new MockClock()});
             await timeline.load(new User(alice), "join", new NullLogItem());
-            await timeline.addEntries([decryptedEntry]);
+            timeline.addEntries([decryptedEntry]);
             const observer = new ListObserver();
             timeline.entries.subscribe(observer);
             // 3. replace the entry with one that is not decrypted
             //    (as would happen when receiving a reaction,
             //    as it does not rerun the decryption)
             //    and check that the decrypted content is preserved
-            await timeline.replaceEntries([encryptedEntry]);
+            timeline.replaceEntries([encryptedEntry]);
             const {value, type} = await observer.next();
             assert.equal(type, "update");
             assert.equal(value.eventType, "m.room.message");
@@ -808,7 +808,7 @@ export function tests() {
             await timeline.load(new User(alice), "join", new NullLogItem());
             await timeline._loadRelatedEvents([entryB]);
             const redactingEntry = new EventEntry({ event: withRedacts(entryA.id, "foo", createEvent("m.room.redaction", "event_id_3", alice)) });
-            await timeline.addEntries([redactingEntry]);
+            timeline.addEntries([redactingEntry]);
             const contextEntry = timeline._contextEntriesNotInTimeline.get(entryA.id);
             assert.strictEqual(contextEntry.isRedacted, true);
         },
@@ -830,7 +830,8 @@ export function tests() {
             const entryC = new EventEntry({ event: withContent(content, createEvent("m.room.message", "event_id_3", bob)) });
             await timeline.load(new User(alice), "join", new NullLogItem());
             timeline._getEventFromStorage = () => entryA;
-            await timeline.addEntries([entryB, entryC]);
+            timeline.addEntries([entryB, entryC]);
+            await poll(() => timeline._remoteEntries.array.length === 2);
             const bin = [entryB, entryC];
             timeline.entries.subscribe({
                 onUpdate: (index) => {
@@ -840,7 +841,7 @@ export function tests() {
                 onAdd: () => { }
             });
             const redactingEntry = new EventEntry({ event: withRedacts(entryA.id, "foo", createEvent("m.room.redaction", "event_id_3", alice)) });
-            await timeline.addEntries([redactingEntry]);
+            timeline.addEntries([redactingEntry]);
             assert.strictEqual(bin.length, 0);
         },
 
@@ -860,10 +861,10 @@ export function tests() {
             const entryB = new EventEntry({ event });
             timeline._getEventFromHomeserver = () => entryA
             await timeline.load(new User(alice), "join", new NullLogItem());
-            await timeline.addEntries([entryB]);
+            timeline.addEntries([entryB]);
             await timeline._loadRelatedEvents([entryB]);
             assert.strictEqual(timeline._contextEntriesNotInTimeline.has(entryA.id), true);
-            await timeline.addEntries([entryA]);
+            timeline.addEntries([entryA]);
             assert.strictEqual(timeline._contextEntriesNotInTimeline.has(entryA.id), false);
             const movedEntry = timeline.remoteEntries[0];
             assert.deepEqual(movedEntry, entryA);
--- a/src/matrix/room/timeline/Timeline.js
+++ b/src/matrix/room/timeline/Timeline.js
@@ -267,11 +267,11 @@ export class Timeline {
             if (!relatedEntry) {
                 continue;
             }
-            // update dependents with this new entry indicating that this is an update to contextEntry
             const newEntry = this._createEntryFromRelatedEntries(entry, relatedEntry);
             if (newEntry) {
                 Timeline._entryUpdater(relatedEntry, newEntry);
                 this._contextEntriesNotInTimeline.set(newEntry.id, newEntry);
+                // update other entries for which this entry is a context entry
                 relatedEntry.contextForEntries?.forEach(e => {
                     this._remoteEntries.findAndUpdate(te => te.id === e.id, () => { return { reply: newEntry }; });
                 });
--- a/src/matrix/room/timeline/Timeline.js
+++ b/src/matrix/room/timeline/Timeline.js
@@ -842,6 +842,33 @@ export function tests() {
             const redactingEntry = new EventEntry({ event: withRedacts(entryA.id, "foo", createEvent("m.room.redaction", "event_id_3", alice)) });
             await timeline.addEntries([redactingEntry]);
             assert.strictEqual(bin.length, 0);
+        },
+
+        "context entries fetched from storage/hs are moved to remoteEntries": async assert => {
+            const timeline = new Timeline({roomId, storage: await createMockStorage(), closeCallback: () => {},
+                fragmentIdComparer, pendingEvents: new ObservableArray(), clock: new MockClock()});
+            const entryA = new EventEntry({ event: withTextBody("foo", createEvent("m.room.message", "event_id_1", alice)) });
+            let event = withContent({
+                body: "bar",
+                msgtype: "m.text",
+                "m.relates_to": {
+                    "m.in_reply_to": {
+                        "event_id": "event_id_1"
+                    }
+                }
+            }, createEvent("m.room.message", "event_id_2", bob));
+            const entryB = new EventEntry({ event });
+            timeline._getEventFromHomeserver = () => entryA
+            await timeline.load(new User(alice), "join", new NullLogItem());
+            await timeline.addEntries([entryB]);
+            await timeline._loadRelatedEvents([entryB]);
+            assert.strictEqual(timeline._contextEntriesNotInTimeline.has(entryA.id), true);
+            await timeline.addEntries([entryA]);
+            assert.strictEqual(timeline._contextEntriesNotInTimeline.has(entryA.id), false);
+            const movedEntry = timeline.remoteEntries[0];
+            assert.deepEqual(movedEntry, entryA);
+            assert.deepEqual(movedEntry.contextForEntries.pop(), entryB);
+            assert.deepEqual(entryB.contextEntry, movedEntry);
         }
     };
 }
--- a/src/matrix/room/timeline/Timeline.js
+++ b/src/matrix/room/timeline/Timeline.js
@@ -271,6 +271,8 @@ export class Timeline {
             const newEntry = this._createEntryFromRelatedEntries(entry, relatedEntry);
             if (newEntry) {
                 Timeline._entryUpdater(relatedEntry, newEntry);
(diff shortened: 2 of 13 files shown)


src/matrix/room/BaseRoom.js
```diff
@@ -501,7 +501,9 @@ export class BaseRoom extends EventEmitter {
                 },
                 clock: this._platform.clock,
                 logger: this._platform.logger,
-                powerLevelsObservable: await this.observePowerLevels()
+                powerLevelsObservable: await this.observePowerLevels(),
+                fetchEventFromStorage: eventId => this._readEventById(eventId),
```
bwindels: why inject this? the homeserver already has access to storage ...

src/matrix/room/BaseRoom.js
```diff
@@ -501,7 +501,9 @@ export class BaseRoom extends EventEmitter {
...
+                fetchEventFromHomeserver: eventId => this._getEventFromHomeserver(eventId)
```
bwindels: I think we can also consider to pass hsApi into timeline for this... unless there's a good reason to inject? Unit tests perhaps?

src/matrix/room/timeline/Timeline.js
```diff
@@ -211,8 +214,9 @@ export class Timeline {
     }
 
     /** @package */
-    replaceEntries(entries) {
+    async replaceEntries(entries) {
         this._addLocalRelationsToNewRemoteEntries(entries);
+        await this._loadRelatedEvents(entries);
```
bwindels: I'm slightly worried about this blocking getting the initial event add/update out, Especially in the worst case scenario where we hit the homeserver to get the context event ... Perhaps we could not await this call and send an update (e.g. call replaceEntries again) once we have obtained the context event.

src/matrix/room/timeline/Timeline.js
```diff
@@ -234,10 +238,32 @@ export class Timeline {
     }
 
     /** @package */
-    addEntries(newEntries) {
+    async addEntries(newEntries) {
         this._addLocalRelationsToNewRemoteEntries(newEntries);
+        await this._loadRelatedEvents(newEntries);
```
bwindels: Same here, slightly worried about blocking here.

src/matrix/room/timeline/Timeline.js
```diff
@@ -234,10 +238,32 @@ export class Timeline {
...
         this._remoteEntries.setManySorted(newEntries);
     }
+
+    async _loadRelatedEvents(entries) {
+        const filteredEntries = entries.filter(e => !!e.relation);
+        for (const entry of filteredEntries) {
+            const id = entry.relatedEventId;
```
bwindels: Two things:
 - the id can be undefined/null
 - I don't think we should fetch the event for every relation. This would include for example the reacted-to-event for reactions (which are present at this stage in the timeline, but filtered out in TilesCollection later on). I would add a separate property (e.g. `contextEventId`, adding a comment on top of the getter to say what it is for) that only returns the event id in case of a reply.

src/matrix/room/timeline/entries/EventEntry.js
```diff
@@ -41,6 +42,10 @@ export class EventEntry extends BaseEventEntry {
         }
     }
 
+    setRelatedEntry(entry) {
```
bwindels: Wrt to the earlier comment, perhaps this can be called contextEntry?

bwindels (commented):
good start! first quick review, left some comments

src/matrix/room/BaseRoom.js
MidhunSureshR: `_readEventById()` is already defined in BaseRoom (and is used in `BaseRoom.observeEvent()`?
We wouldn't want to duplicate the same logic in Timeline?

src/matrix/room/BaseRoom.js
MidhunSureshR: Felt more cohesive to have this code in the same place that we have the code to fetch an event from storage. Would encapsulating these two methods into an `eventFetcher` be more agreeable?

src/matrix/room/timeline/Timeline.js
```diff
@@ -261,6 +261,8 @@ export class Timeline {
             }
             if (contextEvent) {
                 entry.setContextEntry(contextEvent);
+                // emit this change
+                this._remoteEntries.update(entry);
```
MidhunSureshR: I'm not sure if this is safe to do... is there a race between sortedArray operations done by `replaceEntries`/`addEntries` and this update :thinking: 
Maybe we need to do:
```js
this._remoteEntries.getAndUpdate(entry, (entry, existingEntry) => existingEntry);
```

src/matrix/room/BaseRoom.js
bwindels: Ahh, yes, I see now. This takes care of decryption and everything :+1:

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,8 +247,50 @@ export class Timeline {
...
+        for (const entry of entries) {
+            const relatedEntry = this._fetchedEventEntries.find(e => e.id === entry.relatedEventId);
+            if (relatedEntry?.addLocalRelation(entry)) {
+                relatedEntry.dependents.forEach(e => this._findAndUpdateRelatedEntry(null, e.id, () => true));
+            }
+        }
+    }
+
+    async _loadRelatedEvents(entries) {
+        const filteredEntries = entries.filter(e => !!e.contextEventId);
```
bwindels:
```suggestion
        const entriesNeedingContext = entries.filter(e => !!e.contextEventId);
```

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,8 +247,50 @@ export class Timeline {
...
+            }
+            // fetch from hs
+            if (!contextEvent) {
+                contextEvent = await this._fetchEventFromHomeserver(id);
+            }
+            if (contextEvent) {
+                if (needToTrack) {
+                    // this entry was created from storage/hs, so it's not tracked by remoteEntries
+                    // we track them here so that we can update reply preview of dependents on redaction
+                    this._fetchedEventEntries.push(contextEvent);
```
bwindels: If this only needs to happen if the entry comes from storage, can we move this below `await this._fetchEventFromStorage(id);` on line 275? Looks like we wouldn't need the `needToTrack` flag then? If you can replace the flag with control flow, that might be more readable.

OTOH, I wonder why we wouldn't need to track if the event came from the homeserver? Isn't it possible an event we don't have in storage gets redacted?

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,8 +247,50 @@ export class Timeline {
...
+            // find in remote events
+            contextEvent = this.getByEventId(id);
+            // find in storage
+            if (!contextEvent) {
+                needToTrack = true;
+                contextEvent = await this._fetchEventFromStorage(id);
+            }
+            // fetch from hs
+            if (!contextEvent) {
+                contextEvent = await this._fetchEventFromHomeserver(id);
```
bwindels: Shouldn't we also look into `_fetchedEventEntries` entries here first?

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,8 +247,50 @@ export class Timeline {
...
+            if (contextEvent) {
+                if (needToTrack) {
+                    // this entry was created from storage/hs, so it's not tracked by remoteEntries
+                    // we track them here so that we can update reply preview of dependents on redaction
+                    this._fetchedEventEntries.push(contextEvent);
+                }
+                contextEvent.addDependent(entry);
+                entry.setContextEntry(contextEvent);
+                // emit this change
+                this._findAndUpdateRelatedEntry(null, entry.id, () => true);
```
bwindels:
```suggestion
                const txnId = entry.isPending? entry.id : null;
                const eventId = entry.isPending? null : entry.id;
                this._findAndUpdateRelatedEntry(txnId, eventId, () => true);
```

src/matrix/room/timeline/Timeline.js
MidhunSureshR: I'm using this for now:
```js
this._findAndUpdateRelatedEntry(null, entry.id, () => true); 
```

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,8 +247,50 @@ export class Timeline {
...
+        for (const entry of filteredEntries) {
+            const id = entry.contextEventId;
+            let needToTrack = false;
+            let contextEvent;
+            // find in remote events
+            contextEvent = this.getByEventId(id);
+            // find in storage
+            if (!contextEvent) {
+                needToTrack = true;
+                contextEvent = await this._fetchEventFromStorage(id);
```
bwindels: perhaps we could also use this in _getOrLoadEntry?

src/matrix/room/BaseRoom.js
bwindels: Hmm, thinking about this a bit more, I wonder if we should move _readEventById into `TimelineReader` as a public method? In _readEventById in here we would just create a new ephemeral TimelineReader to read a single event, and in the timeline we have a TimelineReader as well. TimelineReader already takes care of decryption as well.

src/matrix/room/BaseRoom.js
bwindels: So if we end up with readEventById in TimelineReader, I would put this in the Timeline. I don't thing we will ever really need to call /context from outside of the Timeline.

src/matrix/room/BaseRoom.js
```diff
@@ -559,6 +562,23 @@ export class BaseRoom extends EventEmitter {
         }
     }
 
+    async _getEventFromHomeserver(eventId) {
+        const response = await this._hsApi.context(this._roomId, eventId, 0).response();
+        const sender = response.event.sender;
+        const member = response.state.find(e => e.type === "m.room.member" && e.user_id === sender);
```
bwindels: you can import the "m.room.member" string from "./members/RoomMember.js" if you wish.

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,8 +247,50 @@ export class Timeline {
...
     addEntries(newEntries) {
         this._addLocalRelationsToNewRemoteEntries(newEntries);
+        this._updateFetchedEntries(newEntries);
+        this._loadRelatedEvents(newEntries);
         this._remoteEntries.setManySorted(newEntries);
     }
+
+    _updateFetchedEntries(entries) {
+        for (const entry of entries) {
+            const relatedEntry = this._fetchedEventEntries.find(e => e.id === entry.relatedEventId);
```
bwindels: Should _fetchedEventEntries be a `Map` rather? Seems we would only need to look things up by id and never iterate otherwise?

src/matrix/room/timeline/Timeline.js
bwindels: Can we rename _findAndUpdateRelatedEntry to _findAndUpdateEntryById (and the local variables called related... therein)? It doesn't really do anything specific for relations and the name is a bit confusing here.

bwindels: Also, you could put the above in a `_updateEntry` method.

bwindels (commented):
I'm thinking of the nomenclature introduced by this PR, so far:
 - context event: the event entry on a related entry that we need to display ourselves (on a reply)
 - dependants: the inverse relation of context events, e.g. target events that are mentioned as a context event on their relations (replies), needed to notify the dependent entries that their context event has changed.
 - fetched entries: context events that are not in the timeline, and tracked separatedly so they can trigger an update in their dependents when they get updated (e.g redacted)

Perhaps we could change it to:
 - stick with the term context (event)
 - rename dependants to `contextForEntries`
 - rename fetched entries to `contextEntriesNotInTimeline` or `contextEntriesOutOfScope`?

I'm also thinking:
 - we should somehow document (comments or elsewhere) this
 -  if we can possibly come up with something a bit more coherent

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,8 +247,50 @@ export class Timeline {
...
+            if (relatedEntry?.addLocalRelation(entry)) {
```
bwindels: this is to make fetched event appear as redacted (and later edited) when we get a redaction? perhaps add a comment saying as much.

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,8 +247,50 @@ export class Timeline {
...
+                relatedEntry.dependents.forEach(e => this._findAndUpdateRelatedEntry(null, e.id, () => true));
```
bwindels: Could use the elsewhere proposed `_updateEntry` method here.

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,8 +247,50 @@ export class Timeline {
     /** @package */
     addEntries(newEntries) {
         this._addLocalRelationsToNewRemoteEntries(newEntries);
+        this._updateFetchedEntries(newEntries);
```
bwindels: addEntries is not just called for sync, but also for gap fills. So it is possible that an event we previously didn't know about has now been added to the storage and timeline by a gap fill. Should we transition the fetched entry to a remote event entry from the timeline in all of the dependants?

bwindels (commented):
some more thoughts/questions

src/matrix/room/timeline/Timeline.js
MidhunSureshR: Yes we should. Otherwise the dependents of the entry created before gap fill will become dead/forgotten.

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +250,109 @@ export class Timeline {
...
+                this._contextEntriesNotInTimeline.set(newEntry.id, newEntry);
+                // update other entries for which this entry is a context entry
+                relatedEntry.contextForEntries?.forEach(e => {
+                    this._remoteEntries.findAndUpdate(te => te.id === e.id, () => { return { reply: newEntry }; });
+                });
+            }
+        }
+    }
+
+    _createEntryFromRelatedEntries(entry, relatedEntry) {
```
bwindels: This seems unused?

src/matrix/room/timeline/entries/EventEntry.js
```diff
@@ -39,6 +42,27 @@ export class EventEntry extends BaseEventEntry {
...
+    }
+
+    setAsContextOf(entry) {
+        if (!this._contextForEntries) {
+            this._contextForEntries = [];
+        }
+        this._contextForEntries.push(entry);
+    }
+
+    setAsRedacted() {
```
bwindels: If `_createEntryFromRelatedEntries` is unused, this is unused as well

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +250,109 @@ export class Timeline {
...
+            }
+            if (contextEvent) {
+                contextEvent.setAsContextOf(entry);
+                entry.setContextEntry(contextEvent);
+                this._updateEntry(entry);
+            }
+        }
+    }
+
+    _getTrackedEntry(id) {
```
bwindels: what does tracked mean here? is there perhaps a more appropriate name we can pick here?

bwindels: is id always an eventId? if so, consider renaming it

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +250,109 @@ export class Timeline {
...
+        }
+    }
+
+    _updateEntry(entry) {
+        const txnId = entry.isPending ? entry.id : null;
+        const eventId = entry.isPending ? null : entry.id;
+        this._findAndUpdateEntryById(txnId, eventId, () => true);
+    }
+
+    async _loadRelatedEvents(entries) {
```
bwindels: I feel this name is both too vague and not using the terminology used elsewhere ("context" rather than "related"), which makes it hard to guess what this method does when seeing it used elsewhere in this class. What about `_loadContextEntriesWhereNeeded`?

src/matrix/room/timeline/Timeline.js
```diff
@@ -43,6 +46,10 @@ export class Timeline {
         });
         this._readerRequest = null;
         this._allEntries = null;
+        /** Stores event entries that we had to fetch from hs/storage for reply previews (because they were not in timeline) */ 
+        this._contextEntriesNotInTimeline = new Map();
+        this._decryptEntries = null;
```
bwindels, line 52: perhaps add a comment that this is only used for decrypting non-persisted context entries loaded from the HS?

bwindels (commented):
seems like there is a merge conflict, and if I'm not missing anything some unused code. Perhaps best to go over it one last time to see everything is ready for final review?

src/matrix/room/timeline/Timeline.js
MidhunSureshR: It's used in [line 270, Timeline.js](https://github.com/vector-im/hydrogen-web/pull/611/files#diff-95a8d7280cef101142b1363df405903543be8531502e1b4aee38f208fb9a0ac0R270)

src/matrix/room/timeline/Timeline.js
MidhunSureshR: > what does tracked mean here? is there perhaps a more appropriate name we can pick here?

tracked as in we track two kinds of entries:
- **remoteEntries** for entries in the timeline
- **contextEntriesNotInTimeline** for entries we had to manually fetch to render replies

I've added a jsdoc comment saying as much.

src/matrix/room/timeline/entries/EventEntry.js
```diff
@@ -24,6 +24,9 @@ export class EventEntry extends BaseEventEntry {
         this._eventEntry = eventEntry;
         this._decryptionError = null;
         this._decryptionResult = null;
+        this._contextEntry = null;
+        this._contextForEntries = null;
```
bwindels: I'm wondering if we can avoid putting a new member variable in this class for a very specific use case (which tends to have a lot of instances, and I'd like to keep light). What about that instead of putting it here, we put the array of entries we are context for into `Timeline._contextEntriesNotInTimeline`, by changing the value from an entry to an object (of a new class, e.g. `ContextEntryLinks` or something) with the context entry and an array of entries for which the former entry is the context. Does that make sense?

bwindels: Ah, this won't work because we only put the context entry in `_contextEntriesNotInTimeline` if it wasn't in the remoteEntries already, but in case of an update, we need to know which entries are using the context entry regardless of whether it came from remoteEntries or storage/hs.

bwindels: Ok, never mind, let's keep it as it is :+1:

src/matrix/room/timeline/entries/EventEntry.js
```diff
@@ -39,6 +42,27 @@ export class EventEntry extends BaseEventEntry {
...
         }
+        this._contextForEntries = other.contextForEntries;
+        this._contextEntry = other.contextEntry;
+    }
+
+    setContextEntry(entry) {
+        this._contextEntry = entry;
+    }
+
+    setAsContextOf(entry) {
```
bwindels: can we make this method private and call it from `setContextEntry` above on the `entry` passed in? that way it's a little less noisy in Timeline...

src/matrix/room/timeline/entries/EventEntry.js
bwindels: I'm still wondering if adding a third way to mark an event as redacted is the best way to go here. Also, we only really need this method in `NonPersistedEventEntry` (as context entries not in the timeline but fetched from storage would get redacted by the relation writer and be passed to replaceEntries). So how about we add a `redact` method there (perhaps with a comment saying that redaction needs a method because non persisted entries don't go through relation writer) that runs the [redaction algorithm](https://github.com/vector-im/hydrogen-web/blob/master/src/matrix/room/timeline/persistence/RelationWriter.js#L130-L143) (which will need to be extracted from `RelationWriter`, perhaps to room/common.ts?) on a copy of the event and creates a new NonPersistedEventEntry with it.

Then we can get rid of this method and the flag.

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +251,128 @@ export class Timeline {
...
+    }
+
+    /**
+     * Update entries based on newly received events.
+     * eg: a newly received redaction event may mark an existing event in contextEntriesNotInTimeline as being redacted
+     * This is only for the events that are not in the timeline but had to fetched from elsewhere to render reply previews.
+     */
+    _updateEntriesNotInTimeline(entries) {
+        for (const entry of entries) {
+            const relatedEntry = this._contextEntriesNotInTimeline.get(entry.relatedEventId);
```
bwindels, line 278: This doesn't seem to take into account that a context entry that came from storage (but not in the timeline) will still be updated by the syncwriter, and the entry that is passed in will actually be the relation target.

So let's say we have a message E. Then there is a reply F to message E. E wasn't in the timeline, so is put into _contextEntriesNotInTimeline. Now a redaction G is sent to redact message E. When G is received by our client through sync, E is known in storage. So the sync writer, delegating to the relation writer, will fetch and update E from storage, and thus an already redacted version of E will be passed to replaceEntries, while G will be passed to addEntries.

I'm not sure it makes sense to call the same method from both addEntries and replaceEntries for this purpose. I think we should do what you already do for events that are not in storage (e.g. fetched from the homeserver) in addEntries, but we should also look for entry.id in _contextEntriesNotInTimeline for the above case in replaceEntries.

Does that make sense?

bwindels (commented):
Thanks for the PR description, that is very helpful!
A few more comments after another read of the code.

src/matrix/room/timeline/Timeline.js
```diff
@@ -206,16 +215,22 @@ export class Timeline {
...
+        existingEntry.contextForEntries?.forEach(event => event.setContextEntry(entry));
         entry.updateFrom(existingEntry);
         return entry;
     }
 
     /** @package */
     replaceEntries(entries) {
         this._addLocalRelationsToNewRemoteEntries(entries);
+        this._updateEntriesNotInTimeline(entries);
+        this._loadContextEntriesWhereNeeded(entries);
```
bwindels: Not sure why we would need to load context after an event has been updated in any way (redacted mostly, and later on edited)? What could update an entry that changes the context entry that it needs? Relations can't be edited AFAIK...

src/matrix/room/timeline/entries/EventEntry.js
MidhunSureshR: We'd still want it to be public though:
```js
    setContextEntry(entry) {
        this._contextEntry = entry;
        entry.setAsContextOf(this);
    }

    setAsContextOf(entry) {
        if (!this._contextForEntries) {
            this._contextForEntries = [];
        }
        this._contextForEntries.push(entry);
    }
```

bwindels: why is that? is it called from anywhere else now but setContextEntry?

src/matrix/room/timeline/Timeline.js
```diff
@@ -292,7 +296,7 @@ export class Timeline {
     _createEntryFromRelatedEntries(entry, relatedEntry) {
         if (entry.isRedaction) {
             const newEntry = relatedEntry.clone();
-            newEntry.setAsRedacted();
+            newEntry.redact(entry);
```
bwindels, line 299: newEntry isn't a NonPersistedEventEntry here when it was fetched from storage (but not in the timeline).

src/matrix/room/timeline/entries/EventEntry.js
MidhunSureshR: nvm, I'm an idiot.

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +254,111 @@ export class Timeline {
...
         this._addLocalRelationsToNewRemoteEntries(newEntries);
+        this._updateEntriesFetchedFromHomeserver(newEntries);
+        this._moveEntryToRemoteEntries(newEntries);
         this._remoteEntries.setManySorted(newEntries);
+        this._loadContextEntriesWhereNeeded(newEntries);
+    }
+
+    /**
+     * Update entries based on newly received events.
+     * This is specific to events that are not in the timeline but had to be fetched from the homeserver.
```
bwindels: this comment on its own seems confusing - can you mention that the events had to be fetched from the homeserver because they are context entries for other events in the timeline?

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +254,111 @@ export class Timeline {
...
+         * Updates for entries in timeline is handled by remoteEntries observable collection
+         * Updates for entries not in timeline but fetched from storage is handled in this.replaceEntries()
+         * This code is specific to entries fetched from HomeServer i.e NonPersistedEventEntry
+         */
+        for (const entry of entries) {
+            const relatedEntry = this._contextEntriesNotInTimeline.get(entry.relatedEventId);
+            if (relatedEntry?.addRelation(entry)) {
+                // update other entries for which this entry is a context entry
+                relatedEntry.contextForEntries?.forEach(e => {
+                    this._remoteEntries.findAndUpdate(te => te.id === e.id, () => { return { reply: relatedEntry }; });
```
bwindels: I thought you didn't need reply in the end here as a param? You could just pass the string "contextEntry", the name of the property that changed, which is what is usually used as a param.

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +254,111 @@ export class Timeline {
...
+     */
+    _updateEntriesFetchedFromHomeserver(entries) {
+        /**
+         * Updates for entries in timeline is handled by remoteEntries observable collection
+         * Updates for entries not in timeline but fetched from storage is handled in this.replaceEntries()
+         * This code is specific to entries fetched from HomeServer i.e NonPersistedEventEntry
+         */
+        for (const entry of entries) {
+            const relatedEntry = this._contextEntriesNotInTimeline.get(entry.relatedEventId);
+            if (relatedEntry?.addRelation(entry)) {
```
bwindels: Won't this also add local relations for context entries that came from storage? Those as mentioned in the comment above will already be updated by replaceEntries, and could potentially see their relations double counted like this?

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +254,111 @@ export class Timeline {
...
+    /**
+     * If an event we had to fetch from hs/storage is now in the timeline (for eg, due to gap fill),
+     * remove the event from _contextEntriesNotInTimeline since it is now in remoteEntries
+     */
+    _moveEntryToRemoteEntries(entries) {
+        for (const entry of entries) {
+            const fetchedEntry = this._contextEntriesNotInTimeline.get(entry.id);
+            if (fetchedEntry) {
+                fetchedEntry.contextForEntries.forEach(e => e.setContextEntry(entry));
+                entry.updateFrom(fetchedEntry);
```
bwindels: Why do we need this? `contextForEntries` has been set by the previous line, not sure what else needs to be copied, as the entry that just came from the server has the latest event?

Also, to be completely consistent, we could also emit an update here for all the entries that have this entry as context? It likely wouldn't change anything on screen but it would ensure that the new entry is used everywhere down in the TilesCollection and the view, ...

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +254,111 @@ export class Timeline {
...
+                this._contextEntriesNotInTimeline.delete(entry.id);
+            }
+        }
+    }
+
+    _updateEntry(entry) {
```
bwindels: Would `_emitUpdateForEntry` be a better name? When I see it used elsewhere, I keep thinking that it updates the entry somehow inline, rather than just emitting to inform an update already happened.

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +254,111 @@ export class Timeline {
...
+    /**
+     * For each entry in entries, this method associates a context-entry (if needed) to it.
+     * The context-entry is fetched using the following strategies (in the same order as given):
+     * - timeline
+     * - storage
+     * - homeserver
+     * @param {EventEntry[]} entries 
+     */
+    async _loadContextEntriesWhereNeeded(entries) {
+        const entriesNeedingContext = entries.filter(e => !!e.contextEventId);
```
bwindels: Since you already have a loop below, it would be so easy to `break` with an `if` rather than to create a new array?

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +254,111 @@ export class Timeline {
...
+            }
+        }
+    }
+
+    /**
+     * Fetches an entry with the given event-id from remoteEntries or contextEntriesNotInTimeline.
+     * @param {string} eventId event-id of the entry
+     * @returns entry if found, undefined otherwise
+     */
+    _getTrackedEntry(eventId) {
```
bwindels: Still wonder if there is a less generic name for this ... what about `_findLoadedEventById`?

src/matrix/room/timeline/entries/BaseEventEntry.js
```diff
@@ -53,10 +53,13 @@ export class BaseEventEntry extends BaseEntry {
...
     /**
-        aggregates local relation or local redaction of remote relation.
+        Aggregates relation or redaction of remote relation.  
+        Used in two situations:
+        - to aggregate local relation/redaction of remote relation
+        - to mark this entry as being redacted in Timeline._updateEntriesFetchedFromHomeserver
         @return [string] returns the name of the field that has changed, if any
     */
-    addLocalRelation(entry) {
+    addRelation(entry) {
```
bwindels: The comment above is great, thanks, but not sure about this name change. It gives the impression that this is the only way relations are tracked, while 99% of the time, relations are tracked by RelationWriter, and this is just for some "edge" cases like local echo and when fetching something from /context.

If we can't come up with a better name (I'm not able to come up with one), I'd keep `addLocalRelation` tbh.

src/matrix/room/timeline/Timeline.js
```diff
@@ -216,6 +227,13 @@ export class Timeline {
         for (const entry of entries) {
             try {
                 this._remoteEntries.getAndUpdate(entry, Timeline._entryUpdater);
+                const oldEntry = this._contextEntriesNotInTimeline.get(entry.id)
```
bwindels, line 232: I wonder what will happen if you reply to a pending message (e.g. while offline), and then everything gets sent when  off-line. `id` returns the transaction id for `PendingEventEntry`s and the event_id for `EventEntry`s. Could it happen that we are left with both the PendingEventEntry and the EventEntry for the same message in the map, as they use a different id?

bwindels: Actually, in that scenario, the context entry will fail to load before it is sent, right? As:
 - it's not in remoteEntries
 - it's not in storage
 - it's not in the homeserver

bwindels: Can you test this scenario? We might need to also support getting the context entry from `this._localEntries` as well.

bwindels (commented):
Thank you for your changes! Had some more thoughts while going over it another time.

src/matrix/room/timeline/Timeline.js
MidhunSureshR: Previously, `updateFrom` was needed to copy over `contextFor` to the new entry. Since we now do that implicitly in `setContextEntry()`, we can remove it.

src/matrix/room/timeline/Timeline.js
MidhunSureshR: Yes, the context entry would fail to load and this would render a `ReplyPreviewError`. When we're back online, the reply preview would be rendered. We need to extend ~~getTrackedEntry~~ findLoadedEventById to also look in localEntries for the context.

MidhunSureshR: > I wonder what will happen if you reply to a pending message (e.g. while offline), and then everything gets sent when off-line. `id` returns the transaction id for `PendingEventEntry`s and the event_id for `EventEntry`s. Could it happen that we are left with both the PendingEventEntry and the EventEntry for the same message in the map, as they use a different id?

As we discussed earlier, this should probably never happen since by definition a pending message would always be in timeline and therefore the map is never involved.

src/matrix/room/timeline/entries/EventEntry.js
```diff
@@ -142,18 +128,13 @@ export class EventEntry extends BaseEventEntry {
         return getRelatedEventId(this.event);
     }
 
-    // similar to relatedEventID but only for replies
-    get contextEventId() {
-        if (this.isReply) {
-            return this.relatedEventId;
+    get threadEventId() {
```
bwindels: this does not seem used?

src/matrix/room/timeline/entries/EventEntry.js
```diff
@@ -142,18 +128,13 @@ export class EventEntry extends BaseEventEntry {
...
+        if (this.isThread) {
```
bwindels: Also, this isn't defined anywhere?

src/matrix/room/timeline/entries/EventEntry.js
```diff
@@ -45,20 +43,8 @@ export class EventEntry extends BaseEventEntry {
...
-    _setAsContextOf(entry) {
-        if (!this._contextForEntries) {
-            this._contextForEntries = [];
-        }
-        this._contextForEntries.push(entry);
-    }
-
-    get contextForEntries() {
-        return this._contextForEntries;
+    setRelatedEntry(entry) {
```
bwindels: Same, this doesn't seem to be used?

src/matrix/room/timeline/Timeline.js
```diff
@@ -751,6 +751,8 @@ export function tests() {
             }, createEvent("m.room.message", "event_id_2", bob));
             const entryB = new EventEntry({ event });
             await timeline.load(new User(alice), "join", new NullLogItem());
+            timeline._setupEntries([]);
+            timeline._localEntries.onSubscribeFirst();
```
bwindels: this seems a bit brittle ... consider subscribing to timeline.entries instead.

src/matrix/room/timeline/Timeline.js
```diff
@@ -751,6 +751,8 @@ export function tests() {
             }, createEvent("m.room.message", "event_id_2", bob));
             const entryB = new EventEntry({ event });
             await timeline.load(new User(alice), "join", new NullLogItem());
+            timeline._setupEntries([]);
```
bwindels: I'd try to avoid calling private methods from tests if possible, as they can make the tests more brittle. Why not use load?

src/matrix/room/timeline/Timeline.js
```diff
@@ -811,7 +817,12 @@ export function tests() {
             };
             const entryB = new EventEntry({ event: withContent(content, createEvent("m.room.message", "event_id_2", bob)) });
             const entryC = new EventEntry({ event: withContent(content, createEvent("m.room.message", "event_id_3", bob)) });
+            entryA._eventEntry.eventIndex = 1;
```
bwindels: why not just set it in the object literal above?

src/matrix/room/timeline/Timeline.js
```diff
@@ -141,6 +141,7 @@ export class Timeline {
             // redactingEntry might be a PendingEventEntry or an EventEntry, so don't assume pendingEvent
             const relatedTxnId = pee.redactingEntry.pendingEvent?.relatedTxnId;
             this._findAndUpdateEntryById(relatedTxnId, pee.redactingEntry.relatedEventId, updateOrFalse);
+            pee.redactingEntry.contextForEntries?.forEach(e => this._emitUpdateForEntry(e));
```
bwindels: the reason `e` is emitting an update here is because it's contextEntry might have changed. Should we add a `param` parameter to `_emitUpdateForEntry` as an argument to the update event so we can specify `"contextEntry"` here (and elsewhere where we call `_emitUpdateForEntry`)?

bwindels (commented):
some more comments, sorry :)

src/matrix/room/timeline/entries/EventEntry.js
MidhunSureshR: It's used in PR-3 (createReplyContent)

src/matrix/room/timeline/entries/EventEntry.js
MidhunSureshR: sorry about this, but it seems to be added in PR-3?

src/matrix/room/timeline/Timeline.js
MidhunSureshR: yes, we do that for updates everywhere else 👍

src/matrix/room/timeline/entries/EventEntry.js
MidhunSureshR: Sorry, moved to PR-3

src/matrix/room/timeline/entries/EventEntry.js
MidhunSureshR: moved this code to PR-3

src/matrix/room/timeline/Timeline.js
```diff
@@ -304,6 +423,19 @@ export class Timeline {
         return null;
     }
 
+    getFromAllEntriesById(eventId) {
+        if (!this._allEntries) {
+            // if allEntries isn't loaded yet, fallback to looking only in remoteEntries
+            return this.getByEventId(eventId);
+        }
+        for (const entry of this._allEntries) {
```
bwindels: Given that we won't allow replying to unsent events, should we revert this to be just looking in remoteEntries now?

src/matrix/room/timeline/Timeline.js
```diff
@@ -236,7 +256,117 @@ export class Timeline {
...
+         * Updates for entries in timeline is handled by remoteEntries observable collection
+         * Updates for entries not in timeline but fetched from storage is handled in this.replaceEntries()
+         * This code is specific to entries fetched from HomeServer i.e NonPersistedEventEntry
+         */
+        for (const entry of entries) {
+            const relatedEntry = this._contextEntriesNotInTimeline.get(entry.relatedEventId);
+            if (relatedEntry?.isNonPersisted && relatedEntry?.addLocalRelation(entry)) {
+                // update other entries for which this entry is a context entry
+                relatedEntry.contextForEntries?.forEach(e => {
+                    this._remoteEntries.findAndUpdate(te => te.id === e.id, () => "contextEntry");
```
bwindels: can we use `_emitUpdateForEntry` here?

src/matrix/room/timeline/Timeline.js
```diff
@@ -600,6 +734,177 @@ export function tests() {
...
+                body: "bar",
+                msgtype: "m.text",
+                "m.relates_to": {
+                    "m.in_reply_to": {
+                        "event_id": "event_id_1"
+                    }
+                }
+            }, createEvent("m.room.message", "event_id_2", bob));
+            const entryB = new EventEntry({ event, eventIndex: 2 });
+            timeline._getEventFromStorage = () => entryA
```
bwindels: Ideally we'd put the entry in the mock storage, rather than monkey-patch a private method. But not a blocker. See the `local reaction removal gets applied after remote echo is added to timeline with reaction not loaded` test how this can be easily done.

src/matrix/room/timeline/Timeline.js
```diff
@@ -600,6 +734,177 @@ export function tests() {
...
+                body: "bar",
+                msgtype: "m.text",
+                "m.relates_to": {
+                    "m.in_reply_to": {
+                        "event_id": "event_id_1"
+                    }
+                }
+            }, createEvent("m.room.message", "event_id_2", bob));
+            const entryB = new EventEntry({ event, eventIndex: 2 });
+            timeline._getEventFromHomeserver = () => entryA
```
bwindels: Ideally we'd pass a mock for the hsApi (with just the context method) rather than monkey-patch a private method, but not a blocker.

src/matrix/room/timeline/Timeline.js
```diff
@@ -600,6 +734,177 @@ export function tests() {
...
+            await poll(() => entryA.contextForEntries.length === 2);
+            assert.deepEqual(entryA.contextForEntries, [entryB, entryC]);
+        },
+
+        "context entry in contextEntryNotInTimeline gets updated based on incoming redaction": async assert => {
+            const timeline = new Timeline({roomId, storage: await createMockStorage(), closeCallback: () => {},
+                fragmentIdComparer, pendingEvents: new ObservableArray(), clock: new MockClock()});
+            const entryA = new NonPersistedEventEntry({ event: withTextBody("foo", createEvent("m.room.message", "event_id_1", alice)), eventIndex: 1 });
+            let event = withContent({
+                body: "bar",
```
bwindels: feel free to put the  "m.relates_to" in a `withReply` method with the other event mock functions, up to you, not blocking.

src/matrix/room/timeline/Timeline.js
```diff
@@ -600,6 +734,177 @@ export function tests() {
...
+            }, createEvent("m.room.message", "event_id_2", bob));
+            const entryB = new EventEntry({ event, eventIndex: 2 });
+            timeline._getEventFromHomeserver = () => entryA;
+            await timeline.load(new User(alice), "join", new NullLogItem());
+            timeline.entries.subscribe({ onAdd: () => null, onUpdate: () => null });
+            timeline.addEntries([entryB]);
+            await poll(() => entryB.contextEntry);
+            const redactingEntry = new EventEntry({ event: withRedacts(entryA.id, "foo", createEvent("m.room.redaction", "event_id_3", alice)), eventIndex: 3 });
+            timeline.addEntries([redactingEntry]);
+            const contextEntry = timeline._contextEntriesNotInTimeline.get(entryA.id);
```
bwindels: could we do `entryB.contextEntry.isRedacted` here to avoid accessing a private member, or you want to test if it's in entries without context as well?

src/matrix/room/timeline/Timeline.js
```diff
@@ -600,6 +734,177 @@ export function tests() {
...
+                    }
+                }
+            };
+            const entryB = new EventEntry({ event: withContent(content, createEvent("m.room.message", "event_id_2", bob)), eventIndex: 2 });
+            const entryC = new EventEntry({ event: withContent(content, createEvent("m.room.message", "event_id_3", bob)), eventIndex: 3 });
+            await timeline.load(new User(alice), "join", new NullLogItem());
+            const bin = [entryB, entryC];
+            timeline.entries.subscribe({
+                onUpdate: (index) => {
+                    const i = bin.findIndex(e => e.id === index);
```
bwindels: I'm not sure why we're comparing the entry id to the index here? Also, wouldn't it just be simpler to log the indices of the updates here and later assert that to `[1, 2]` or whatever the right value is?

src/matrix/room/timeline/Timeline.js
```diff
@@ -600,6 +734,177 @@ export function tests() {
...
+            await timeline.load(new User(alice), "join", new NullLogItem());
+            timeline.entries.subscribe({ onAdd: () => null, onUpdate: () => null });
+            timeline.addEntries([entryB]);
+            await poll(() => entryB.contextEntry);
+            assert.strictEqual(timeline._contextEntriesNotInTimeline.has(entryA.id), true);
+            timeline.addEntries([entryA]);
+            assert.strictEqual(timeline._contextEntriesNotInTimeline.has(entryA.id), false);
+            const movedEntry = timeline.remoteEntries[0];
+            assert.deepEqual(movedEntry, entryA);
+            assert.deepEqual(movedEntry.contextForEntries.pop(), entryB);
```
bwindels: Fine either way, but curious why you pop rather than `[0]`?

bwindels (approved):
Here's my final review, looking great! Just a few more small things, some of them up to you and not blocking. Thanks again for persisting at this!

src/matrix/room/timeline/entries/NonPersistedEventEntry.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    // overridden here because we reuse addLocalRelation() for updating this entry
+    // we don't want the RedactedTile created using this entry to ever show "is being redacted"
+    get isRedacting() {
+        return false;
+    }
+
+    get isRedacted() {
+        return !!this._pendingRedactions;
```
bwindels:
```suggestion
        return super.isRedacting;
```

bwindels (commented):
small suggestion

src/matrix/room/timeline/Timeline.js
MidhunSureshR: Sorry, the logic for this test is borked. I've improved it somewhat. 
The problem with asserting indices is that update events may be emitted twice so the indices might be something like [0, 1, 0, 1].

bwindels (approved):
Thanks for the last changes, this is good to go!

PR #612, opened 2021-12-07
Threading fallback - PR 2 - Support rich reply

MidhunSureshR:
- [x] If the related tile cannot be fetched, render an error where the reply should have been rendered

## Overview
- Adds a reply tile property to BaseMessageTile that gives a vm based on the context-entry
- This reply tile is updated based on updates from TilesCollection
- Implements `ReplyPreviewView` to render the reply tile
- Removes mx-reply tags


src/domain/session/room/timeline/deserialize.js
```diff
@@ -346,6 +346,10 @@ class Deserializer {
 }
 
 export function parseHTMLBody(platform, mediaRepository, allowReplies, html) {
+    if (allowReplies) {
```
bwindels: I think we might want to remove the flag entirely? If we want to still be able to fallback to the embedded replied-to text if we can't fetch the context event, I would add another function to give us the `MessageBody` for the embedded reply and parse **only** what is in the mx-reply tag.

src/domain/session/room/timeline/tiles/BaseTextTile.js
```diff
@@ -56,4 +57,43 @@ export class BaseTextTile extends BaseMessageTile {
...
+        if (!this._entry.contextEventId) {
+            return null;
+        }
+        const entry = this._entry.contextEntry;
+        const error = this._generateError(entry);
+        if (error?.name === "ContextEntryNotFound") {
+            return { error };
+        }
+        const format = entry.content.format === "org.matrix.custom.html" ? BodyFormat.Html : BodyFormat.Plain;
+        const body = entry.content["formatted_body"] ?? entry.content["body"];
```
bwindels: if format is plain, we should never look at formatted_body.
Perhaps we can extract two functions to get the format from an event and the body from an event given the format and use that here and in TextTile where it is already implemented?

src/domain/session/room/timeline/tiles/BaseTextTile.js
```diff
@@ -56,4 +57,43 @@ export class BaseTextTile extends BaseMessageTile {
...
+        return {
```
bwindels: I'd create a `ReplyPreviewViewModel` (extending from ViewModel) class for this taking the entry and perhaps moving all the parsing logic in there, then you can also use getters for the avatar properties. Mentioned this before I think, but this will need to be cached like we already do for the main message body otherwise the binding would fire every time it is evaluated.

src/domain/session/room/timeline/tiles/BaseTextTile.js
```diff
@@ -56,4 +57,43 @@ export class BaseTextTile extends BaseMessageTile {
...
+                avatarUrl: (size) => getAvatarHttpUrl(entry.avatarUrl, size, this.platform, this._mediaRepository)
+            },
+            error
+        };
+    }
+
+    _generateError(entry) {
+        const createError = (name) => { const e = new Error(); e.name = name; return e;}
+        if (!entry) {
+            return createError("ContextEntryNotFound");
```
bwindels: does this mean that we create an error for any message that does not reply to another message?

src/domain/session/room/timeline/tiles/BaseTextTile.js
```diff
@@ -56,4 +57,43 @@ export class BaseTextTile extends BaseMessageTile {
...
+            avatar: {
+                avatarColorNumber: getIdentifierColorNumber(entry.sender),
+                avatarLetter: avatarInitials(entry.displayName),
+                avatarUrl: (size) => getAvatarHttpUrl(entry.avatarUrl, size, this.platform, this._mediaRepository)
+            },
+            error
+        };
+    }
+
+    _generateError(entry) {
```
bwindels: I'd perhaps be more inclined to work with flags on the proposed ReplyPreviewViewModel class rather than using Error objects for this.

bwindels: The not found case we can handle if the reply entry has `isReply === true` but there is no contextEvent.

bwindels (commented):
had a first look, some comments but looking good so far

src/domain/session/room/timeline/tiles/BaseTextTile.js
MidhunSureshR: No, but the code has changed anyway.

src/domain/session/room/timeline/tiles/BaseTextTile.js
MidhunSureshR: Marking as resolved since we're just using `TextTile` as the vm.

src/domain/session/room/timeline/deserialize.js
MidhunSureshR: I've removed this flag but currently we don't rely on the fallback at all. If we can't fetch the context event, we render an error instead.

src/platform/web/ui/session/room/timeline/TextMessageView.js
```diff
@@ -24,19 +25,37 @@ export class TextMessageView extends BaseMessageView {
...
             for (const part of body.parts) {
                 container.appendChild(renderPart(part));
             }
             container.appendChild(time);
         });
+
         return container;
     }
+
+    _shouldRemove(element) {
```
bwindels: I would put this method as a local function variable just above the mapSideEffect where it is used, as it's a bit out of context here, especially with a name that doesn't give a lot of context.

src/matrix/room/timeline/Timeline.js
```diff
@@ -221,7 +221,7 @@ export class Timeline {
     }
 
     /** @package */
-    replaceEntries(entries) {
+     replaceEntries(entries) {
```
bwindels: any reason for this space?

src/domain/session/room/timeline/deserialize.js
```diff
@@ -345,9 +344,10 @@ class Deserializer {
     }
 }
 
-export function parseHTMLBody(platform, mediaRepository, allowReplies, html) {
+export function parseHTMLBody(platform, mediaRepository, html) {
+    html = html.replace(/<mx-reply>.+<\/mx-reply>/s, "");
```
bwindels: Not sure about adding yet another way to filter tags here. I'd remove `ADD_TAGS` in `parsehtml.js`, so the sanitizer will already filter it, and the change above in `_isAllowedNode` should make sure the parser doesn't decend into the `mx-reply` tag IIRC?

src/platform/web/ui/session/room/timeline/ReplyPreviewView.js
```diff
@@ -0,0 +1,51 @@
...
+        const viewClass = viewClassForEntry(vm);
+        const view = new viewClass(vm, { reply: true, interactive: false });
+        const rendered = view.render(t, vm);
+        return this._renderReplyHeader(t, vm, [rendered]);
+    }
+
+    _renderReplyHeader(t, vm, children = []) {
+        return t.blockquote(
+            [
+            t.a({ className: "link", href: `https://matrix.to/#/${vm.roomId}/${vm.eventId}` }, "In reply to"),
```
bwindels: I'd rather expose a `permaLink` property from `BaseMessageTile` I think, rather than these two rather "low-level"  properties.

src/platform/web/ui/session/room/timeline/ReplyPreviewView.js
```diff
@@ -0,0 +1,51 @@
...
+            t.a({ className: "pill", href: `https://matrix.to/#/${vm.sender}` }, [renderStaticAvatar(vm, 12, undefined, true), vm.displayName]),
```
bwindels: could do the same with this one, `senderProfileLink`perhaps?

src/platform/web/ui/session/room/timeline/TextMessageView.js
```diff
@@ -24,19 +25,37 @@ export class TextMessageView extends BaseMessageView {
...
+        return element && element.className !== "ReplyPreviewView" && element.nodeName !== "#comment";
```
bwindels: I'd rather exclude any node type not an element
```suggestion
        return element?.nodeType === Node.ELEMENT_NODE && element.className !== "ReplyPreviewView";
```

src/platform/web/ui/session/room/timeline/ReplyPreviewView.js
```diff
@@ -0,0 +1,51 @@
...
+export class ReplyPreviewView extends TemplateView {
+    render(t, vm) {
+        return t.div({ className: "ReplyPreviewView" }, this._renderReplyPreview(t, vm));
+    }
+
+    _renderReplyPreview(t, vm) {
+        // todo: this should probably be called viewClassForTile instead
+        const viewClass = viewClassForEntry(vm);
+        const view = new viewClass(vm, { reply: true, interactive: false });
+        const rendered = view.render(t, vm);
```
bwindels: Calling `render` (or any method really) on a view that is not mounted is a no-no. Also `render` is not part of the `IView` contract but a `TemplateView` impl detail, which `view` may or may not be a instance of. Either way, you can use `t.view` for mounting and rendering a view unconditionally:
```suggestion
        const rendered = t.view(view);
```

src/platform/web/ui/session/room/timeline/ReplyPreviewView.js
```diff
@@ -0,0 +1,51 @@
...
+
+export class ReplyPreviewView extends TemplateView {
+    render(t, vm) {
+        return t.div({ className: "ReplyPreviewView" }, this._renderReplyPreview(t, vm));
+    }
+
+    _renderReplyPreview(t, vm) {
+        // todo: this should probably be called viewClassForTile instead
+        const viewClass = viewClassForEntry(vm);
+        const view = new viewClass(vm, { reply: true, interactive: false });
```
bwindels, line 27: should we accomodate here for the case where viewClass is undefined?

src/platform/web/ui/session/room/timeline/ReplyPreviewView.js
```diff
@@ -0,0 +1,51 @@
...
+import {renderStaticAvatar} from "../../../avatar";
+import {TemplateView} from "../../../general/TemplateView";
+import {viewClassForEntry} from "../common";
+
+export class ReplyPreviewView extends TemplateView {
+    render(t, vm) {
+        return t.div({ className: "ReplyPreviewView" }, this._renderReplyPreview(t, vm));
+    }
+
+    _renderReplyPreview(t, vm) {
```
bwindels: i'd personally inline this in render, given both methods together are so small

src/platform/web/ui/session/room/timeline/ReplyPreviewView.js
```diff
@@ -0,0 +1,51 @@
...
+        // todo: this should probably be called viewClassForTile instead
+        const viewClass = viewClassForEntry(vm);
+        const view = new viewClass(vm, { reply: true, interactive: false });
+        const rendered = view.render(t, vm);
+        return this._renderReplyHeader(t, vm, [rendered]);
+    }
+
+    _renderReplyHeader(t, vm, children = []) {
```
bwindels: and perhaps even inline this if you want

src/platform/web/ui/session/room/timeline/BaseMessageView.js
```diff
@@ -24,12 +24,13 @@ import {Menu} from "../../../general/Menu.js";
 import {ReactionsView} from "./ReactionsView.js";
 
 export class BaseMessageView extends TemplateView {
-    constructor(value, interactive = true, tagName = "li") {
+    constructor(value, renderFlags, tagName = "li") {
```
bwindels, line 27: when converting to typescript we could use a bitflag enum here, but this is fine for now.

src/domain/session/room/timeline/tiles/BaseMessageTile.js
```diff
@@ -202,4 +215,17 @@ export class BaseMessageTile extends SimpleTile {
             this._reactions.update(annotations, pendingAnnotations);
         }
     }
+
+    get replyTextTile() {
```
bwindels: why text tile rather than tile? is anything limiting this to a `TextTile`?

src/domain/session/room/timeline/tiles/SimpleTile.js
```diff
@@ -21,6 +21,7 @@ import {SendStatus} from "../../../../../matrix/room/sending/PendingEvent.js";
 export class SimpleTile extends ViewModel {
     constructor(options) {
         super(options);
+        this._tileCreator = options.tileCreator;
```
bwindels: Didn't we discuss passing `_tileCreator` as a parameter to `updateEntry`? If that doesn't work, could you remind me what the problem was?

bwindels (commented):
added some comments

src/matrix/room/timeline/Timeline.js
MidhunSureshR: sorry

src/domain/session/room/timeline/deserialize.js
MidhunSureshR: I had to add this to sanitizeConfig in addition to removing `ADD_TAGS`:
```js
    FORBID_TAGS: ['mx-reply'],
    KEEP_CONTENT: false,
```

src/platform/web/ui/session/room/timeline/ReplyPreviewView.js
MidhunSureshR: done but in practice hydrogen would swallow unknown event-types so it would render a `RepyPreviewError`

src/domain/session/room/timeline/tiles/BaseMessageTile.js
MidhunSureshR: sorry, was only considering text-tiles initially

src/domain/session/room/timeline/tiles/SimpleTile.js
MidhunSureshR: I also use `tileCreator` in` BaseMessageTile.replyTile`

src/domain/session/room/timeline/tiles/SimpleTile.js
```diff
@@ -93,6 +94,17 @@ export class SimpleTile extends ViewModel {
...
+        if (replyEntry) {
+            // this is an update to contextEntry used for replyPreview
+            const action = this._replyTile?.updateEntry(replyEntry);
+            if (action?.shouldReplace) {
+                this._replyTile = this._tileCreator(replyEntry);
+            }
+            else {
+                this._replyTile?.emitChange();
+            }
+        }
```
MidhunSureshR: This logic should probably be moved to `BaseMessageTile.updateEntry()`. For one, `SimpleTile` does not know what a replyTile is (it's defined in `BaseMessageTile`). Also not all SimpleTiles are message tiles.

src/domain/session/room/timeline/tiles/SimpleTile.js
MidhunSureshR: This can also be moved to `BaseMessageTile`.

MidhunSureshR: should also be `tilesCreator` since that's what is used elsewhere

src/platform/web/ui/session/room/timeline/BaseMessageView.js
```diff
@@ -24,12 +24,13 @@ import {Menu} from "../../../general/Menu.js";
...
         super(value);
         this._menuPopup = null;
         this._tagName = tagName;
         // TODO An enum could be nice to make code easier to read at call sites.
-        this._interactive = interactive;
+        this._interactive = renderFlags?.interactive ?? true;
```
bwindels: why copy the individual flags rather than just store renderFlags on this?

src/domain/session/room/timeline/tiles/BaseMessageTile.js
```diff
@@ -21,9 +21,11 @@ import {getIdentifierColorNumber, avatarInitials, getAvatarHttpUrl} from "../../
 export class BaseMessageTile extends SimpleTile {
     constructor(options) {
         super(options);
+        this._tilesCreator = options.tilesCreator;
```
bwindels: About this, I'd prefer to not add another member to this class if possible. Can we have a method where we create the replyTile that is called from the ctor and updateEntries when needed and receiving the tilesCreator as an argument?

bwindels (approved):
Looking good, just some tiny nits left. Thanks!

src/domain/session/room/timeline/tiles/BaseMessageTile.js
MidhunSureshR: Turns out, we don't really need to create the reply tile initially. We can just create it on updateEntry since we'll be notified when context-entry is loaded.

PR #616, opened 2021-12-09
Fixed autocomplete bug by enclosing div within a form tag #380

### LINKED ISSUE #380: Chrome Autofills Security Key
**Describe the bug**
Good Chrome thinks the security key field in Settings is a password field and will autofill

**To Reproduce**
1. Open settings in hydrogen on Google Chrome with a save credentialf ro the site
2. Chrome will autofile the secutiy key and the search bar with your username and password

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Windows and Linux
 - Browser Chrome

aaryan-gautam:
> Description

For issue #380, I have inserted a `<form>` that is wrapped around the `<div>` as it is needed for the autocomplete as given at [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) so the `utilitiesRow` variable now returns this `<form>`.
I have assigned a name attribute of `utilities-form` to the `<form>`.
> Before:

image

> After:

image

> Testing Change

1. Save a login credential on the website on google chrome
2. Open settings to check whether the username is autofilled on click in room filter search and password for the security field.

Resolves #380 

Signed-off-by: Aaryan Gautam <aaryanyo123@gmail.com>


src/platform/web/ui/session/leftpanel/LeftPanelView.js
```diff
@@ -71,7 +68,8 @@ export class LeftPanelView extends TemplateView {
                 }
             }
         ));
-        const utilitiesRow = t.div({className: "utilities"}, [
+        const utilitiesRow = t.form({name:"utilities-form"}, [
```
bwindels, line 71: won't this submit the form now when you hit enter on the filter field?

src/platform/web/ui/session/leftpanel/LeftPanelView.js
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2020 Bruno Windels <bruno@windels.cloud>
-
```
bwindels, line 3: Please don't introduce any unrelated white space changes

src/platform/web/ui/session/leftpanel/LeftPanelView.js
bwindels, line 71: probably only the filter field should be in the form, not the button links?

bwindels (commented):
Thanks for your PR, I have left some comments

src/platform/web/ui/session/leftpanel/LeftPanelView.js
pruthula0: `const utilitiesRow = t.form({
  name: "utilities-form",
  onSubmit: (event) => {
    event.preventDefault(); // Prevent form submission
    // Additional code logic if needed
  },
}, [
  // Form elements
]);
`


> By adding the onSubmit event listener to the form element and calling event.preventDefault(), you can prevent the form from submitting when the enter key is pressed within the filter field or any other input field inside the form.

```

PR #619, opened 2021-12-14
Threading fallback - PR 3 - Send replies to threaded message as thread

MidhunSureshR:
This PR treats threads as replies. Replies to threaded messages are sent as both reply and as a threaded message.


src/matrix/room/timeline/entries/reply.js
```diff
@@ -72,3 +75,14 @@ export function createReplyContent(entry, msgtype, body) {
...
     return _createReplyContent(entry.id, msgtype, newBody, newFormattedBody);
 }
+
+function createThreadContent(entry, msgtype, body) {
+    return {
+        msgtype,
+        body,
+        "m.relates_to": {
+            "rel_type": "m.thread",
+            "event_id": entry.id
```
DanilaFe: Should this perhaps be `entry.content["m.relates_to"].event_id`? [MSC 3440](https://github.com/matrix-org/matrix-doc/blob/gsouquet/threading-via-relations/proposals/3440-threading-via-relations.md) specifies:
```
"m.relates_to": {
  "rel_type": "m.thread",
  "event_id": "$thread_root"
}
```
But `entry.id` is the ID of the current event being replied to.

src/matrix/room/timeline/entries/reply.js
```diff
@@ -51,6 +51,9 @@ function _createReplyContent(targetId, msgtype, body, formattedBody) {
 }
 
 export function createReplyContent(entry, msgtype, body) {
+    if (entry.isThread) {
+        return createThreadContent(entry, msgtype, body);
```
DanilaFe: Perhaps it would be good to _also_ include the reply content here? [This is how MSC 3440](https://github.com/matrix-org/matrix-doc/blob/gsouquet/threading-via-relations/proposals/3440-threading-via-relations.md#quote-replies-in-a-thread) seem to do it.

src/matrix/room/timeline/entries/reply.js
```diff
@@ -72,3 +75,14 @@ export function createReplyContent(entry, msgtype, body) {
     const newFormattedBody = formattedFallback + htmlEscape(body);
     return _createReplyContent(entry.id, msgtype, newBody, newFormattedBody);
 }
+
+function createThreadContent(entry, msgtype, body) {
+    return {
+        msgtype,
+        body,
+        "m.relates_to": {
+            "rel_type": "m.thread",
```
DanilaFe: In `isThread`, we use `io.element.thread`, but here we use `m.thread`. Should this be just one thing? Would it be good to extract it into a constant so that one can be switch into the other once (if) the threads MSC lands?

src/matrix/room/timeline/entries/reply.js
MidhunSureshR: I'm not sure if we always need to do that(?). Replies to replies in a thread should probably also be replies. But otherwise we just send a thread message.

MidhunSureshR: hmm.. not sure about this, maybe we should always include reply content :thinking:

MidhunSureshR: Yeah you're right, we should always include the reply content as well 👍

bwindels (approved):
Looking good!

PR #623, opened 2021-12-23
Bootstrap enough registration functionality for embedded-hydrogen work

### LINKED ISSUE #538: SDK support for account registration
https://spec.matrix.org/v1.1/client-server-api/#post_matrixclientv3register
which follows the interactive UIA api, where you submit {auth: {...}} to the same endpoint until you don't get 401 anymore.
https://spec.matrix.org/v1.1/client-server-api/#user-interactive-authentication-api

we'll need support for m.login.terms, which is sort of spec'ed in https://github.com/matrix-org/matrix-doc/pull/1692, it's not clear from the msc but it looks like it doesn't need any auth data when sending it back, judging from the [element impl](https://github.com/matrix-org/matrix-react-sdk/pull/2221/files#diff-80decdd03362142bcf54284bea43e0fadeb8605bc18706cf582c60467f87e7b0R290).

I think it might be best to always use inhibit_login and to not login through the /register endpoint. This way, we have only one path to login, which with things like dehydrated devices, etc might be a bit easier.

I would add a `async queryRegistration` method to <del>`SessionContainer`</del> `Client`, analog to `queryLogin`, which gets the flows and returns a Registration class that allows you to set the user/pass/email and access to the first stage of the chosen flow (for now I would perhaps just pick the first flow of which we support all the login stages?) with a wrapper class for each stage, with a submit method, returning a promise with the object for the next stage, if any? just an idea, haven't completely thought this through yet what would be a nice API.

MidhunSureshR:


src/matrix/net/HomeServerRequest.ts
```diff
@@ -30,13 +31,18 @@ export class HomeServerRequest implements IHomeServerRequest {
     // as we add types for expected responses from hs, this could be a generic class instead
     private readonly _promise: Promise<any>;
 
-    constructor(method: string, url: string, sourceRequest: RequestResult, log?: ILogItem) {
+    constructor(method: string, url: string, sourceRequest: RequestResult, options?: IRequestOptions) {
```
MidhunSureshR: `IRequestOptions` are the options for the `HomeServerAPI._requestFn`, should we be using it here?

src/matrix/net/HomeServerRequest.ts
```diff
@@ -30,13 +31,18 @@ export class HomeServerRequest implements IHomeServerRequest {
...
+            const parent = options?.log;
+            log = parent.child({ t: "network", url, method, }, parent.level.Info);
+        }
         this._log = log;
         this._sourceRequest = sourceRequest;
         this._promise = sourceRequest.response().then(response => {
             log?.set("status", response.status);
             // ok?
-            if (response.status >= 200 && response.status < 300) {
+            if (response.status >= 200 && response.status < 300 || options?.allowedErrors?.find(e => e === response.status)) {
```
bwindels:
```suggestion
            if (response.status >= 200 && response.status < 300 || options?.allowedErrors?.includes(response.status)) {
```

src/matrix/registration/Registration.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export class Registration {
+    private _hsApi: HomeServerApi;
+    private _accountDetails: AccountDetails;
+    private _flowSelector: FlowSelector;
+
+    constructor(hsApi: HomeServerApi, accountDetails: AccountDetails, flowSelector?: FlowSelector) {
+        this._hsApi = hsApi;
+        this._accountDetails = accountDetails;
+        this._flowSelector = flowSelector ?? (flows => flows.pop());
```
bwindels: best not to modify the argument? can be confusing during debugging.
```suggestion
        this._flowSelector = flowSelector ?? (flows => flows[flows.length - 1]);
```

src/matrix/registration/types/types.ts
```diff
@@ -0,0 +1,58 @@
+/*
```
bwindels: why not put this file directly in `registration`?

src/matrix/net/HomeServerApi.ts
```diff
@@ -163,6 +156,21 @@ export class HomeServerApi {
         return this._unauthedRequest("GET", this._url("/login"));
     }
 
+    register(username: string | null, password: string, initialDeviceDisplayName: string, auth?: Record<string, any>, inhibitLogin: boolean = true , options: IRequestOptions = {}): IHomeServerRequest {
+        Object.assign(options, { allowedErrors: [401] });
```
bwindels:
```suggestion
        options.allowedErrors = [401];
```

src/matrix/registration/stages/BaseRegistrationStage.ts
```diff
@@ -0,0 +1,61 @@
...
+     * - user-id (string) if registration is completed
+     */
+    abstract complete(auth?: AuthenticationData): Promise<BaseRegistrationStage | string>;
+
+    setNextStage(stage: BaseRegistrationStage) {
+        this._nextStage = stage;
+    }
+
+    parseResponse(response: RegistrationResponse) {
+        if ("user_id" in response) {
```
bwindels: probably better to check the status code being 200 here rather than look for properties that happen to only be there when we're done?

src/matrix/registration/stages/BaseRegistrationStage.ts
```diff
@@ -0,0 +1,61 @@
...
+import type {AccountDetails, RegistrationResponse, AuthenticationData, RegistrationParams} from "../types/types";
+
+export abstract class BaseRegistrationStage {
+    protected _hsApi: HomeServerApi;
+    protected _accountDetails: AccountDetails;
+    protected _session: string;
+    protected _nextStage: BaseRegistrationStage;
+    protected _params?: Record<string, any>
+
+    constructor(hsApi: HomeServerApi, accountDetails: AccountDetails, session: string, params?: RegistrationParams) {
```
bwindels: I would personally move doing the actual /register request to the `Registration` class, and make the `*Stage` classes only responsible for creating the next `AuthenticationData` to send. This way, `Registration` would have a `submitStage` method that does the register request with the auth data of the stage?

src/matrix/registration/stages/BaseRegistrationStage.ts
```diff
@@ -0,0 +1,61 @@
...
+     * eg: m.login.recaptcha or m.login.dummy
+     */
+    abstract get type(): string;
+
+    /**
+     * Finish a registration stage, return value is:
+     * - the next stage if this stage was completed successfully
+     * - user-id (string) if registration is completed
+     */
+    abstract complete(auth?: AuthenticationData): Promise<BaseRegistrationStage | string>;
```
bwindels: why not store the access token, user name, ... on the Registration once were done and just return undefined to indicate that we're done? `BaseRegistrationStage | string` feels like it's a bit of a confusing API.

Taking the `submitStage` suggestion from below into account, this would actually become the return type of that method. I imagine you'd write something like this to consume this API:

```js
const registration = ...;
let stage = await registration.start();
while (stage) {
  switch (stage.type) {
   case ...:
       stage.setSomething(...);
       break;
 }
  stage = await registration.submitStage(stage);
}
return registration.sessionInfo; // contains homeServer, accessToken, userId, deviceId, ...
```

src/matrix/net/HomeServerRequest.ts
bwindels: No, the options passed to the different methods on HomeServerApi is actually a different object, the typing seems wrong. Well spotted. This object has log, timeout, uploadProgress and prefix (which we should not put on there really, but should just be an argument of _get, _post, ...). At the same time, `IRequestOptions` does not have log or prefix. Can you add a type and fix it? :pray:

src/matrix/registration/Registration.ts
```diff
@@ -0,0 +1,69 @@
...
+        const response = await this._hsApi.register(
+            this._accountDetails.username,
+            this._accountDetails.password,
+            this._accountDetails.initialDeviceDisplayName,
+            undefined,
+            this._accountDetails.inhibitLogin).response();
+        return this.parseStagesFromResponse(response);
+    }
+
+    parseStagesFromResponse(response: RegistrationResponseMoreDataNeeded): BaseRegistrationStage {
```
bwindels:
```suggestion
    private parseStagesFromResponse(response: RegistrationResponseMoreDataNeeded): BaseRegistrationStage {
```

bwindels: After refactoring, can you also make other remaining methods that don't need to be public private?
Non-public methods are methods SDK users can't use and hence complain about when we change them, so less maintenance burden.

bwindels (commented):
great start, some suggested changes to simplify things.

src/matrix/registration/stages/BaseRegistrationStage.ts
MidhunSureshR: RegistrationResponse is a union of different types so we need some way of [narrowing the type](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing) (otherwise typescript complains). We could check the status code and then narrow the type explicitly but where do we get the status code?
We could add a `statusCode` method to HomeServerRequest.

src/matrix/net/RequestScheduler.ts
```diff
@@ -50,6 +50,13 @@ class Request implements IHomeServerRequest {
     response(): Promise<any> {
         return this._responsePromise;
     }
+
+    async responseCode(): Promise<number> {
+        let resolve;
+        const promise: Promise<number> = new Promise(r => resolve = r);
+        this.requestResult?.responseCode().then(code => resolve(code));
+        return promise;
+    }
```
MidhunSureshR: I'm not sure if this is what I'm supposed to do.. wdyt?

src/matrix/net/HomeServerApi.ts
```diff
@@ -133,37 +134,52 @@ export class HomeServerApi {
...
         return this._get(`/rooms/${encodeURIComponent(roomId)}/state/${encodeURIComponent(eventType)}/${encodeURIComponent(stateKey)}`, {}, undefined, options);
     }
 
     getLoginFlows(): IHomeServerRequest {
         return this._unauthedRequest("GET", this._url("/login"));
     }
 
-    passwordLogin(username: string, password: string, initialDeviceDisplayName: string, options?: IRequestOptions): IHomeServerRequest {
+    register(username: string | null, password: string, initialDeviceDisplayName: string, auth?: Record<string, any>, inhibitLogin: boolean = true , options: BaseRequestOptions = {}): IHomeServerRequest {
+        options.allowedErrors = [401];
```
bwindels: I'd put this at the call site in Registration I think, as that's where we'll actually handle the 401.

src/matrix/registration/types.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export type RegistrationResponseMoreDataNeeded = {
+    completed?: string[];
+    flows: RegistrationFlow[];
+    params: Record<string, any>;
+    session: string;
+    status: 401;
+} 
+
+type RegistrationResponseError = {
```
bwindels: I don't think you'll ever receive a response like this from HomeServerApi as the only allowedError is currently 401.
Probably best to use the existing HomeServerError for this though, as both the errcode and error are available on that?

src/matrix/registration/stages/BaseRegistrationStage.ts
```diff
@@ -0,0 +1,52 @@
...
+import type {AccountDetails, AuthenticationData, RegistrationParams} from "../types";
+
+export abstract class BaseRegistrationStage {
+    protected readonly _hsApi: HomeServerApi;
+    protected readonly _accountDetails: AccountDetails;
+    protected _session: string;
+    protected _nextStage: BaseRegistrationStage;
+    protected readonly _params?: Record<string, any>
+
+    constructor(hsApi: HomeServerApi, accountDetails: AccountDetails, session: string, params?: RegistrationParams) {
```
bwindels: I think we don't need hsApi or accountDetails here anymore, do we?

src/matrix/registration/registrationStageFromType.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import type {BaseRegistrationStage} from "./stages/BaseRegistrationStage";
+import type {HomeServerApi} from "../net/HomeServerApi";
+import type {AccountDetails, RegistrationParams} from "./types";
+import {DummyAuth} from "./stages/DummyAuth";
+import {TermsAuth} from "./stages/TermsAuth";
+
+type ClassDerivedFromBaseRegistration = { new(hsApi: HomeServerApi, registrationData: AccountDetails, session: string, params?: RegistrationParams): BaseRegistrationStage } & typeof BaseRegistrationStage;
+
+export function registrationStageFromType(type: string): ClassDerivedFromBaseRegistration | undefined{
```
bwindels: A nit, but I would just put this in registration and return an instance immediately rather than the constructor?

src/matrix/registration/Registration.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private parseStagesFromResponse(response: RegistrationResponseMoreDataNeeded): BaseRegistrationStage {
+        const { session, params } = response;
+        const flow = this._flowSelector(response.flows);
+        if (!flow) {
+            throw new Error("flowSelector did not return any flow!");
+        }
+        let firstStage: BaseRegistrationStage | undefined;
+        let lastStage: BaseRegistrationStage;
```
bwindels:
```suggestion
        let lastStage: BaseRegistrationStage | undefined;
```

src/matrix/net/RequestScheduler.ts
bwindels: Let's discuss OOB

bwindels (commented):
Thanks for the changes, some final thoughts but looking good.

src/matrix/registration/stages/BaseRegistrationStage.ts
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * eg: m.login.recaptcha or m.login.dummy
+     */
+    abstract get type(): string;
+
+    /**
+     * This method should return auth part that must be provided to
+     * /register endpoint to successfully complete this stage
+     */
+    abstract generateAuthenticationData(): AuthenticationData;
```
bwindels, line 39: Forgot before, but we should mark methods that are not meant to be used outside of the SDK as `@internal`, we can then prevent these from being put in the d.ts file with the [`stripInternal`](https://www.typescriptlang.org/tsconfig#stripInternal) option. 
```suggestion
    /** @internal */
    abstract generateAuthenticationData(): AuthenticationData;
```

src/matrix/net/RequestScheduler.ts
```diff
@@ -25,31 +25,58 @@ import type {IHomeServerRequest} from "./HomeServerRequest.js";
...
 
     abort(): void {
-        if (this.requestResult) {
-            this.requestResult.abort();
+        if (this._requestResult) {
+            this._requestResult.abort();
         } else {
-            this.reject(new AbortError());
+            this.responseReject(new AbortError());
+            this.responseCodeReject(new AbortError());
```
bwindels: this might not be set yet, I think?

src/matrix/net/RequestScheduler.ts
```diff
@@ -113,9 +140,7 @@ export class RequestScheduler {
                         request.methodName
                     ].apply(this._hsApi, request.args);
                     // so the request can be aborted
-                    request.requestResult = requestResult;
-                    const response = await requestResult.response();
-                    request.resolve(response);
+                    request.setRequestResult(requestResult);
```
bwindels: the await was important here as it prevents tight-looping and allows to catch any errors.

bwindels (approved):
Thanks for the changes, looks good!

PR #652, opened 2022-01-30
Added: Dependabot & Dockerfile image tags

GoliathLabs:
I already use Dependabot for many projects. With Dependabot it is relatively easy to stay up to date with the dependencies and packages used. Instead of always checking for updates manually, you can use the dependabot.yml configuration to tell Dependabot to check the project once a week for new releases of e.g. the npm dependencies that are used or newer Docker image tags. If there are new updates, a new pull request is automatically created for them.

Since this procedure takes at least some work off my hands over at my projects, I wanted to ask if you would be up to using it as well.

I've also added versioned image tags to both Dockerfiles so we have more control over which tag we use in case a new version of Node breaks something on our end in the future

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+# Docs: <https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/customizing-dependency-updates>
+
+version: 2
+
+updates:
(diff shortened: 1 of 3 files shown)


GoliathLabs:
@bwindels what do you think? Can we merge this or is there something I should change or still needs adjustments?

PR #655, opened 2022-02-02
Docker image improvements

sandhose:
The commits are individually reviewable. This includes:

 - the docker image is now multi-arch capable thanks to BuildKit. It does not rely on QEMU emulation, rather using BuildKit's `TARGETPLATFORM` and `BUILDPLATFORM` build args for doing the bundling under the host arch
 - use an unprivileged nginx base image to run as non-root. This also allows running the whole thing with a read-only root filesystem. It still needs a writable `/tmp`, e.g. `docker run --publish 80:8080 --read-only --volume tmp:/tmp hydrogen`
 - have the CI build a multi-arch image (currently building for x86_64, ARM and ARM64)
 - some updates to the documentation
 - a `config.json` templated from environment variables

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -23,6 +23,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
       - name: Log into registry ${{ env.REGISTRY }}
         uses: docker/login-action@v1
         with:
@@ -39,6 +42,7 @@ jobs:
       - name: Build and push Docker image
         uses: docker/build-push-action@v2
         with:
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,5 +10,21 @@ RUN yarn install
 COPY . .
 RUN yarn build
 
+# Remove the default config, replace it with a symlink to somewhere that will be updated at runtime
+RUN rm -f target/config.json \
+    && ln -sf /tmp/config.json target/config.json
+
 FROM --platform=${TARGETPLATFORM} docker.io/nginxinc/nginx-unprivileged:1.21-alpine
+
+# Copy the config template as well as the templating script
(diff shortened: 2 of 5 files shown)


bwindels:
As [mentioned](https://github.com/vector-im/hydrogen-web/pull/660#discussion_r804578721) elsewhere, I don't think we want to load the config file like this. Why did you need to change it?

Apart from that, I don't know much about Docker, so can't really review that. If you're confident that these changes to the docker setup are an improvement and won't break, I'm happy to merge them.

bwindels (commented):
As [mentioned](https://github.com/vector-im/hydrogen-web/pull/660#discussion_r804578721) elsewhere, I don't think we want to load the config file like this. Why did you need to change it?

Apart from that, I don't know much about Docker, so can't really review that. If you're confident that these changes to the docker setup are an improvement and won't break, I'm happy to merge them.

sandhose:
> As [mentioned](https://github.com/vector-im/hydrogen-web/pull/660#discussion_r804578721) elsewhere, I don't think we want to load the config file like this. Why did you need to change it?

This change is here to be able to change the config without rebuilding the app. Currently, the config lives in the bundled JS, which makes it impossible to change without rebuilding the app.

This allows to have on the filesystem a plain old `config.js` file that can be easily changed at runtime, and this is what is done in the Docker image: there is a [`config.js.tmpl`](https://github.com/vector-im/hydrogen-web/pull/655/files#diff-f6340e60dd9291d71f2d06a821cb6316bbb42ecc9860a870d63e8810358f8f3c) which is templated by [this script](https://github.com/vector-im/hydrogen-web/pull/655/files#diff-c4348ed3b58f6c552273507b4539afaed0af16cd86e8432a008f5813a31aeb84) on the start of the container, which fills it with environement variable.

The end result, is that one can run the image with a custom default homeserver without having to build a custom one:
```
docker --rm -p 8080:8080 -e DEFAULT_HOMESERVER=https://my-hs.com ghcr.io/vector-im/hydrogen-web
```

If that is not desirable, I can move the config handling changes in another PR and merge the docker improvements first

bwindels:
> If that is not desirable, I can move the config handling changes in another PR and merge the docker improvements first

Thanks, that would be good.

hughns:
Closing this in preference to the following smaller PRs:

- https://github.com/vector-im/hydrogen-web/pull/983
- https://github.com/vector-im/hydrogen-web/pull/948
- https://github.com/vector-im/hydrogen-web/pull/982
- https://github.com/vector-im/hydrogen-web/pull/980

I've parked running nginx as non-root for now.

PR #660, opened 2022-02-04
Draft: OIDC native login

sandhose:
This is on top of #655, so ignore the first few commits.

There are some changes that could be extracted in separate PRs if needed, including:

 - the refresh token mechanism, which could be adapted to work for MSC2918 (#235)
 - I added a `ObservableValue#map` method, which creates an observable from another one plus a mapper ; basically like `#flatMap` but not with the `flat` part
 - Make the `accessToken` an Observable in the HSAPI (needed for the refresh token)
 - The ability to override the access token for a single HSAPI call

There are still some stuff to do, especially around teardown, like pausing the token refresher when the session is not active


src/domain/login/StartOIDCLoginViewModel.js
```diff
@@ -0,0 +1,56 @@
...
+        });
+        this._homeserver = options.loginOptions.homeserver;
+    }
+
+    get isBusy() { return this._isBusy; }
+    get authorizationEndpoint() { return this._authorizationEndpoint; }
+
+    async start() {
+        const p = this._api.generateParams("openid");
+        await Promise.all([
```
bwindels, line 68: Shouldn't we only store the OIDC params of the flow we're currently doing only once you've actually started the flow and clicked the link? If so, the link should probably be a button. This is also how we do it for SSO.

src/domain/login/StartOIDCLoginViewModel.js
```diff
@@ -0,0 +1,56 @@
...
+import { ViewModel } from "../ViewModel.js";
+
+export class StartOIDCLoginViewModel extends ViewModel {
+    constructor(options) {
+        super(options);
+        this._isBusy = true;
+        this._authorizationEndpoint = null;
+        this._api = new OidcApi({
+            clientId: "hydrogen-web",
+            issuer: options.loginOptions.oidc.issuer,
```
bwindels: consider only passing in the oidc login method here, as you don't need the other ones?

src/matrix/net/OidcApi.ts
```diff
@@ -0,0 +1,221 @@
...
+        assert(Array.isArray(m.grant_types_supported) && m.grant_types_supported.includes("authorization_code"), "Supports the authorization_code grant type");
+        assert(Array.isArray(m.code_challenge_methods_supported) && m.code_challenge_methods_supported.includes("S256"), "Supports the authorization_code grant type");
+    }
+
+    async _generateCodeChallenge(
+        codeVerifier: string
+    ): Promise<string> {
+        const encoder = new TextEncoder();
+        const data = encoder.encode(codeVerifier);
+        const digest = await window.crypto.subtle.digest("SHA-256", data);
```
bwindels: you can use `platform.crypto.digest("SHA-256", data)` here, which deals better with cross-browser differences and maybe one day different platforms.

src/matrix/net/OidcApi.ts
```diff
@@ -0,0 +1,221 @@
...
+        assert(Array.isArray(m.response_types_supported) && m.response_types_supported.includes("code"), "Supports the code response type");
+        assert(Array.isArray(m.response_modes_supported) && m.response_modes_supported.includes("fragment"), "Supports the fragment response mode");
+        assert(Array.isArray(m.grant_types_supported) && m.grant_types_supported.includes("authorization_code"), "Supports the authorization_code grant type");
+        assert(Array.isArray(m.code_challenge_methods_supported) && m.code_challenge_methods_supported.includes("S256"), "Supports the authorization_code grant type");
+    }
+
+    async _generateCodeChallenge(
+        codeVerifier: string
+    ): Promise<string> {
+        const encoder = new TextEncoder();
```
bwindels: You can use `platform.encoding.utf8.encode()` here, which deals better with cross-browser differences and maybe one day different platforms.

src/matrix/net/OidcApi.ts
```diff
@@ -0,0 +1,221 @@
...
+    get metadataUrl() {
+        return new URL(WELL_KNOWN, this._issuer).toString();
+    }
+
+    get issuer() {
+        return this._issuer;
+    }
+
+    get redirectUri() {
+        return window.location.origin;
```
bwindels: We generally don't use DOM apis like `window` in any code not in platform/web, see startSSOLogin in StartSSOLoginViewModel how to deal with this.

src/platform/web/index.html
```diff
@@ -9,6 +9,7 @@
         <meta name="apple-mobile-web-app-status-bar-style" content="black">
         <meta name="apple-mobile-web-app-title" content="Hydrogen Chat">
         <meta name="description" content="A matrix chat application">
+        <script src="config.js"></script>
```
bwindels: I assume changing the way the config file is loaded is not needed for this PR? If not, better to remove this as I don't think this is how we'd ideally go about it.

vite.common-config.js
```diff
@@ -7,7 +7,6 @@ const version = manifest.version;
 
 const commonOptions = {
     logLevel: "warn",
-    publicDir: false,
```
bwindels: What is this change about?

src/matrix/net/TokenRefresher.ts
```diff
@@ -0,0 +1,125 @@
...
+    async start() {
+        if (this.needsRenewing) {
+            await this.renew();
+        }
+
+        this._renewingLoop();
+    }
+
+    stop() {
+        // TODO
```
bwindels: This would of course need to be done before it can be merged

src/domain/RootViewModel.js
```diff
@@ -77,7 +81,20 @@ export class RootViewModel extends ViewModel {
         } else if (loginToken) {
             this.urlCreator.normalizeUrl();
             if (this.activeSection !== "login") {
-                this._showLogin(loginToken);
+                this._showLogin({ loginToken });
+            }
+        } else if (oidcError) {
+            this._setSection(() => this._error = new Error(`OIDC error: ${oidcError[1]}`));
+        } else if (oidcCallback) {
+            this._setSection(() => this._error = new Error(`OIDC callback: state=${oidcCallback[0]}, code=${oidcCallback[1]}`));
```
bwindels: I guess this is debug code?

src/domain/RootViewModel.js
```diff
@@ -77,7 +81,20 @@ export class RootViewModel extends ViewModel {
         } else if (loginToken) {
             this.urlCreator.normalizeUrl();
             if (this.activeSection !== "login") {
-                this._showLogin(loginToken);
+                this._showLogin({ loginToken });
+            }
+        } else if (oidcError) {
+            this._setSection(() => this._error = new Error(`OIDC error: ${oidcError[1]}`));
```
bwindels: Ideally, we'd show this error in a dedicated screen, like for SSO, see CompleteSSOLoginViewModel

src/domain/navigation/index.js
```diff
@@ -105,11 +105,35 @@ export function addPanelIfNeeded(navigation, path) {
...
 export function parseUrlPath(urlPath, currentNavPath, defaultSessionId) {
+    const segments = [];
+
+    // Special case for OIDC callback
+    if (urlPath.includes("state")) {
+        const params = new URLSearchParams(urlPath);
+        if (params.has("state")) {
+            // This is a proper OIDC callback
+            if (params.has("code")) {
+                segments.push(new Segment("oidc-callback", [
```
bwindels: why not use an object as a value here rather than an array, then you can pass it straight to _showLogin in RootViewModel...

src/domain/navigation/index.js
```diff
@@ -14,23 +14,23 @@ See the License for the specific language governing permissions and
...
 }
 
 function allowsChild(parent, child) {
-    const {type} = child;
+    const { type } = child;
     switch (parent?.type) {
         case undefined:
             // allowed root segments
-            return type === "login"  || type === "session" || type === "sso" || type === "logout";
+            return type === "login" || type === "session" || type === "sso" || type === "logout" || type === "oidc-callback" || type === "oidc-error";
```
bwindels: Why not just go with one `oidc` segment for both an error and callback, with different values? Would reduce the amount of boilerplate I think.

src/platform/web/ui/login/LoginView.js
```diff
@@ -76,3 +77,14 @@ class StartSSOLoginView extends TemplateView {
         );
     }
 }
+
+class StartOIDCLoginView extends TemplateView {
+    render(t, vm) {
+        return t.div({ className: "StartOIDCLoginView" },
+            t.a({
+                className: "StartOIDCLoginView_button button-action secondary",
+                href: vm => (vm.isBusy ? "#" : vm.authorizationEndpoint),
```
bwindels: As mentioned with the view model, I'd only calculate the redirect url once we've clicked this button/link. Once it's clicked, we should show a spinner so people don't think it didn't work. Once the redirect url is known, we programatically redirect.

src/domain/login/StartOIDCLoginViewModel.js
```diff
@@ -0,0 +1,56 @@
...
+        const p = this._api.generateParams("openid");
+        await Promise.all([
+            this.platform.settingsStorage.setInt(`oidc_${p.state}_started_at`, Date.now()),
+            this.platform.settingsStorage.setString(`oidc_${p.state}_nonce`, p.nonce),
+            this.platform.settingsStorage.setString(`oidc_${p.state}_code_verifier`, p.codeVerifier),
+            this.platform.settingsStorage.setString(`oidc_${p.state}_homeserver`, this._homeserver),
+            this.platform.settingsStorage.setString(`oidc_${p.state}_issuer`, this._api.issuer),
+        ]);
+
+        this._authorizationEndpoint = await this._api.authorizationEndpoint(p);
```
bwindels: Could we postpone this until the button is clicked? It could be confusing that the button doesn't work as long as this request is running. If we do it afterwards, we can show a spinner until we can redirect.

src/matrix/Client.js
```diff
@@ -210,9 +243,41 @@ export class Client {
...
+            });
+
+            // TODO: stop/pause the refresher?
+            const tokenRefresher = new TokenRefresher({
+                oidcApi,
+                clock: this._platform.clock,
+                accessToken: sessionInfo.accessToken,
+                accessTokenExpiresAt: sessionInfo.accessTokenExpiresAt,
+                refreshToken: sessionInfo.refreshToken,
+                anticipation: 30 * 1000,
```
bwindels, line 317: What is this? A margin to make sure we're never too late?

src/matrix/Client.js
```diff
@@ -120,11 +122,29 @@ export class Client {
...
     }
 
-    queryLogin(homeserver) {
+    queryLogin(initialHomeserver) {
         return new AbortableOperation(async setAbortable => {
-            homeserver = await lookupHomeserver(homeserver, (url, options) => {
+            const { homeserver, issuer } = await lookupHomeserver(initialHomeserver, (url, options) => {
                 return setAbortable(this._platform.request(url, options));
             });
+            if (issuer) {
```
bwindels, line 135: This should follow the same pattern as the other login options (e.g. set all the applicable ones in _parseLoginOptions, and put the code below in a separate class, see SSOLoginHelper), unless there's a reason that wouldn't work?

bwindels (commented):
Thank you for your PR! Great start, while it still seems a bit WIP. I like the `observableValue.map()`! I had a first look and left some comments. In a few places it seems to deviate a bit from how things are doing for SSO, which is probably not intentional, unless I'm missing something?

src/matrix/Client.js
sandhose: The logic behind this is that if the sever advertises OIDC, we most probably want to use it. There will be a time where m.org will support both OIDC-based login and using the legacy endpoints, and we don't want to have a confusing UI during that time?

src/matrix/Client.js
sandhose: Yep, we refresh the token 30s before it expires to avoid doing requests with a just-expired token

src/domain/RootViewModel.js
sandhose: Yep, got rid of it

src/domain/login/StartOIDCLoginViewModel.js
sandhose: Done!

src/domain/navigation/index.js
sandhose: Done!

src/domain/navigation/index.js
sandhose: Done!

src/matrix/net/OidcApi.ts
sandhose: Fixed in  69a3404e

src/matrix/net/OidcApi.ts
sandhose: Done in  ea1bcd10

src/matrix/net/OidcApi.ts
sandhose: Done in  ea1bcd10

src/platform/web/index.html
sandhose: This PR/branch is on top of #672, mostly because I need this for the OIDC playground to work nicely

vite.common-config.js
sandhose: From #672

src/domain/login/StartOIDCLoginViewModel.js
sandhose: I did it this way so it is consistent with StartSSOLoginViewModel

src/platform/web/ui/login/LoginView.js
sandhose: Done in 62054783

src/domain/login/StartOIDCLoginViewModel.js
sandhose: Done in 62054783

src/matrix/net/TokenRefresher.ts
sandhose: Done in  a6d0585

src/matrix/Client.js
bwindels: Can you add a comment to the token refresher please?

src/matrix/Client.js
bwindels: Right, I'd prefer that decision to be taken higher up the stack though, in LoginViewModel, and not in SDK-level code. So I'd not prevent other login options from being advertised here in case OIDC support is found, and implement OIDC it just like the other options. And then choose to ignore the other options in the LoginViewModel.

src/domain/login/CompleteOIDCLoginViewModel.js
```diff
@@ -0,0 +1,87 @@
...
+    constructor(options) {
+        super(options);
+        const {
+            state,
+            code,
+            attemptLogin,
+        } = options;
+        this._request = options.platform.request;
+        this._encoding = options.platform.encoding;
+        this._crypto = options.platform.crypto;
```
bwindels, line 32: Nit: In view models, there no need to store the properties of the options in member variables as the options are stored in the ViewModel base class. In this case, you can just do `this.platform.crypto`/encoding/request at any point in a view model.

src/domain/login/CompleteOIDCLoginViewModel.js
```diff
@@ -0,0 +1,87 @@
...
+        this._state = state;
+        this._code = code;
+        this._attemptLogin = attemptLogin;
+        this._errorMessage = "";
+        this.performOIDCLoginCompletion();
```
bwindels, line 37: Generally we try (there is some places where we sin though, like LoginViewModel) to not call async methods from the constructor (which can't be async itself), unless there is really no other way and we can be 100% sure the method wont throw.

Usually, we deal with this by adding an async `start` or `init` method that is called from the parent view model after creating the child view model.

src/matrix/Client.js
```diff
@@ -121,11 +123,30 @@ export class Client {
...
+                        crypto: this._platform.crypto,
+                    });
+                    await oidcApi.validate();
+
+                    return {
+                        homeserver,
+                        oidc: { issuer },
+                    };
+                } catch (e) {
+                    console.log(e);
```
bwindels, line 153: This code might change wrt to the comment above, but wrt to console logging, it is usually only used during development. For production code we use the structured logging api from src/logging/

You could log an operation here by wrapping it in
```js
return this._platform.logger.run("queryLogin", async log => {
  // log in here is an ILogItem, happy to explain the API if needed
  // pass it through the call stack to log create a tree of log items
});
```

src/domain/navigation/URLRouter.js
```diff
@@ -125,6 +125,10 @@ export class URLRouter {
         return window.location.origin;
     }
 
+    createOIDCRedirectURL() {
```
bwindels: Just realized this is also not in /src/platform/web, but we can clean that up with the method above some other time, so fine for now :+1:

src/platform/web/index.html
bwindels: Can we take this out of this PR though?

src/matrix/Client.js
```diff
@@ -121,11 +123,30 @@ export class Client {
...
+            if (issuer) {
+                try {
+                    const oidcApi = new OidcApi({
+                        issuer,
+                        clientId: "hydrogen-web",
+                        request: this._platform.request,
+                        encoding: this._platform.encoding,
+                        crypto: this._platform.crypto,
+                    });
+                    await oidcApi.validate();
```
bwindels, line 144: Can we postpone doing OIDC  specific network calls until you've actually chosen to use it in StartOIDCLoginViewModel?
If there is a server misconfiguration (the only thing I assume that would cause validate to throw?), we can then show an error message in the UI and it will actually make sense to the user as it is related to their recent interaction. You would only find out after clicking OIDC, but I think that is better than just not offering the OIDC option, which could be puzzling.

So ofter the OIDC option issuer if there is an issuer in the well-known response, but postpone any other work until you actually choose it.

src/matrix/Client.js
```diff
@@ -222,9 +256,41 @@ export class Client {
             retryDelay: new ExponentialRetryDelay(clock.createTimeout),
             createMeasure: clock.createMeasure
         });
+
+        let accessToken;
+
+        if (sessionInfo.oidcIssuer) {
```
bwindels, line 302: Can you extract the OIDC/token refresh stuff into a separate method? `_loadSessionInfo` is already quite large.

src/matrix/net/OidcApi.ts
bwindels: I think this property can be removed now?

src/matrix/net/OidcApi.ts
```diff
@@ -0,0 +1,228 @@
...
+    state: string,
+    scope: string,
+    redirectUri: string,
+    nonce?: string,
+    codeVerifier?: string,
+};
+
+function assert(condition: any, message: string): asserts condition {
+    if (!condition) {
+        throw new Error(`Assertion failed: ${message}`);
```
bwindels, line 76: not sure "Assertion failed" adds a lot here, especially if we might want to show the error message in the UI at some point.

src/matrix/login/OIDCLoginMethod.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    async login(hsApi: HomeServerApi, _deviceName: string, log: ILogItem): Promise<Record<string, any>> {
+        const { access_token, refresh_token, expires_in } = await this._oidcApi.completeAuthorizationCodeGrant({
+            code: this._code,
+            codeVerifier: this._codeVerifier,
+            redirectUri: this._redirectUri,
+        });
+
+        // TODO: validate the id_token and the nonce claim
```
bwindels, line 64: Should this be done before merging?

src/domain/login/CompleteOIDCLoginViewModel.js
```diff
@@ -0,0 +1,87 @@
...
+        this._errorMessage = message;
+        this.emitChange("errorMessage");
+    }
+
+    async performOIDCLoginCompletion() {
+        if (!this._state || !this._code) {
+            return;
+        }
+        const code = this._code;
+        // TODO: cleanup settings storage
```
bwindels, line 52: Is this comment still relevant?

src/domain/navigation/index.js
```diff
@@ -105,11 +105,35 @@ export function addPanelIfNeeded(navigation, path) {
 }
 
 export function parseUrlPath(urlPath, currentNavPath, defaultSessionId) {
+    const segments = [];
+
+    // Special case for OIDC callback
+    if (urlPath.includes("state")) {
+        const params = new URLSearchParams(urlPath);
```
bwindels: can you extract the logic within the if statement here to a separate function?

src/matrix/login/OIDCLoginMethod.ts
```diff
@@ -0,0 +1,72 @@
...
+
+        // Do a "whoami" request to find out the user_id and device_id
+        const { user_id, device_id } = await hsApi.whoami({
```
bwindels, line 67: Just OOI; this is how it is supposed to work with OIDC? The OIDC endpoints only give an access token but not which user it is for?

src/matrix/login/OIDCLoginMethod.ts
```diff
@@ -0,0 +1,72 @@
...
+            log,
+            accessTokenOverride: access_token,
```
bwindels, line 69: If you create a new HomeServerApi object here, you could avoid having to add this option.

src/matrix/net/OidcApi.ts
```diff
@@ -0,0 +1,228 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import type {RequestFunction} from "../../platform/types/types";
+
+const WELL_KNOWN = ".well-known/openid-configuration";
+
+const RANDOM_CHARSET = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+const randomChar = () => RANDOM_CHARSET.charAt(Math.floor(Math.random() * 1e10) % RANDOM_CHARSET.length);
```
bwindels, line 24: For unit-testability I usually use platform.random() instead of Math.random(). May or may not be a concern here.

Also, would it make sense to use `crypto.getRandomValues()` (wrapped in the platform Crypto module) here rather to use a better random source for the nonce, etc...?

src/domain/navigation/index.js
```diff
@@ -105,11 +105,35 @@ export function addPanelIfNeeded(navigation, path) {
...
+                }));
+                return segments;
+            }
+        }
+    }
+
     // substr(1) to take of initial /
     const parts = urlPath.substr(1).split("/");
     const iterator = parts[Symbol.iterator]();
-    const segments = [];
```
bwindels: why is this removed?

src/domain/navigation/index.js
```diff
@@ -105,11 +105,35 @@ export function addPanelIfNeeded(navigation, path) {
...
-    let next; 
+    let next;
```
bwindels: What is going on here?

src/domain/navigation/index.js
```diff
@@ -191,6 +215,8 @@ export function stringifyPath(path) {
                 break;
             case "right-panel":
             case "sso":
+            case "oidc-callback":
+            case "oidc-error":
```
bwindels: should both of these be just "oidc"?

src/observable/ObservableValue.ts
```diff
@@ -174,6 +178,34 @@ export class FlatMapObservableValue<P, C> extends BaseObservableValue<C | undefi
...
+    onSubscribeFirst() {
+        super.onSubscribeFirst();
+        this.sourceSubscription = this.source.subscribe(() => {
+            this.emit(this.get());
+        });
+    }
+
+    get(): C {
+        const sourceValue = this.source.get();
+        return this.mapper(sourceValue);
```
bwindels, line 205: if the mapper creates a new object, the result of multiple get calls won't be equal to each other as the mapper is run each time. Probably better to run the mapper on update once and store the result in a member variable.

src/platform/types/types.ts
```diff
@@ -21,11 +21,12 @@ import type {ILogItem} from "../../logging/types";
 export interface IRequestOptions {
     uploadProgress?: (loadedBytes: number) => void;
     timeout?: number;
-    body?: EncodedBody;
+    body?: EncodedBody["body"];
```
bwindels: what is this about?

bwindels (commented):
Thanks for the changes, I've added some more comments.

PR #653, opened 2022-02-02
Draft: Scratch changes for `matrix-public-archive`

MadLittleMods:
Scratch changes for `matrix-public-archive` (not available yet)

See plan https://docs.google.com/document/d/1wP_TIqmBQjtt862vb2CWWmnmVxTyolcF3J1scuiYMdg/edit#

 1. Trying to make it faster/easier to build `hydrogen.es.js` for local linking and dev in `matrix-public-arhive` project
    - `yarn run vite build -c vite.sdk-lib-config.js --watch`
 1. Some random changes to accomodate using raw `EventEntry`'s
 
 ## Dev notes
 
  - https://github.com/vector-im/hydrogen-web/blob/master/doc/SDK.md
 
 
### Supporting the Node.js CommonJS environment (`package.json` `exports`)
 
Getting CommonJS vs EcmaScript Modules working, https://github.com/vector-im/hydrogen-web/commit/d7290bf750a5c136b56c159c33fa2fd972aa6acb
`package.json`
```
"main": "./hydrogen.cjs.js",
  "exports": {
      ".": {
          "import": "./lib-build/hydrogen.es.js",
          "require": "./lib-build/hydrogen.cjs.js"
      },
      "./paths/vite": "./paths/vite.js",
      "./style.css": "./style.css",
      "./assets/": "./asset-build/assets/"
  }
```

Also added `"./assets/": "./asset-build/assets/"` to the `exports` so we can simply serve them with `app.use(express.static(path.dirname(require.resolve('hydrogen-view-sdk/assets/'))));`

`eslint-plugin-node` will complain about these entries but that's because it doesn't support `exports`, see https://github.com/mysticatea/eslint-plugin-node/issues/255

#### `package.json` `exports` docs

Docs source: https://github.com/webpack/webpack.js.org/blob/627c26026678f2f44c40a7631c7799c04936ff40/src/content/guides/package-exports.mdx (served at https://webpack.js.org/guides/package-exports/)

> In general the `exports` field should contain an object
> where each properties specifies a sub path of the module request.
> For the examples above the following properties could be used:
> `"."` for `import "package"` and `"./sub/path"` for `import "package/sub/path"`.
> Properties ending with a `/` will forward a request with this prefix to the old file system lookup algorithm.
> For properties ending with `*`, `*` may take any value and any `*` in the property value is replaced with the taken value.
> 
> An example:
> 
> ```json
> {
>   "exports": {
>     ".": "./main.js",
>     "./sub/path": "./secondary.js",
[...]

PR #666, opened 2022-02-12
Fix missing reply text when message body is parsed as HTML in `linkedom` (SSR)

MadLittleMods:
Fix missing reply text when message body is parsed as HTML in [`linkedom`](https://github.com/WebReflection/linkedom) (SSR).

 - [`linkedom`](https://github.com/WebReflection/linkedom) is being used https://github.com/matrix-org/matrix-public-archive to server-side render (SSR) Hydrogen (`hydrogen-view-sdk`)
 - This is being fixed by using a explicit HTML wrapper boilerplate with `DOMParser` to get a matching result in the browser and `linkedom`.

Currently `parseHTML` is only used for HTML content bodies in events. Events with replies have content bodies that look like `<mx-reply>Hello</mx-reply> What's up` so they're parsed as HTML to strip out the `<mx-reply>` part.

Before | After
---  |  ---
 | 

## Dev notes

Before:
```js
// Browser (Chrome, Firefox)
new DOMParser().parseFromString(`<div>foo</div>`, "text/html").body.outerHTML;
// '<body><div>foo</div></body>'

// `linkedom` ❌
new DOMParser().parseFromString(`<div>foo</div>`, "text/html").body.outerHTML;
// '<body></body>'
```

After (consistent matching output):

```js
// Browser (Chrome, Firefox)
new DOMParser().parseFromString(`<!DOCTYPE html><html><body><div>foo</div></body></html>`, "text/html").body.outerHTML;
// '<body><div>foo</div></body>'

// `linkedom`
new DOMParser().parseFromString(`<!DOCTYPE html><html><body><div>foo</div></body></html>`, "text/html").body.outerHTML;
// '<body><div>foo</div></body>'
```

### Why does `linkedom` behavior differ from real browsers?

`linkedom` goal is to be close to the current DOM standard, but [not too close](https://github.com/WebReflection/linkedom#faq). Focused on the streamlined cases for server-side rendering (SSR).

Here is some context around getting `DOMParser` to interpret things better. The conclusion was to only support the explicit standard cases with a `<html><body></body></html>` specified instead of adding the magic HTML document creation and massaging that the browser does.

 - https://github.com/WebReflection/linkedom/issues/106
 - https://github.com/WebReflection/linkedom/pull/108
 
 
 ---
 
Part of https://github.com/vector-im/hydrogen-web/pull/653 to support server-side rendering Hydrogen for the [`matrix-public-archive`](https://github.com/matrix-org/matrix-public-archive) project.

### DIFF
--- a/src/platform/web/parsehtml.js
+++ b/src/platform/web/parsehtml.js
@@ -64,6 +64,6 @@ export function parseHTML(html) {
     // If DOMPurify uses DOMParser, can't we just get the built tree from it
     // instead of re-parsing?
     const sanitized = DOMPurify.sanitize(html, sanitizeConfig);
-    const bodyNode = new DOMParser().parseFromString(sanitized, "text/html").body;
+    const bodyNode = new DOMParser().parseFromString(`<!DOCTYPE html><html><body>${sanitized}</body></html>`, "text/html").body;
     return new HTMLParseResult(bodyNode);
 }


bwindels (approved):
LGTM!

MadLittleMods:
Thanks for the review and merge @bwindels 🐖

PR #670, opened 2022-02-15
Convert olm code to typescript

bwindels:
Related to #528

### DIFF
--- a/src/matrix/e2ee/DecryptionResult.ts
+++ b/src/matrix/e2ee/DecryptionResult.ts
@@ -26,35 +26,41 @@ limitations under the License.
  *                                      see DeviceTracker
  */
 
+import type {DeviceIdentity} from "../storage/idb/stores/DeviceIdentityStore";
 
+type DecryptedEvent = {
+    type?: string,
+    content?: Record<string, any>
+}
 
 export class DecryptionResult {
-    constructor(event, senderCurve25519Key, claimedEd25519Key) {
-        this.event = event;
-        this.senderCurve25519Key = senderCurve25519Key;
-        this.claimedEd25519Key = claimedEd25519Key;
-        this._device = null;
-        this._roomTracked = true;
+    private device?: DeviceIdentity;
+    private roomTracked: boolean = true;
+
+    constructor(
+        public readonly event: DecryptedEvent,
+        public readonly senderCurve25519Key: string,
+        public readonly claimedEd25519Key: string
+    ) {}
+
+    setDevice(device: DeviceIdentity) {
+        this.device = device;
     }
 
-    setDevice(device) {
-        this._device = device;
+    setRoomNotTrackedYet(): void {
+        this.roomTracked = false;
     }
 
-    setRoomNotTrackedYet() {
-        this._roomTracked = false;
-    }
-
-    get isVerified() {
-        if (this._device) {
-            const comesFromDevice = this._device.ed25519Key === this.claimedEd25519Key;
+    get isVerified(): boolean {
+        if (this.device) {
+            const comesFromDevice = this.device.ed25519Key === this.claimedEd25519Key;
             return comesFromDevice;
         }
         return false;
     }
 
-    get isUnverified() {
-        if (this._device) {
+    get isUnverified(): boolean {
+        if (this.device) {
             return !this.isVerified;
         } else if (this.isVerificationUnknown) {
             return false;
@@ -63,8 +69,8 @@ export class DecryptionResult {
         }
     }
 
-    get isVerificationUnknown() {
+    get isVerificationUnknown(): boolean {
         // verification is unknown if we haven't yet fetched the devices for the room
-        return !this._device && !this._roomTracked;
+        return !this.device && !this.roomTracked;
     }
 }
--- a/src/matrix/e2ee/DecryptionResult.ts
+++ b/src/matrix/e2ee/DecryptionResult.ts
@@ -43,7 +43,7 @@ export class DecryptionResult {
         public readonly claimedEd25519Key: string
     ) {}
 
-    setDevice(device: DeviceIdentity) {
+    setDevice(device: DeviceIdentity): void {
         this.device = device;
     }
 
--- a/src/matrix/e2ee/olm/Decryption.ts
+++ b/src/matrix/e2ee/olm/Decryption.ts
@@ -16,32 +16,52 @@ limitations under the License.
 
 import {DecryptionError} from "../common.js";
 import {groupBy} from "../../../utils/groupBy";
-import {MultiLock} from "../../../utils/Lock";
+import {MultiLock, ILock} from "../../../utils/Lock";
 import {Session} from "./Session.js";
-import {DecryptionResult} from "../DecryptionResult.js";
+import {DecryptionResult} from "../DecryptionResult";
+
+import type {OlmMessage, OlmPayload} from "./types";
+import type {Account} from "../Account";
+import type {LockMap} from "../../../utils/LockMap";
+import type {Storage} from "../../storage/idb/Storage";
+import type {Transaction} from "../../storage/idb/Transaction";
+import type {OlmEncryptedEvent} from "./types";
+import type * as OlmNamespace from "@matrix-org/olm";
+type Olm = typeof OlmNamespace;
 
 const SESSION_LIMIT_PER_SENDER_KEY = 4;
 
-function isPreKeyMessage(message) {
+type DecryptionResults = {
+    results: DecryptionResult[],
+    errors: DecryptionError[],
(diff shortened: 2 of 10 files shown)


src/matrix/e2ee/olm/Decryption.ts
```diff
@@ -266,7 +289,7 @@ class SenderKeyDecryption {
             try {
                 const plaintext = olmSession.decrypt(message.type, message.body);
                 session.save(olmSession);
-                session.lastUsed = this._timestamp;
+                session.data.lastUsed = this.timestamp;
```
bwindels, line 286: It seems like there was a bug here actually

src/matrix/e2ee/olm/Decryption.ts
```diff
@@ -16,32 +16,52 @@ limitations under the License.
...
-        this._storage = storage;
-        this._olm = olm;
-        this._senderKeyLock = senderKeyLock;
-    }
+    constructor(
+        private readonly account: Account,
+        private readonly pickleKey: string,
+        private readonly now: () => number,
+        private readonly ownUserId: string,
+        private readonly storage: Storage,
```
MidhunSureshR: Looks like this isn't used anywhere?

src/matrix/e2ee/olm/types.ts
```diff
@@ -0,0 +1,43 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+export type OlmMessage = {
+    type?: 0 | 1,
```
MidhunSureshR: Should this be a const enum?
```typescript
const enum OlmPayloadType { INITIAL_PREKEY_MESSAGE, ORDINARY_MESSAGE };
```

src/matrix/e2ee/olm/types.ts
```diff
@@ -0,0 +1,43 @@
...
+    sender_key?: string,
+    ciphertext?: {
+        [deviceCurve25519Key: string]: OlmMessage
+    }
+}
+
+export type OlmEncryptedEvent = {
+    type?: "m.room.encrypted",
+    content?: OlmEncryptedMessageContent
+    sender?: string
```
MidhunSureshR, line 38: Should there be a `sender` here? [This](https://spec.matrix.org/v1.2/client-server-api/#molmv1curve25519-aes-sha2) does not mention a sender?

src/matrix/e2ee/DecryptionResult.ts
```diff
@@ -26,35 +26,41 @@ limitations under the License.
...
+
+    constructor(
+        public readonly event: DecryptedEvent,
+        public readonly senderCurve25519Key: string,
+        public readonly claimedEd25519Key: string
+    ) {}
 
-    setDevice(device) {
-        this._device = device;
+    setDevice(device: DeviceIdentity) {
```
MidhunSureshR:
```suggestion
    setDevice(device: DeviceIdentity): void {
```

src/matrix/e2ee/olm/Decryption.ts
```diff
@@ -16,32 +16,52 @@ limitations under the License.
...
+    session: Session,
+    plaintext: string
+};
+
+function isPreKeyMessage(message: OlmMessage): boolean {
     return message.type === 0;
 }
 
-function sortSessions(sessions) {
+function sortSessions(sessions: Session[]) {
```
MidhunSureshR:
```suggestion
function sortSessions(sessions: Session[]): void {
```

src/matrix/e2ee/olm/Decryption.ts
```diff
@@ -219,21 +239,21 @@ export class Decryption {
...
-    constructor(senderKey, sessions, olm, timestamp) {
-        this.senderKey = senderKey;
-        this.sessions = sessions;
-        this._olm = olm;
-        this._timestamp = timestamp;
-    }
+    constructor(
+        public readonly senderKey: string,
+        public readonly sessions: Session[],
+        private readonly olm: Olm,
```
MidhunSureshR: This also looks like it's unused?

src/matrix/e2ee/olm/Decryption.ts
```diff
@@ -286,27 +309,27 @@ class SenderKeyDecryption {
...
-        this._senderKeyDecryptions = senderKeyDecryptions;
-        this._account = account;    
-        this.results = results;
-        this.errors = errors;
-        this._lock = lock;
-    }
+    constructor(
+        private readonly senderKeyDecryptions: SenderKeyDecryption[],
+        private readonly results: DecryptionResult[],
+        private readonly errors: DecryptionError[],
```
MidhunSureshR: unused?

src/matrix/e2ee/olm/Encryption.ts
```diff
@@ -36,19 +62,19 @@ const OTK_ALGORITHM = "signed_curve25519";
...
-        this._olm = olm;
-        this._olmUtil = olmUtil;
-        this._ownUserId = ownUserId;
-        this._storage = storage;
-        this._now = now;
-        this._pickleKey = pickleKey;
-        this._senderKeyLock = senderKeyLock;
-    }
+    constructor(
+        private readonly account: Account,
```
MidhunSureshR, line 66: Are we deciding that we longer want to prepend '_' to private members? We certainly don't need it with typescript but so far we've kept both.

MidhunSureshR (commented):
Sorry, took me a bit of time to get to this

src/matrix/e2ee/olm/types.ts
bwindels: TIL about const enums, but I can't seem to get it inlined though. When I run the build and look for PreKey, it is still there. Any ideas?

src/matrix/e2ee/olm/types.ts
bwindels: Yeah, I also noticed this when converting to typescript and figured it must be ok if the only place where it is used in validatePayload didn't throw. After some more digging now, it seems that this is defined for [to_device messages](https://spec.matrix.org/v1.2/client-server-api/#extensions-to-sync) in general and indeed confusingly omitted from the examples for `m.room.encrypted`.

src/matrix/e2ee/olm/Decryption.ts
bwindels: Well spotted, these needed to remain public actually.

src/matrix/e2ee/olm/Encryption.ts
bwindels: I think either are fine. I still use _prop when I can't make it `public readonly` and I create a `get prop() { return this._prop; }`. Would you like that we mandate either style?

src/matrix/e2ee/olm/Decryption.ts
bwindels: Well spotted, and it made me realize that I needed to fix the parameter order in Session.js (which isn't converted yet and hence didn't error)

src/matrix/e2ee/olm/types.ts
MidhunSureshR: I think this might be an esbuild limitation. [This](https://vitejs.dev/guide/features.html#typescript-compiler-options) says: 
> It is because esbuild only performs transpilation without type information, it doesn't support certain features like const enum and implicit type-only imports.

src/matrix/e2ee/olm/Encryption.ts
MidhunSureshR: no, just wanted to get some clarity on whether we were preferring one over the other.

src/matrix/e2ee/olm/Decryption.ts
```diff
@@ -16,32 +16,48 @@ limitations under the License.
...
-import {DecryptionResult} from "../DecryptionResult.js";
+import {MultiLock, ILock} from "../../../utils/Lock";
+import {Session} from "./Session";
+import {DecryptionResult} from "../DecryptionResult";
+import {OlmPayloadType} from "./types";
+
+import type {OlmMessage, OlmPayload} from "./types";
+import type {Account} from "../Account";
+import type {LockMap} from "../../../utils/LockMap";
+import type {Storage} from "../../storage/idb/Storage";
```
MidhunSureshR: This type isn't used anywhere?

MidhunSureshR (approved):
One more thing and we're good to go :+1:

src/matrix/e2ee/olm/types.ts
bwindels: Aha, thanks

PR #671, opened 2022-02-16
Implement token authenticated registration

MidhunSureshR:
- Adds registration stage for [MSC3231](https://github.com/matrix-org/matrix-doc/pull/3231)

### DIFF
--- a/src/matrix/registration/Registration.ts
+++ b/src/matrix/registration/Registration.ts
@@ -18,6 +18,7 @@ import type {HomeServerApi} from "../net/HomeServerApi";
 import type {BaseRegistrationStage} from "./stages/BaseRegistrationStage";
 import {DummyAuth} from "./stages/DummyAuth";
 import {TermsAuth} from "./stages/TermsAuth";
+import {TokenAuth, TOKEN_AUTH_TYPE} from "./stages/TokenAuth";
 import type {
     AccountDetails,
     RegistrationFlow,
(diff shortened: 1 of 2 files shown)


src/matrix/registration/stages/TokenAuth.ts
```diff
@@ -0,0 +1,43 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import {AuthenticationData} from "../types";
+import {BaseRegistrationStage} from "./BaseRegistrationStage";
+
+export const TOKEN_AUTH_TYPE = "org.matrix.msc3231.login.registration_token";
```
MidhunSureshR: So MSC3231 has been merged and the PR to [remove the unstable prefix](https://github.com/matrix-org/synapse/issues/11949) is also merged but a release containing this change has not been made. So this is okay for now?

src/matrix/registration/Registration.ts
```diff
@@ -108,6 +109,8 @@ export class Registration {
                 return new DummyAuth(session, params?.[type]);
             case "m.login.terms":
                 return new TermsAuth(session, params?.[type]);
+            case TOKEN_AUTH_TYPE:
```
bwindels: accept both the prefixed and non-prefixed type here, pass in the actual type used by the server and use that in generateAuthenticationData, but perhaps return the non-prefixed type to return from `get type` so it's easy to compare when seeing which stage you're in...

bwindels (approved):

PR #676, opened 2022-02-16
Convert /domain/navigation to typescript

MidhunSureshR:


src/domain/navigation/Navigation.ts
```diff
@@ -15,28 +15,53 @@ limitations under the License.
 */
 
 import {BaseObservableValue, ObservableValue} from "../../observable/ObservableValue";
+import type {allowsChild as AllowsChild} from "./index.js";
```
bwindels: I'd write the signature of the allowsChild in here as a type. Best not to have the definition be the implementation. You can probably also define it so it just takes keys of (a generic) SegmentType.

src/domain/navigation/Navigation.ts
```diff
@@ -60,7 +85,7 @@ export class Navigation {
         this._pathObservable.set(this._path);
     }
 
-    observe(type) {
+    observe(type: keyof SegmentType): SegmentObservable {
```
bwindels: Probably best to make Navigation generic over the SegmentType, so it can be used for different setups (e.g. in apps built on the sdk). EventEmitter is an example of how to do this. The allowsChild function would also be generic over the segment type, so that way you can enforce that the allowsChild function is for the right segment type.

src/domain/navigation/Navigation.ts
```diff
@@ -16,27 +16,39 @@ limitations under the License.
...
 import {BaseObservableValue, ObservableValue} from "../../observable/ObservableValue";
 
-export class Navigation {
-    constructor(allowsChild) {
+
+type AllowsChild<T> = (parent: Segment<T> | undefined, child: Segment<T>) => boolean;
+
+export type OptionalValue<T> = T extends true? [(undefined | true)?]: [T];
+
+export class Navigation<T> {
```
bwindels: if you use `keyof T`, don't you need to say T extends object?

src/domain/ViewModel.ts
```diff
@@ -27,17 +27,18 @@ import type {Platform} from "../platform/web/Platform";
...
     logger: ILogger
-    urlCreator: URLRouter
-    navigation: Navigation
+    urlCreator: URLRouter<N>
+    navigation: Navigation<N>
     emitChange?: (params: any) => void
 }
 
-export class ViewModel<O extends Options = Options> extends EventEmitter<{change: never}> {
+export class ViewModel<N extends {session: string} = SegmentType, O extends Options<N> = Options<N>> extends EventEmitter<{change: never}> {
```
bwindels: where is `session` needed?

src/domain/ViewModel.ts
```diff
@@ -131,11 +132,11 @@ export class ViewModel<O extends Options = Options> extends EventEmitter<{change
         return this.platform.logger;
     }
 
-    get urlCreator(): URLRouter {
+    get urlCreator(): URLRouter<N> {
         return this._options.urlCreator;
     }
 
-    get navigation(): Navigation {
+    get navigation(): Navigation<N> {
```
bwindels, line 144: remind me, the reason we use a generic type here is so that in projects that use the sdk, they can use their own segment type?

src/domain/navigation/Navigation.ts
```diff
@@ -16,27 +16,39 @@ limitations under the License.
 
 import {BaseObservableValue, ObservableValue} from "../../observable/ObservableValue";
 
-export class Navigation {
-    constructor(allowsChild) {
+
+type AllowsChild<T> = (parent: Segment<T> | undefined, child: Segment<T>) => boolean;
+
+export type OptionalValue<T> = T extends true? [(undefined | true)?]: [T];
```
bwindels, line 32: 👀 this one is taking some effort to decipher... what do the square brackets do? It's not an array if I'm not mistaken.

bwindels: It's a [tuple](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types), can you elaborate a bit why this is needed/how it is used?

src/domain/ViewModel.ts
bwindels: Ah, because of `_getLastSessionId` in URLRouter. Can we add this constraint in URLRouter rather than here? That way, we aren't constrained if we don't use the URLRouter.

bwindels (commented):
Had a first look and looks great. Just a few questions.

src/domain/ViewModel.ts
MidhunSureshR: Exactly. The navigation is mostly always going to be accessed via this getter. One downside of this is that all viewmodels in projects that use hydrogen SDK will always need to specify the type parameter.

src/domain/ViewModel.ts
MidhunSureshR: Typescript complains if we don't duplicate the constraint here. I've made a [change](https://github.com/vector-im/hydrogen-web/pull/676/commits/48fd2c3aa25eee1f2328e5febf8feddeda229821) that would allow other projects to use ViewModel even if they use a SegmentType that lacks support for session segment.

src/domain/navigation/Navigation.ts
MidhunSureshR: Yeah this is a bit unsightly.
It is used like this (code from Segment ctor):
```js
    constructor(public type: K, ...value: OptionalValue<T[K]>) {
        this.value = (value[0] === undefined ? true : value[0]) as unknown as T[K];
    }
```
The `value` parameter is made a [rest parameter](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generic-parameter-defaults) so that we can make it both optional and give it a default value of undefined depending on the segment type.
So for example, these should be accepted:
```js
new Segment("create-room"); // value is optional, internally given the value true
new Segment("create-room", true);
```
where as these should not be accepted:
```js
new Segment("room") // error: second argument is needed and it should be string (room-id)
```
`OptionalValue` is basically stating that if `SegmentType[type] = true`:
1. Allow this argument to be optional
2. Give it a default value of undefined
3. Also allow it to be true

I tried a lot of things but they would all require that we explicitly pass undefined as second argument, eg:
```js
constructor(public type: K, value: T[K] extends true? undefined: T[K]) {
    this.value = (value[0] === undefined ? true : value[0]) as unknown as T[K];
}

const s: Segment<SegmentType> = new Segment("create-room"); // error: An argument for 'value' was not provided
const s: Segment<SegmentType> = new Segment("create-room", undefined); //works, but not what we want
```
or they would make the second parameter always optional.

bwindels: right, makes sense! Can you perhaps put a summary of this in a comment?

src/domain/LogoutViewModel.ts
```diff
@@ -16,10 +16,11 @@ limitations under the License.
 
 import {Options, ViewModel} from "./ViewModel";
 import {Client} from "../matrix/Client.js";
+import {SegmentType} from "./navigation/index";
 
 type LogoutOptions = { sessionId: string; } & Options;
 
-export class LogoutViewModel extends ViewModel<LogoutOptions> {
+export class LogoutViewModel extends ViewModel<SegmentType, LogoutOptions> {
```
bwindels: Would it make sense to swap the order of the generic params in ViewModel? I think within Hydrogen, once everything is TS-ified, every view model will specify the options type, but I _think_ none will explicitly set the segment type? IIRC, that would avoid have to the above in most view models (once TS).

bwindels (commented):
Apart from comments, looks good to me!

src/domain/LogoutViewModel.ts
MidhunSureshR:
```ts
export class ViewModel<N extends object = SegmentType, O extends Options<N> = Options<N>> extends EventEmitter<{change: never}>
```
Not sure if it can be done that way since the Options generic is dependent on the first one.

bwindels (approved):

PR #678, opened 2022-02-17
Check options exist on emitChange

MidhunSureshR:
This was causing the composer to throw because it does not have options


bwindels (approved):
good spot

PR #690, opened 2022-02-25
Add `data-event-id="$xxx"` attributes to timeline items for easy selecting in end-to-end tests

MadLittleMods:
Add `data-event-id="$xxx"` attributes to timeline items for easy selecting in end-to-end tests.

Example test assertions: https://github.com/matrix-org/matrix-public-archive/blob/db6d3797d74104ad7c93e572ed106f1a685a90d0/test/e2e-tests.js#L248-L252

```js
// Make sure the $abc event on the page has "foobarbaz" text in it
assert.match(
  dom.document.querySelector(`[data-event-id="$abc"]`).outerHTML,
  new RegExp(`.*foobarbaz.*`)
);
```

It's also nice for debugging to see if a given event is on the page by searching the DOM in the browser devtools.

### DIFF
--- a/src/domain/session/room/timeline/tiles/BaseMessageTile.js
+++ b/src/domain/session/room/timeline/tiles/BaseMessageTile.js
@@ -49,10 +49,6 @@ export class BaseMessageTile extends SimpleTile {
         return `https://matrix.to/#/${encodeURIComponent(this.sender)}`;
     }
 
-    get eventId() {
-        return this._entry.id;
-    }
-
     get displayName() {
         return this._entry.displayName || this.sender;
     }
--- a/src/domain/session/room/timeline/tiles/BaseMessageTile.js
+++ b/src/domain/session/room/timeline/tiles/BaseMessageTile.js
@@ -49,6 +49,10 @@ export class BaseMessageTile extends SimpleTile {
         return `https://matrix.to/#/${encodeURIComponent(this.sender)}`;
     }
 
+    get eventId() {
+        return this._entry.id;
+    }
+
     get displayName() {
         return this._entry.displayName || this.sender;
     }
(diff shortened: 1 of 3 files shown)


src/platform/web/ui/session/room/timeline/BaseMessageView.js
```diff
@@ -40,14 +40,17 @@ export class BaseMessageView extends TemplateView {
...
+        const li = t.el(this._tagName, {
+            className: {
+                "Timeline_message": true,
+                own: vm.isOwn,
+                unsent: vm.isUnsent,
+                unverified: vm.isUnverified,
+                disabled: !this._interactive,
+                continuation: vm => vm.isContinuation,
+            },
+            'data-event-id': vm.eventId
```
MadLittleMods, line 52: Only change is here to add the `data-event-id="$abc"` attribute

bwindels (approved):
Looks good, thanks!

src/domain/session/room/timeline/tiles/BaseMessageTile.js
```diff
@@ -49,6 +49,10 @@ export class BaseMessageTile extends SimpleTile {
         return `https://matrix.to/#/${encodeURIComponent(this.sender)}`;
     }
 
+    get eventId() {
```
bwindels: Actually, might be better if this were in SimpleTile to ensure it's always there. There's a few classes that extent SimpleTile.

bwindels (changes requested):

src/domain/session/room/timeline/tiles/BaseMessageTile.js
MadLittleMods: Good call 👍

MadLittleMods:
Thanks for the review @bwindels 🐨

Issue #579, opened 2021-11-08
Make app work without olm


bwindels:
Currently, the two awaits of the olmPromise don't handle the error that it can throw when the olm file is not found (if we want to distribute a copy of hydrogen without olm, for example for legal reasons). We could probably await the olm promise once at the beginning of starting with login or an existing session and handle the error there, and if olm is not present, don't look for dehydrated devices, ... might need some checks if olm is present deeper down the stack as this scenario has been taken into account but not really tested.

PR #691, opened 2022-02-25
Only initialize `Crypto` when olm is provided

MadLittleMods:
Only initialize `Crypto` when olm is provided.

As a note, `window.crypto.subtle` is only avaialble in secure contexts(`window.isSecureContext`). Relevant error if you crypto is used in a non-secure context like a local LAN IP `http://192.168.1.151:3050/`
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'deriveBits')
	at new Crypto
	at new Platform
	at mountHydrogen
```

For my use-case with https://github.com/matrix-org/matrix-public-archive, I don't need crypto/encryption at all so if there was a way to completely disable, that would be even better. Related to https://github.com/vector-im/hydrogen-web/issues/579

---

Docs:

 - https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts
 - https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle
    - "Secure context: This feature is available only in secure contexts (HTTPS), in some or all supporting browsers."

### DIFF
--- a/src/platform/web/Platform.js
+++ b/src/platform/web/Platform.js
@@ -143,7 +143,10 @@ export class Platform {
             this._serviceWorkerHandler.registerAndStart(assetPaths.serviceWorker);
         }
         this.notificationService = new NotificationService(this._serviceWorkerHandler, config.push);
-        this.crypto = new Crypto(cryptoExtras);
+        // `window.crypto.subtle` is only available in a secure context
+        if(window.isSecureContext) {
+            this.crypto = new Crypto(cryptoExtras);
+        }
         this.storageFactory = new StorageFactory(this._serviceWorkerHandler);
         this.sessionInfoStorage = new SessionInfoStorage("hydrogen_sessions_v1");
         this.estimateStorageUsage = estimateStorageUsage;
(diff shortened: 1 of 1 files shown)


bwindels:
If you don't pass in the olm paths to the platform class, isn't encryption disabled? That's the _idea_, but this is very untested ...

src/platform/web/Platform.js
```diff
@@ -143,7 +143,10 @@ export class Platform {
             this._serviceWorkerHandler.registerAndStart(assetPaths.serviceWorker);
         }
         this.notificationService = new NotificationService(this._serviceWorkerHandler, config.push);
-        this.crypto = new Crypto(cryptoExtras);
+        // `window.crypto.subtle` is only available in a secure context
+        if(window.isSecureContext) {
```
bwindels: I think I would rather see a solution to disable e2ee that makes it the explicit choice of the consuming app, rather than whether or not you happened to have activated https when deploying. The latter will be just as annoying to debug as it is now.
I think it's already sort of the idea that if you don't pass the olm paths in the platform options, that e2ee will be disabled, although this is very untested, so we can adopt that approach here. 

```suggestion
        if(this._assetPaths.olm) {
```

bwindels (changes requested):
Thanks for your PR, there might be a more explicit way to disable e2ee though.

bwindels (approved):

MadLittleMods:
Thanks for the review and merge @bwindels 🐻

Good shout on the alternative way to disable 👍

PR #700, opened 2022-03-03
Add observeNavigation in ViewModel

ajbura:
These changes added observeNavigation method in ViewModel class to help observe navigation without adding much boilerplate code in view models.

### DIFF
--- a/src/domain/ViewModel.ts
+++ b/src/domain/ViewModel.ts
@@ -58,9 +58,9 @@ export class ViewModel<O extends Options = Options> extends EventEmitter<{change
         return this._options[name];
     }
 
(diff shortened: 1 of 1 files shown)


src/domain/ViewModel.ts
```diff
@@ -58,6 +58,14 @@ export class ViewModel<O extends Options = Options> extends EventEmitter<{change
         return this._options[name];
     }
 
+    observeNavigation(type: string, onChange: (value: string | true | undefined, type: string) => void) {
```
bwindels, line 61: Once #676 is merged, we can make the type an value type-safe here, but this works for now :+1:

bwindels (approved):

PR #703, opened 2022-03-14
Theming - Postcss plugin to replace URL values with css variable

MidhunSureshR:
- Extract `url(..)`s into css variables if the link ends with `.svg` and contains color query parameters (foo/bar/icon.svg?primary=blue&secondary=red)
- Populates shared map (compiledVariables) with the extracted icon links.

[see structure of shared map](https://github.com/vector-im/hydrogen-web/pull/704)

### DIFF
--- a/scripts/postcss/css-url-to-variables.js
+++ b/scripts/postcss/css-url-to-variables.js
@@ -15,6 +15,10 @@ limitations under the License.
 */
 
 const valueParser = require("postcss-value-parser");
+/**
+ * This plugin extracts content inside url() into css variables.
+ * The extracted css variables are added to the :root section.
+ */
 let counter;
 const urlVariables = new Map();
 const idToPrepend = "icon-url";
--- a/scripts/postcss/css-url-to-variables.js
+++ b/scripts/postcss/css-url-to-variables.js
@@ -15,12 +15,13 @@ limitations under the License.
 */
 
 const valueParser = require("postcss-value-parser");
+
 /**
  * This plugin extracts content inside url() into css variables.
  * The extracted css variables are added to the :root section.
  */
 let counter;
-const urlVariables = new Map();
+let urlVariables;
 const idToPrepend = "icon-url";
 
 function extractUrl(decl) {
@@ -58,6 +59,7 @@ function addResolvedVariablesToRootSelector(root, { Rule, Declaration }) {
  * @type {import('postcss').PluginCreator}
  */
 module.exports = (opts = {}) => {
+    urlVariables = new Map();
     counter = 0;
     return {
         postcssPlugin: "postcss-url-to-variable",
--- a/scripts/postcss/css-url-to-variables.js
+++ b/scripts/postcss/css-url-to-variables.js
@@ -24,7 +24,7 @@ let counter;
 let urlVariables;
 const idToPrepend = "icon-url";
 
-function extractUrl(decl) {
+function findAndReplaceUrl(decl) {
     const value = decl.value;
     const parsed = valueParser(value);
     parsed.walk(node => {
@@ -60,7 +60,7 @@ module.exports = (opts = {}) => {
         postcssPlugin: "postcss-url-to-variable",
 
         Once(root, { Rule, Declaration }) {
-            root.walkDecls(decl => extractUrl(decl));
+            root.walkDecls(decl => findAndReplaceUrl(decl));
             addResolvedVariablesToRootSelector(root, { Rule, Declaration });
         },
     };
--- a/scripts/postcss/css-url-to-variables.js
+++ b/scripts/postcss/css-url-to-variables.js
@@ -33,7 +33,7 @@ function extractUrl(decl) {
         }
         const urlStringNode = node.nodes[0];
         const variableName = `${idToPrepend}-${counter++}`;
-        urlVariables.set(variableName, `"${urlStringNode.value}"`);
+        urlVariables.set(variableName, urlStringNode.value);
         const varNode = {
             type: "function",
             value: "var",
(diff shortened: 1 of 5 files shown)


scripts/postcss/css-url-to-variables.js
```diff
@@ -0,0 +1,75 @@
...
+function extractUrl(decl) {
+    const value = decl.value;
+    const parsed = valueParser(value);
+    parsed.walk(node => {
+        if (node.type !== "function" || node.value !== "url") {
+            return;
+        }
+        const urlStringNode = node.nodes[0];
+        const variableName = `${idToPrepend}-${counter++}`;
+        urlVariables.set(variableName, `"${urlStringNode.value}"`);
```
bwindels: Wrt to adding the double quotes here; like for the other PR, I'd keep CSS specific syntax out of your data structures and only add them when you're actually outputting CSS, below.

scripts/postcss/css-url-to-variables.js
```diff
@@ -0,0 +1,75 @@
...
+        const varNode = {
+            type: "function",
+            value: "var",
+            nodes: [{ type: "word", value: `--${variableName}` }],
+        };
+        // replace the url-string node with this var-node
+        node.nodes[0] = varNode;
+    });
+    decl.assign({prop: decl.prop, value: parsed.toString()})
```
bwindels, line 43: IIRC you need to replace the entire `url()` expression with a `var()` expression for it to work in browsers, e.g. this
```css
button {
  background: no-repeat url("foo-icon.svg");
}
```
becomes
```css
button {
  background: no-repeat var(--icon1);
}

:root {
  --icon1: url("foo-icon.svg");
}

File: scripts/postcss/css-url-to-variables.js [hunk 3]
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * This plugin extracts content inside url() into css variables.
+ * The extracted css variables are added to the :root section.
+ */
+let counter;
+let urlVariables;
+const idToPrepend = "icon-url";
+
+function extractUrl(decl) {
```
bwindels: Is does more than extract, right?
```suggestion
function findAndReplaceUrl(decl) {
```

scripts/postcss/css-url-to-variables.js
```diff
@@ -0,0 +1,75 @@
...
+    const value = decl.value;
+    const parsed = valueParser(value);
+    parsed.walk(node => {
+        if (node.type !== "function" || node.value !== "url") {
+            return;
+        }
+        const urlStringNode = node.nodes[0];
```
bwindels: We should probably only replace images that have a `?colors` query string

bwindels: We should probably only replace images that:
 1. are SVGs
 2. have a `?colors` query string

scripts/postcss/css-url-to-variables.js
```diff
@@ -0,0 +1,75 @@
...
+        node.nodes[0] = varNode;
+    });
+    decl.assign({prop: decl.prop, value: parsed.toString()})
+}
+
+function addResolvedVariablesToRootSelector(root, { Rule, Declaration }) {
+    const newRule = new Rule({ selector: ":root", source: root.source });
+    // Add derived css variables to :root
+    urlVariables.forEach((value, key) => {
+        const declaration = new Declaration({ prop: `--${key}`, value });
```
bwindels: So where is the SVG color replacing supposed to take place if not here? I assumed we'd write to a new file here with the colors replaced.

scripts/postcss/css-url-to-variables.js
```diff
@@ -15,6 +15,10 @@ limitations under the License.
 */
 
 const valueParser = require("postcss-value-parser");
+/**
+ * This plugin extracts content inside url() into css variables.
```
bwindels: can you also mention how we're using this and for what reason (e.g. to replace colors in SVG icons with theme colors)

scripts/postcss/css-url-to-variables.js
bwindels: This is handled by another plugin in #707

scripts/postcss/tests/css-url-to-variables.test.js
```diff
@@ -46,6 +47,21 @@ module.exports.tests = function tests() {
...
+            }
+            button {
+                background: url("/home/foo/bar/cool.svg?primary=blue&secondary=green");
+            }`;
+            const expectedObject = {
+                "icon-url-0": "../img/image.svg?primary=main-color--darker-20",
+                "icon-url-1": "/home/foo/bar/cool.svg?primary=blue&secondary=green",
+            };
+            await postcss([plugin({compiledVariables})]).process(inputCSS, { from: "/foo/bar/test.css", });
+            assert.deepEqual(expectedObject, compiledVariables.get("/foo/bar")["icon"]);
```
bwindels: are the compiled variables cached by the directory of the source file (rather than the source file name)? :thinking:

scripts/postcss/css-url-to-variables.js
```diff
@@ -55,15 +55,16 @@ function addResolvedVariablesToRootSelector(root, { Rule, Declaration }) {
 
 function populateMapWithIcons(map, cssFileLocation) {
     const location = cssFileLocation.match(/(.+)\/.+\.css/)?.[1];
-    if (map.has(location)) {
+    const sharedObject = map.get(location);
+    if (sharedObject?.["icon"]) {
```
bwindels: I do wonder if the complication of only setting icon once per file is worth it given we need to redo all the complicated work anyway (parsing the css) and the `map.set` should have a very minimal runtime cost. Or is this not just about performance and will something break if we set it multiple times (for each theme variant)? Anyway not a big deal, fine to leave it as it is.

bwindels (approved):
lgtm, just a few questions

scripts/postcss/tests/css-url-to-variables.test.js
MidhunSureshR: Yep, so we'd have a directory like `/themes` and each theme would be in their own sub-directory, like `/themes/awesome-dark`, `/themes/hydrogen-default` ...
These theme directories are passed to the rollup plugin so that it knows where to find the `.json` and `.css`.
The source file name would always be something like `index.css` or `theme.css`.

bwindels (approved):

PR #704, opened 2022-03-14
Theming - Rollup plugin to enumerate and compile themes (and their variants)

MidhunSureshR:
### Structure of the shared map
The keys are locations of each theme eg(`/foo/dev/hydrogen/.../themes/element_default`). The values are objects of the form 
```json
{
"derived_variables": array of derived variables,
"icon": object whose keys are icon-variable name (eg: icon-url-1) and values are icon URLs extracted via css-url-variables plugin,
}
```

### DIFF
--- a/scripts/build-plugins/rollup-plugin-build-themes.js
+++ b/scripts/build-plugins/rollup-plugin-build-themes.js
@@ -38,8 +38,8 @@ module.exports = function buildThemes(options) {
             for (const location of manifestLocations) {
                 manifest = require(`${location}/manifest.json`);
                 variants = manifest.values.variants;
+                const themeName = manifest.name;
                 for (const [variant, details] of Object.entries(variants)) {
-                    const themeName = manifest.name;
                     const fileName = `theme-${themeName}-${variant}.css`;
                     if (details.default) {
                         // This theme is the default for when Hydrogen launches for the first time
@@ -57,6 +57,12 @@ module.exports = function buildThemes(options) {
                         fileName,
                     });
                 }
+                // emit the css as runtime theme bundle
+                this.emitFile({
+                    type: "chunk",
+                    id: `${location}/theme.css`,
+                    fileName: `theme-${themeName}-runtime.css`,
+                });
             }
         },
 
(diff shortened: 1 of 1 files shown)


scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -0,0 +1,174 @@
...
+        async buildStart() {
+            const { manifestLocations } = options;
+            for (const location of manifestLocations) {
+                manifest = require(`${location}/manifest.json`);
+                variants = manifest.values.variants;
+                const themeName = manifest.name;
+                for (const [variant, details] of Object.entries(variants)) {
+                    const fileName = `theme-${themeName}-${variant}.css`;
+                    if (details.default) {
+                        // This theme is the default for when Hydrogen launches for the first time
```
bwindels: The comment seems a bit off, this is about the default variant rather than the default theme (which is set in the config file).

scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -0,0 +1,174 @@
...
+                    type: "chunk",
+                    id: `${location}/theme.css?type=runtime`,
+                    fileName: `theme-${themeName}-runtime.css`,
+                });
+            }
+        },
+
+        async load(id) {
+            const result = id.match(/(.+)\/theme.css\?variant=(.+)/);
+            if (result) {
```
bwindels, line 111: where does `type=runtime` get loaded?

scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -0,0 +1,174 @@
...
+    let manifest, variants, defaultDark, defaultLight;
+
+    return {
+        name: "build-themes",
+        enforce: "pre",
+
+        async buildStart() {
+            const { manifestLocations } = options;
+            for (const location of manifestLocations) {
+                manifest = require(`${location}/manifest.json`);
```
bwindels, line 79: Nit: I'd keep the naming convention out of this plugin perhaps ...

bwindels (commented):
A few questions, but looks good to me!

scripts/build-plugins/rollup-plugin-build-themes.js
MidhunSureshR: It gets loaded like a normal .css file (the query parameter is ignored). `type=runtime` is used in `compile-css-variables` to skip running the derive logic.

bwindels (approved):

PR #705, opened 2022-03-14
Group calls

bwindels:


robertlong:
Just saw one minor bug. It looks like when trying to connect/reconnect to a member that never successfully connected it has an undefined `connection` object

Screen Shot 2022-11-09 at 2 08 16 PM

PR #707, opened 2022-03-25
Theming - Postcss plugin to process URLs

MidhunSureshR:
This plugin replaces each url in the css with the result of running a callback. 
We'll use the callback to colorize svgs

### DIFF
--- a/scripts/postcss/css-url-processor.js
+++ b/scripts/postcss/css-url-processor.js
@@ -80,7 +80,6 @@ module.exports = (opts = {}) => {
             of running replacer(url)
             */
             root.walkDecls(decl => processURL(decl, opts.replacer, colorMap));
-            console.log("result", colorMap);
         },
     };
 };
--- a/scripts/postcss/css-url-processor.js
+++ b/scripts/postcss/css-url-processor.js
@@ -1,0 +1,81 @@
+/*
+Copyright 2021 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


scripts/postcss/svg-colorizer.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Builds a new svg with the colors replaced and returns its location.
+ * @param {string} svgLocation The location of the input svg file
+ * @param {string} primaryColor Primary color for the new svg
+ * @param {string} secondaryColor Secondary color for the new svg
+ */
+module.exports.buildColorizedSVG = function (svgLocation, primaryColor, secondaryColor) {
+    const svgCode = fs.readFileSync(svgLocation, { encoding: "utf8"});
+    let coloredSVGCode = svgCode.replaceAll("#ff00ff", primaryColor);
+    coloredSVGCode = coloredSVGCode.replaceAll("#00ffff", secondaryColor);
```
bwindels, line 28: can we throw an error here if nothing was replaced for both the primary and secondary color? It probably means you got the hardcoded colors wrong (e.g. uppercased) in your SVG.

scripts/postcss/css-url-processor.js
```diff
@@ -0,0 +1,87 @@
...
+        throw new Error(`Variable ${secondary} not found in resolved color variables!`);
+    }
+    return [primaryColor, secondaryColor];
+}
+
+function processURL(decl, replacer, colorMap) {
+    const value = decl.value;
+    const parsed = valueParser(value);
+    parsed.walk(async node => {
+        if (node.type !== "function" || node.value !== "url") {
```
bwindels, line 42: I thought the urls would already be replaced with a variable at this point by [the other plugin](https://github.com/vector-im/hydrogen-web/pull/703/files#diff-edd3a0e0c14c7f3a611b4b972dcc1d9e86be041945df96c9fab21ab87e264a31R40)?

MidhunSureshR: Right, the URLs would be in the :root section like:
```css
:root {
    --icon-url-1: url('./foo/bar/../test.svg?primary=accent-color);
}
```
The code is trying to find these URL values.

bwindels: Oh, right, that makes sense!

bwindels (approved):

PR #709, opened 2022-03-27
Theming - postcss-compile-variables improvement

MidhunSureshR:
- Passes the resolved colors to other postcss plugins via `result.messages`
- Converts postcss-value-parser to a devDependency
- Includes code to skip running plugin if the css file is a runtime theme
- Populates shared map with derived variables

[see structure of shared map](https://github.com/vector-im/hydrogen-web/pull/704)

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "node-html-parser": "^4.0.0",
     "postcss-css-variables": "^0.18.0",
     "postcss-flexbugs-fixes": "^5.0.2",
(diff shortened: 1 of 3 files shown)


bwindels (approved):
Same nit as for icons whether or not the code to only set the compiledVariables once doesn't complicate things more than it has to, but possibly missing something (see other PR where this is done for icons). Otherwise looks great!

bwindels (approved):

PR #712, opened 2022-04-07
Theming - Convert existing theme to use new theming architecture

MidhunSureshR:
- [x] Convert icons to use predefined color (fuchsia, #ff00ff)


vite.sdk-lib-config.js
```diff
@@ -1,6 +1,7 @@
 const path = require("path");
 const mergeOptions = require('merge-options');
-const commonOptions = require("./vite.common-config.js");
+const themeBuilder = require("./scripts/build-plugins/rollup-plugin-build-themes");
```
bwindels: this doesn't seem to be used

scripts/sdk/build.sh
```diff
@@ -11,7 +11,7 @@ cp doc/SDK.md target/README.md
 pushd target
 pushd asset-build/assets
 mv main.*.js ../../main.js
-mv index.*.css ../../style.css
+mv *light.*.css ../../style.css
```
bwindels: I'd suggest creating a copy of the light css file for backwards compatibility, but also rename the hashed file to not have a content hash (so imports don't need to include the content hash).
can we rename the dark theme file to not have a content hash as well?

bwindels (commented):
great, just two things

bwindels (approved):

PR #716, opened 2022-04-10
Theming - Support theming in dev server

MidhunSureshR:


scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -106,18 +143,57 @@ module.exports = function buildThemes(options) {
...
-            if (result) {
-                const [, location, variant] = result;
-                const cssSource = await readCSSSource(location);
-                const config = variants[variant];
-                return await appendVariablesToCSS(config.variables, cssSource);
+            if (isDevelopment) {
+                if (id.startsWith(resolvedVirtualModuleId)) {
+                    let [theme, variant, file] = id.substr(resolvedVirtualModuleId.length).split("/");
+                    if (theme === "default") {
+                        theme = "Element";
```
bwindels: probably best to only use lowercase in theme names?

scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -68,11 +96,20 @@ function parseBundle(bundle) {
...
 
+        configResolved(config) {
+            if (config.command === "serve") {
+                isDevelopment = true;
+            }
+        },
+
         async buildStart() {
             const { manifestLocations } = options;
             for (const location of manifestLocations) {
```
bwindels: Hmm, perhaps it would be worthwhile to pass the themes as an object mapping the theme name to the location (and the [default theme name](https://gist.github.com/bwindels/16930ddb0bc0f0e14009f8880793f7bf?permalink_comment_id=4128583#gistcomment-4128583)). This is closer to what the config file will look like, and then we can get rid of the lookup functions that load every manifest to find a theme above.

scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -106,18 +143,57 @@ module.exports = function buildThemes(options) {
...
+                        theme = "Element";
+                    }
+                    if (!variant || variant === "default") {
+                        variant = "light";
+                    }
+                    if (!file) {
+                        file = "index.js";
+                    }
+                    switch (file) {
+                        case "index.js": {
```
bwindels, line 153: can we add a comment somewhere why we have to take this different approach for both dev and build mode, e.g. that to be able to have the theme css file pass the normal css pipeline, going through postcss, we can't dynamically return it from our plugin (I think that is the reason?). Also, something something the generate bundle hook isn't supported on dev mode, can't remember the full reason now?

bwindels (commented):
looks good, just a few things.

scripts/build-plugins/rollup-plugin-build-themes.js
bwindels: and have an option to set the default theme rather than hardcode it here

bwindels (commented):
ah forgot to mention, you'll need to change how the theme is imported into the html file with this PR, right? E.g. replace the link tag with a script tag. And have some `transformHtml` logic to undo that at build time?

scripts/build-plugins/rollup-plugin-build-themes.js
bwindels: this is handled in #712

bwindels (approved):

PR #717, opened 2022-04-10
Theming - Fix css-url-processor

MidhunSureshR:
This PR makes the following changes:
- Do not run plugin for runtime theme (because url will processed at runtime)
- Cache `cssPath`; because root.source becomes null in some cases (not sure why though)
- Gives each svg a unique name


bwindels (approved):
lgtm

bwindels (approved):

PR #723, opened 2022-04-20
Read config.json on app start

### LINKED ISSUE #609: Pickup config.json changes at app startup rather than build
Currently, the config.json file is only read at build time. To change for example the default home server, it would be nice if people deploying Hydrogen wouldn't have to run the build again.

One option would be to just disable caching for config.json like we do for index.html and read that before bootstrapping the app, but this would break offline and also cause a delay before the UI can be shown.

So we also need a caching strategy in the service worker that always returns what is in cache for config.json, if anything, and in the background updates it from the server if possible.

If caching is correctly disabled on the server, then applying config changes should just be a refresh away.

MidhunSureshR:
For #609;
- Platform ctor optionally accepts a config URL which is read in `init()`
-  Has vite-config changes to produce config.json during build
-  Special caching strategy for config.json

### DIFF
--- a/src/platform/web/Platform.js
+++ b/src/platform/web/Platform.js
@@ -171,7 +171,7 @@ export class Platform {
             if (!this._configURL) {
                 throw new Error("Neither config nor configURL was provided!");
             }
-            const {body}= await this.request(this._configURL, {method: "GET", format: "json"}).response();
+            const {body}= await this.request(this._configURL, {method: "GET", format: "json", cache: true}).response();
             this._config = body;
         }
         this._notificationService = new NotificationService(
--- a/src/platform/web/Platform.js
+++ b/src/platform/web/Platform.js
@@ -126,7 +126,7 @@ function adaptUIOnVisualViewportResize(container) {
 }
 
 export class Platform {
-    constructor(container, assetPaths, config, options = null, cryptoExtras = null) {
+    constructor({ container, assetPaths, config, options = null, cryptoExtras = null }) {
         this._container = container;
         this._assetPaths = assetPaths;
         this._config = config;
--- a/src/platform/web/Platform.js
+++ b/src/platform/web/Platform.js
@@ -126,10 +126,11 @@ function adaptUIOnVisualViewportResize(container) {
 }
 
 export class Platform {
-    constructor({ container, assetPaths, config, options = null, cryptoExtras = null }) {
+    constructor({ container, assetPaths, config, configURL, options = null, cryptoExtras = null }) {
         this._container = container;
         this._assetPaths = assetPaths;
         this._config = config;
+        this._configURL = configURL;
         this.settingsStorage = new SettingsStorage("hydrogen_setting_v1_");
         this.clock = new Clock();
         this.encoding = new Encoding();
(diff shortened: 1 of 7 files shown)


src/platform/web/Platform.js
```diff
@@ -142,7 +143,7 @@ export class Platform {
             this._serviceWorkerHandler = new ServiceWorkerHandler();
             this._serviceWorkerHandler.registerAndStart(assetPaths.serviceWorker);
         }
-        this.notificationService = new NotificationService(this._serviceWorkerHandler, config.push);
+        this.notificationService = null;
```
bwindels: can we use `undefined`? It's easier to type once we switch to typescript.

src/platform/web/Platform.js
```diff
@@ -126,10 +126,11 @@ function adaptUIOnVisualViewportResize(container) {
 }
 
 export class Platform {
-    constructor(container, assetPaths, config, options = null, cryptoExtras = null) {
+    constructor({ container, assetPaths, config, configURL, options = null, cryptoExtras = null }) {
```
bwindels, line 129: You'll need to adjust the LegacyPlatform ctor as well

src/platform/web/sw.js
```diff
@@ -119,6 +126,32 @@ async function handleRequest(request) {
     }
 }
 
+async function handleConfigRequest(request) {
+    const url = new URL(request.url);
+    // rewrite / to /index.html so it hits the cache
+    if (url.origin === baseURL.origin && url.pathname === baseURL.pathname) {
```
bwindels: I don't think you need to rewrite / to index.html for the config?

src/platform/web/sw.js
```diff
@@ -119,6 +126,32 @@ async function handleRequest(request) {
...
+        request = new Request(new URL("index.html", baseURL.href));
+    }
+    let response = await readCache(request);
+    if (response) {
+        fetchAndUpdateConfig(request);
```
bwindels: perhaps you could unify the code paths to make it a bit easier to read:
```
const networkResponsePromise = fetchAdnUpdateConfig(request);
if (response) {
  return response;
} else {
  return await networkResponsePromise;
}
```

vite.config.js
```diff
@@ -14,6 +14,11 @@ export default defineConfig(({mode}) => {
             outDir: "../../../target",
             minify: true,
             sourcemap: true,
+            rollupOptions: {
+                output: {
+                    assetFileNames: (asset) => asset.name.includes("config.json") ? "assets/[name][extname]": "assets/[name].[hash][extname]",
```
bwindels, line 19: we use this option elsewhere as well to not hash the theme manifests, right? do we have to take that into account here or not?

bwindels (approved):
looks good, just a few nits. Did you test the app still works offline?

src/platform/web/sw.js
```diff
@@ -75,7 +75,11 @@ self.addEventListener('fetch', (event) => {
     This has to do with xhr not being supported in service workers.
     */
     if (event.request.method === "GET") {
-        event.respondWith(handleRequest(event.request));
+        if (event.request.url.includes("config.json")) {
```
bwindels: actually, can we check for config.json in handleRequest and call handleConfigRequest from there?

vite.config.js
MidhunSureshR: I haven't created that PR yet. Those changes will be in the theme chooser PR.

MidhunSureshR:
> Did you test the app still works offline?

Yep :+1:

src/platform/web/sw.js
```diff
@@ -96,7 +96,7 @@ let pendingFetchAbortController = new AbortController();
 async function handleRequest(request) {
     try {
         if (request.url.includes("config.json")) {
-            return handleConfigRequest(request);
+            return await handleConfigRequest(request);
```
bwindels: you actually don't need this I think? :shrug:

bwindels (approved):
looks great, this looks good to go!

PR #724, opened 2022-04-25
Implement theme chooser in settings

MidhunSureshR:
- Includes script to remove .tmp (which contains icons) directory after build
- Adds a list of theme manifest locations in config.json
- Adds the names of the default themes to config.json
- Implements a theme-chooser

### DIFF
--- a/scripts/build-plugins/rollup-plugin-build-themes.js
+++ b/scripts/build-plugins/rollup-plugin-build-themes.js
@@ -31,12 +31,12 @@ function appendVariablesToCSS(variables, cssSource) {
     return cssSource + getRootSectionWithVariables(variables);
 }
 
-function addThemesToConfig(bundle, themeSummary, defaultThemes) {
+function addThemesToConfig(bundle, manifestLocations, defaultThemes) {
     for (const [fileName, info] of Object.entries(bundle)) {
         if (fileName === "assets/config.json") {
             const source = new TextDecoder().decode(info.source);
             const config = JSON.parse(source);
-            config["themes"] = themeSummary;
+            config["themeManifests"] = manifestLocations;
             config["defaultTheme"] = defaultThemes;
             info.source = new TextEncoder().encode(JSON.stringify(config));
         }
@@ -247,13 +247,17 @@ module.exports = function buildThemes(options) {
 
         generateBundle(_, bundle) {
             const { assetMap, chunkMap, runtimeThemeChunk } = parseBundle(bundle);
-            const themeSummary = {};
+            const manifestLocations = [];
             for (const [location, chunkArray] of chunkMap) {
                 const manifest = require(`${location}/manifest.json`);
                 const compiledVariables = options.compiledVariables.get(location);
                 const derivedVariables = compiledVariables["derived-variables"];
                 const icon = compiledVariables["icon"];
                 const builtAsset = {};
+                /**
+                 * Generate a mapping from theme name to asset hashed location of said theme in build output.
+                 * This can be used to enumerate themes during runtime.
+                 */
                 for (const chunk of chunkArray) {
                     const [, name, variant] = chunk.fileName.match(/theme-(.+)-(.+)\.css/);
                     builtAsset[`${name}-${variant}`] = assetMap.get(chunk.fileName).fileName;
@@ -265,24 +269,14 @@ module.exports = function buildThemes(options) {
                     "icon": icon
                 };
                 const name = `theme-${manifest.name}.json`;
+                manifestLocations.push(`assets/${name}`);
                 this.emitFile({
                     type: "asset",
                     name,
                     source: JSON.stringify(manifest),
                 });
             }
-            /**
-             * Generate a mapping from theme name to asset hashed location of said theme in build output.
-             * This can be used to enumerate themes during runtime.
-             */
-            for (const [, chunkArray] of chunkMap) {
-                chunkArray.forEach((chunk) => {
-                    const [, name, variant] = chunk.fileName.match(/theme-(.+)-(.+)\.css/);
-                    const assetHashedFileName = assetMap.get(chunk.fileName).fileName;
-                    themeSummary[`${name}-${variant}`] = assetHashedFileName;
-                });
-            }
-            addThemesToConfig(bundle, themeSummary, defaultThemes);
+            addThemesToConfig(bundle, manifestLocations, defaultThemes);
         },
     }
 }
--- a/scripts/build-plugins/rollup-plugin-build-themes.js
+++ b/scripts/build-plugins/rollup-plugin-build-themes.js
@@ -253,8 +253,13 @@ module.exports = function buildThemes(options) {
                 const compiledVariables = options.compiledVariables.get(location);
                 const derivedVariables = compiledVariables["derived-variables"];
                 const icon = compiledVariables["icon"];
+                const builtAsset = {};
+                for (const chunk of chunkArray) {
+                    const [, name, variant] = chunk.fileName.match(/theme-(.+)-(.+)\.css/);
+                    builtAsset[`${name}-${variant}`] = assetMap.get(chunk.fileName).fileName;
+                }
                 manifest.source = {
-                    "built-asset": chunkArray.map(chunk => assetMap.get(chunk.fileName).fileName),
+                    "built-asset": builtAsset,
                     "runtime-asset": assetMap.get(runtimeThemeChunk.fileName).fileName,
                     "derived-variables": derivedVariables,
                     "icon": icon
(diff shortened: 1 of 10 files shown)


src/platform/web/Platform.js
```diff
@@ -174,10 +174,18 @@ export class Platform {
             const {body}= await this.request(this._configURL, {method: "GET", format: "json", cache: true}).response();
             this._config = body;
         }
-        this._notificationService = new NotificationService(
+        this.notificationService = new NotificationService(
```
MidhunSureshR: This was a bug from the previous PR which implemented reading config on app start.

bwindels: Noticed this too yesterday, have fixed it on master already :)

src/platform/web/Platform.js
```diff
@@ -307,6 +315,48 @@ export class Platform {
         return DEFINE_VERSION;
     }
 
+    get themes() {
```
bwindels: perhaps you could put all of the theme related code into a separate class that is accessible on `platform.theming` or something to prevent Platform from growing too big?

src/platform/web/ui/session/settings/SettingsView.js
```diff
@@ -93,10 +93,13 @@ export class SettingsView extends TemplateView {
                 ]);
             })
         );
-
+        
         settingNodes.push(
             t.h3("Preferences"),
             row(t, vm.i18n`Scale down images when sending`, this._imageCompressionRange(t, vm)),
+            t.map(vm => vm.activeTheme, (theme, t) => {
```
bwindels: nit: you could use `t.if` here

scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -249,6 +266,23 @@ module.exports = function buildThemes(options) {
...
+             */
+            for (const [, chunkArray] of chunkMap) {
+                chunkArray.forEach((chunk) => {
+                    const [, name, variant] = chunk.fileName.match(/theme-(.+)-(.+)\.css/);
+                    const assetHashedFileName = assetMap.get(chunk.fileName).fileName;
+                    themeSummary[`${name}-${variant}`] = assetHashedFileName;
+                });
+            }
+            addThemesToConfig(bundle, themeSummary, defaultThemes);
+            this.emitFile({
```
bwindels: what is this file used for?

src/platform/web/Platform.js
```diff
@@ -307,6 +315,48 @@ export class Platform {
...
+        }
+        return undefined;
+    }
+
+    setTheme(themeName) {
+        const themeLocation = this.config["themes"][themeName];
+        if (!themeLocation) {
+            throw new Error(`Cannot find theme location for theme "${themeName}"!`);
+        }
+        this._replaceStylesheet(themeLocation);
```
bwindels: as mentioned in the standup, to make the config file forwards-compatible, we should perhaps already load the css file location from the theme manifests rather than have the css location straight in the config file. As you said, we'll need this anyway for runtime themes.

scripts/cleanup.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+# Remove icons created in .tmp
+rm -rf .tmp
```
bwindels, line 3: Is this removing .tmp files all over the repository? If so, can we limit this to the `target` directory?

src/platform/web/Platform.js
```diff
@@ -178,6 +180,9 @@ export class Platform {
             this._serviceWorkerHandler,
             this._config.push
         );
+        const manifests = this.config["themeManifests"];
+        await this._themeLoader?.init(manifests);
+        await this._themeLoader?.loadThemeFromSetting();
```
bwindels: Why not just loadThemeFromSettings(await themeLoader.getActiveTheme())? That will also apply the default theme if you don't have a preference set yet.

src/platform/web/Platform.js
```diff
@@ -307,6 +312,32 @@ export class Platform {
         return DEFINE_VERSION;
     }
 
+    get themeLoader() {
+        return this._themeLoader;
+    }
+
+    get preferredColorScheme() {
+        if (window.matchMedia("(prefers-color-scheme: dark)")) {
```
bwindels:
```suggestion
        if (window.matchMedia("(prefers-color-scheme: dark)").matches) {
```

src/platform/web/Platform.js
```diff
@@ -307,6 +312,32 @@ export class Platform {
...
+            return COLOR_SCHEME_PREFERENCE.DARK;
+        } else if (window.matchMedia("(prefers-color-scheme: light)")) {
```
bwindels:
```suggestion
        } else if (window.matchMedia("(prefers-color-scheme: light)").matches) {
```

src/platform/web/Platform.js
```diff
@@ -307,6 +312,32 @@ export class Platform {
         return DEFINE_VERSION;
     }
 
+    get themeLoader() {
+        return this._themeLoader;
+    }
+
+    get preferredColorScheme() {
```
bwindels: Can we move this to the ThemeLoader? It's crowded here already, and this seems more or less related to theme loading ...

src/platform/web/ThemeLoader.ts
```diff
@@ -0,0 +1,78 @@
...
+    async init(manifestLocations: string[]): Promise<void> {
+        for (const manifestLocation of manifestLocations) {
+            const { body } = await this._platform
+                .request(manifestLocation, {
+                    method: "GET",
+                    format: "json",
+                    cache: true,
+                })
+                .response();
+            Object.assign(this._themeMapping, body["source"]["built-asset"]);
```
bwindels: This takes some time to understand if you don't know off the top of your head what `build-asset` (also, shouldn't it be `built-assets`, plural?) contains... Could you write it out a bit or add a comment?

src/platform/web/ui/session/settings/SettingsView.js
```diff
@@ -93,10 +93,13 @@ export class SettingsView extends TemplateView {
                 ]);
             })
         );
-
+        
         settingNodes.push(
             t.h3("Preferences"),
             row(t, vm.i18n`Scale down images when sending`, this._imageCompressionRange(t, vm)),
+            t.if(vm => vm.activeTheme, (t, vm) => {
+                return !import.meta.env.DEV? row(t, vm.i18n`Use the following theme`, this._themeOptions(t, vm)): null;
```
bwindels: You could put the `!import.meta.env.DEV` in the binding, e.g. `!import.meta.env.DEV && vm.activeTheme`, so the whole thing does not render on dev mode

vite.config.js
```diff
@@ -16,25 +16,33 @@ export default defineConfig(({mode}) => {
             sourcemap: true,
             rollupOptions: {
                 output: {
-                    assetFileNames: (asset) => asset.name.includes("config.json") ? "assets/[name][extname]": "assets/[name].[hash][extname]",
+                    assetFileNames: (asset) =>
+                        asset.name.includes("config.json") ||
```
bwindels, line 20: Is there any way we can put the config file in the root directory rather than in `assets/`? Makes it more obvious for people what to edit when looking for the configuration.

src/platform/web/Platform.js
MidhunSureshR: It uses `window.` which is platform dependent?

scripts/cleanup.sh
MidhunSureshR: It's removing the .tmp directory in root

bwindels: Right, can you remind me where in the build this is created? It's not a big deal, would be kind of nice that the build remains completely contained in `target` ...

MidhunSureshR: so the reason that .tmp is in the root as opposed to `target` is that the directory is needed at dev time as well.
It is created via a postcss plugin.

src/platform/web/Platform.js
MidhunSureshR: did you mean `themeLoader.setTheme(await themeLoader.getActiveTheme())`?

src/platform/web/ThemeLoader.ts
MidhunSureshR: > also, shouldn't it be built-assets, plural?

I can change this in another PR.
edit: easier to include in this PR

vite.config.js
MidhunSureshR: I'll do this in a follow up PR ~~(it needs the script that runs after build which is introduced in this PR)~~

MidhunSureshR: done in https://github.com/vector-im/hydrogen-web/pull/735

bwindels (approved):

PR #734, opened 2022-05-11
Fix removing members in handleCallMemberEvent

robertlong:
`memberToCallIds` was keyed on `userId` instead of `roomId + userId` which is needed because it's keeping track of previous room member state events. This should fix the bugs I was seeing in Third Room where users can't see eachother after one refreshes.


bwindels (approved):
Thanks for looking into this! Fix looks good, I'll merge it!

PR #736, opened 2022-05-13
Fix room type with federation allowed

ajbura:
Earlier setting room type only work with federation disabled.


robertlong:
Ah, yeah. Missed that. Thanks!

Also @bwindels my plan is to have a series of PRs ready for you when you come back so we can get rid of this branch.

PR #737, opened 2022-05-16
Add room type in summary

ajbura:
Added room type and a method to get state event from indexedDB.

### DIFF
--- a/scripts/sdk/base-manifest.json
+++ b/scripts/sdk/base-manifest.json
@@ -1,7 +1,7 @@
 {
     "name": "@thirdroom/hydrogen-view-sdk",
     "description": "Embeddable matrix client library, including view components",
-    "version": "0.0.6",
+    "version": "0.0.8",
     "main": "./hydrogen.es.js",
     "type": "module"
 }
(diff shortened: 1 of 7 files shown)


robertlong:
Looks great 👍

bwindels:
Didn't have time to review this before it got merged, but I would use `observeStateType`
& `observeStateTypeAndKey` on room instead of adding a getStateEvent method.

robertlong:
Ah, I thought you had reviewed it based on the comments above. @ajbura can you make that change?

ajbura:
@bwindels , `getStateEvent` is for different purpose. When we use `getStateEvent` we just want to read the current state (one time) without caring about the future changes to that state. Whereas `observeStateType` & `observeStateTypeAndKey` do care that we observe accessed state (subscribe it) for future changes. We do can use it to get the `stateObserver` and read the current state from it, but we need to subscribe it or else we will end up memory leak because [remove state observer from room function](https://github.com/vector-im/hydrogen-web/blob/thirdroom/dev/src/matrix/room/BaseRoom.js#L87,L89) will never get called.

ajbura:
@robertlong , That was [diff](https://github.com/vector-im/hydrogen-web/pull/774) PR.

PR #742, opened 2022-05-23
Theme chooser improvements

MidhunSureshR:
- [x] Fix typescript types
- [x] Theme list should show dark and light default variant for each theme as radio-buttons. Other variants should be separately enumerated in the dropdown list.
- [x] Theme list should show default OS choice as option

### DIFF
--- a/scripts/build-plugins/rollup-plugin-build-themes.js
+++ b/scripts/build-plugins/rollup-plugin-build-themes.js
@@ -254,50 +254,9 @@ module.exports = function buildThemes(options) {
                 const derivedVariables = compiledVariables["derived-variables"];
                 const icon = compiledVariables["icon"];
                 const builtAssets = {};
-                let defaultDarkVariant = {}, defaultLightVariant = {};
-                const themeName = manifest.name;
                 for (const chunk of chunkArray) {
                     const [, name, variant] = chunk.fileName.match(/theme-(.+)-(.+)\.css/);
-                    const {name: variantName, default: isDefault, dark} = manifest.values.variants[variant];
-                    const themeId = `${name}-${variant}`;
-                    const themeDisplayName = `${themeName} ${variantName}`;
-                    if (isDefault) {
-                        /**
-                         * This is a default variant!
-                         * We'll add these to the builtAssets (separately) keyed with just the
-                         * theme-name (i.e "Element" instead of "Element Dark").
-                         * We need to be able to distinguish them from other variants!
-                         * 
-                         * This allows us to render radio-buttons with "dark" and
-                         * "light" options.
-                        */ 
-                        const defaultVariant = dark ? defaultDarkVariant : defaultLightVariant;
-                        defaultVariant.variantName = variantName;
-                        defaultVariant.id = themeId
-                        defaultVariant.cssLocation = assetMap.get(chunk.fileName).fileName;
-                        continue;
-                    }
-                    // Non-default variants are keyed in builtAssets with "theme_name variant_name"
-                    // eg: "Element Dark"
-                    builtAssets[themeDisplayName] = {
-                        cssLocation: assetMap.get(chunk.fileName).fileName,
-                        id: themeId
-                    };
-                }
-                if (defaultDarkVariant.id && defaultLightVariant.id) {
-                    /**
-                     * As mentioned above, if there's both a default dark and a default light variant,
-                     * add them to builtAsset separately.
-                     */
-                    builtAssets[themeName] = { dark: defaultDarkVariant, light: defaultLightVariant };
-                }
-                else {
-                    /**
-                     * If only one default variant is found (i.e only dark default or light default but not both),
-                     * treat it like any other variant.
-                     */
-                    const variant = defaultDarkVariant.id ? defaultDarkVariant : defaultLightVariant;
-                    builtAssets[`${themeName} ${variant.variantName}`] = { id: variant.id, cssLocation: variant.cssLocation };
+                    builtAssets[`${name}-${variant}`] = assetMap.get(chunk.fileName).fileName;
                 }
                 manifest.source = {
                     "built-assets": builtAssets,
--- a/scripts/build-plugins/rollup-plugin-build-themes.js
+++ b/scripts/build-plugins/rollup-plugin-build-themes.js
@@ -264,11 +264,11 @@ module.exports = function buildThemes(options) {
                     if (isDefault) {
                         /**
                          * This is a default variant!
-                         * We'll add these to the builtAssets keyed with just the
-                         * theme-name (i.e "Element" instead of "Element Dark")
-                         * so that we can distinguish them from other variants!
+                         * We'll add these to the builtAssets (separately) keyed with just the
+                         * theme-name (i.e "Element" instead of "Element Dark").
+                         * We need to be able to distinguish them from other variants!
                          * 
-                         * This allows us to render a radio-button with "dark" and
+                         * This allows us to render radio-buttons with "dark" and
                          * "light" options.
                         */ 
                         const defaultVariant = dark ? defaultDarkVariant : defaultLightVariant;
--- a/scripts/build-plugins/rollup-plugin-build-themes.js
+++ b/scripts/build-plugins/rollup-plugin-build-themes.js
@@ -246,6 +246,9 @@ module.exports = function buildThemes(options) {
 },
 
         generateBundle(_, bundle) {
+            // assetMap: Mapping from asset-name (eg: element-dark.css) to AssetInfo
+            // chunkMap: Mapping from theme-location (eg: hydrogen-web/src/.../css/themes/element) to a list of ChunkInfo 
+            // types of AssetInfo and ChunkInfo can be found at https://rollupjs.org/guide/en/#generatebundle
             const { assetMap, chunkMap, runtimeThemeChunk } = parseBundle(bundle);
             const manifestLocations = [];
             for (const [location, chunkArray] of chunkMap) {
--- a/scripts/build-plugins/rollup-plugin-build-themes.js
+++ b/scripts/build-plugins/rollup-plugin-build-themes.js
@@ -272,7 +272,7 @@ module.exports = function buildThemes(options) {
                          * "light" options.
                         */ 
                         const defaultVariant = dark ? defaultDarkVariant : defaultLightVariant;
-                        defaultVariant.themeDisplayName = variantName;
+                        defaultVariant.variantName = variantName;
                         defaultVariant.id = themeId
                         defaultVariant.cssLocation = assetMap.get(chunk.fileName).fileName;
                         continue;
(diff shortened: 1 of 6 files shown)


src/platform/web/ui/css/themes/element/manifest.json
```diff
@@ -9,37 +10,51 @@
...
+                    "background-color-secondary": "#2D3239",
+                    "text-color": "#fff",
+                    "accent-color": "#03B381",
+                    "error-color": "#FF4B55",
+                    "fixed-white": "#fff",
+                    "room-badge": "#61708b",
+                    "link-color": "#238cf5"
+                }
+            }
+        }
```
MidhunSureshR: This is added temporarily so that this PR can be tested!
Will remove before merge.

MidhunSureshR:
## Test Cases
- On fresh launch of Hydrogen (no persistent data yet):
    -  the default theme as specified in config should be loaded
    - theme-chooser dropdown in settings should show "Default" as being selected
    - In this option, changing the default theme in config should change the theme on Hydrogen
- Theme names should be readable i.e "Element Dark" as opposed to "element-dark"
- If a given theme contains both a dark and a light default variant, the UI should render these defaults separately with radio-buttons. Other variants should be selectable from the dropdown.
    -  If a default variant is the currently set theme, when settings is opened it should show the same theme as being selected ( i.e radio-buttons with current scheme selected)
- If only one default variant is available, treat it like any other variant

scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -254,13 +254,50 @@ module.exports = function buildThemes(options) {
...
+                        */ 
+                        const defaultVariant = dark ? defaultDarkVariant : defaultLightVariant;
+                        defaultVariant.themeDisplayName = variantName;
+                        defaultVariant.id = themeId
+                        defaultVariant.cssLocation = assetMap.get(chunk.fileName).fileName;
+                        continue;
+                    }
+                    // Non-default variants are keyed in builtAssets with "theme_name variant_name"
+                    // eg: "Element Dark"
+                    builtAssets[themeDisplayName] = {
```
bwindels: Why not just map the variant id (which we know is unique, as it's used as a key) to the css file here? Are you perhaps trying to have a structure in the manifest that maps 1:1 to how we want to present themes in the settings UI? I would go for a format without any duplication (and mangle the data to present in the settings UI at runtime), so for `built-assets` instead of this: 
```json
{
            "Element Violet":
            {
                "cssLocation": "assets/theme-element-violet.d45003f1.css",
                "id": "element-violet"
            },
            "Element":
            {
                "dark":
                {
                    "variantName": "Dark",
                    "id": "element-dark",
                    "cssLocation": "assets/theme-element-dark.2f1cd6cb.css"
                },
                "light":
                {
                    "variantName": "Light",
                    "id": "element-light",
                    "cssLocation": "assets/theme-element-light.a49cf48a.css"
                }
            }
        }
```
..., just this?
```json
{
    "violet": "assets/theme-element-violet.d45003f1.css",
    "dark": "assets/theme-element-dark.2f1cd6cb.css",
    "light": "assets/theme-element-light.a49cf48a.css"
}

File: scripts/build-plugins/rollup-plugin-build-themes.js [hunk 3]
```diff
@@ -254,13 +254,50 @@ module.exports = function buildThemes(options) {
                 const derivedVariables = compiledVariables["derived-variables"];
                 const icon = compiledVariables["icon"];
                 const builtAssets = {};
-                /**
-                 * Generate a mapping from theme name to asset hashed location of said theme in build output.
-                 * This can be used to enumerate themes during runtime.
-                 */
+                let defaultDarkVariant = {}, defaultLightVariant = {};
```
bwindels: I know this is from a previous PR, but I'm struggling a bit to read this code because it's not clear what assetMap & chunkMap contain, and what parseBundle does specifically. Perhaps some more detailed naming and/or comments could help here.

bwindels (commented):
Thanks, this seems to work well enough, but I'd go for a simpler manifest format (as commented) and a slightly different UI (as discussed in the room):
 - given all the variants of all the themes configured; the dropdown contains (this is already the case I think?):
    - light and dark variants are grouped in one item just stating the theme name, and when selected radio buttons are shown (see below) to select the variant.
    - any other variants each have their own item with `${themeName} ${variantName}`, e.g. 
`Element Violet`.
    - A "Default" option to not set any preference and just take the default theme configured in the config file, matching the light/dark settings of the system theme.
 - the config file contains only a default theme name, not default variants for light and dark
 - the radio buttons allow you to pick between "Match system theme", "light" and "dark", like in the Element settings.

Also, we should probably take the habit of increasing the theme manifest version number when making incompatible changes and reject earlier versions at runtime.

src/platform/web/Platform.js
```diff
@@ -188,8 +188,10 @@ export class Platform {
                     this._config.push
                 );
                 const manifests = this.config["themeManifests"];
-                await this._themeLoader?.init(manifests);
-                this._themeLoader?.setTheme(await this._themeLoader.getActiveTheme(), log);
+                await this._themeLoader?.init(manifests, log);
+                const { themeName, themeVariant } = await this._themeLoader.getActiveTheme();
```
bwindels: Currently this prevents the app to load in develop mode (both the lack of a ? before getActiveTheme and the destructuring of undefined when the ? is added). Probably best to wrap these 5 theme related lines in an if (this._themeLoader) ?

src/platform/web/ui/session/settings/SettingsView.js
```diff
@@ -140,11 +140,52 @@ export class SettingsView extends TemplateView {
...
+                hidden: () => {
+                    const themeName = select.options[select.selectedIndex].value;
+                    return "id" in vm.themeMapping[themeName];
+                }
+            },
+            onChange: (e) => radioButtonCallback(e.target.value)
+        },
+        [
+            defaultRadioButton,
+            t.label({for: "default"}, "Match System Theme"),
```
bwindels: I am not a native speaker, but not sure this should be capitalized?
```suggestion
            t.label({for: "default"}, "Match system theme"),
```

src/platform/web/ThemeLoader.ts
```diff
@@ -17,59 +17,197 @@ limitations under the License.
...
+            if (themeVariant) {
+                this._platform.settingsStorage.setString("theme-variant", themeVariant);
+            }
+            else {
+                this._platform.settingsStorage.remove("theme-variant");
+            }
+        });
+    }
+
+    get themeMapping(): Record<string, ThemeInformation> {
```
bwindels, line 160: Perhaps add a comment what the string contains:
```suggestion
    /** Maps theme display name to theme information */
    get themeMapping(): Record<string, ThemeInformation> {
```

src/platform/web/ThemeLoader.ts
```diff
@@ -17,59 +17,197 @@ limitations under the License.
...
-                .request(manifestLocation, {
-                    method: "GET",
-                    format: "json",
-                    cache: true,
-                })
-                .response();
+    async init(manifestLocations: string[], log?: ILogItem): Promise<void> {
+        await this._platform.logger.wrapOrRun(log, "ThemeLoader.init", async (log) => {
+            this._themeMapping = {};
+            for (const manifestLocation of manifestLocations) {
```
bwindels: Consider doing this in parallel using `await Promise.all(manifestLocations.map(async manifestLocation => {...}))` rather than a for loop, to minimize the time needed to initialize the platform.

src/platform/web/ThemeLoader.ts
```diff
@@ -17,59 +17,197 @@ limitations under the License.
...
+        });
+    }
+
+    private _populateThemeMap(manifest, log: ILogItem) {
+        log.wrap("populateThemeMap", (l) => {
             /*
             After build has finished, the source section of each theme manifest
-            contains `built-assets` which is a mapping from the theme-name to the
-            location of the css file in build.
+            contains `built-assets` which is a mapping from the theme-id to
```
bwindels, line 72: Trying to understand why this isn't just the variant id, like used as a key in the `variants` section of the manifest... AFAICT:

1. In the build plugin, you take whatever comes after `theme-` of the css file for a theme as the theme id.
2. Add `built-assets` to the theme manifest also during the build, using `${theme id}-${variant}` as a key for the assets.
3. When preparing to enumerate the themes in the settings, remove the theme id from the key of built-assets to be able to look up the variant name.
4. In `config.json` use the full built-assets key name to reference to a variant for setting a default theme (e.g. element-light).

I think taking the theme id from the theme css filename is a bit implicit. I personally find this clearer:

1. add an `id` field to the theme manifest to reference the theme
2. Use only the variant id (as used in `values.variants` in the manifest) as a key of `built-assets`.
3. Keep using `${theme id}-${variant-id}` in the config file for the default config, but now look up the theme id from manifest.id

E.g. don't use the theme id within the different files for the same theme.

src/platform/web/ThemeLoader.ts
```diff
@@ -17,59 +17,197 @@ limitations under the License.
...
+            l.log({ l: "Result", themeMapping: this._themeMapping });
+         });
     }
 
-    setTheme(themeName: string, log?: ILogItem) {
-        this._platform.logger.wrapOrRun(log, {l: "change theme", id: themeName}, () => {
-            const themeLocation = this._themeMapping[themeName];
-            if (!themeLocation) {
-                throw new Error( `Cannot find theme location for theme "${themeName}"!`);
+    setTheme(themeName: string, themeVariant?: "light" | "dark" | "default", log?: ILogItem) {
```
bwindels, line 135: Ideally, we'd pass the theme id here, rather than the display name

scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -254,10 +257,6 @@ module.exports = function buildThemes(options) {
                 const derivedVariables = compiledVariables["derived-variables"];
                 const icon = compiledVariables["icon"];
                 const builtAssets = {};
-                /**
-                 * Generate a mapping from theme name to asset hashed location of said theme in build output.
-                 * This can be used to enumerate themes during runtime.
-                 */
                 for (const chunk of chunkArray) {
                     const [, name, variant] = chunk.fileName.match(/theme-(.+)-(.+)\.css/);
                     builtAssets[`${name}-${variant}`] = assetMap.get(chunk.fileName).fileName;
```
bwindels, line 262: Not modified in this PR, but I was thinking if this path shouldn't be relative to the theme manifest ...
Right now it is relative to the build root, (which can be deployed on the document root or anywhere deeper on a server).

When you have the config file refer to theme manifests on a different server, it's unclear how the absolute URL of the css files should be resolved if these paths are not relative to the theme manifest, as you can't really assume that there is another hydrogen deployment in the parent directory with an assets directory on the other server.

bwindels (commented):
Working well, but some more thoughts, especially around the format of the manifest.

src/platform/web/ThemeLoader.ts
MidhunSureshR: > In the build plugin, you take whatever comes after theme- of the css file for a theme as the theme id.

Just to make sure that we're on the same page, theme-id is of the form "`theme`-`variant`" where `theme` is the key provided [here](https://github.com/vector-im/hydrogen-web/blob/master/vite.config.js#L37) and the `variant` is the key provided [here](https://github.com/vector-im/hydrogen-web/blob/master/src/platform/web/ui/css/themes/element/manifest.json#L27).
i.e the relation is inverse in that the filename contains the id.

But I agree that the code currently gives no indication that this is the case.

scripts/build-plugins/rollup-plugin-build-themes.js
MidhunSureshR: I've a list of changes I need to make in the plugin. I'll make this change in another PR.

bwindels (approved):

PR #743, opened 2022-05-26
Make hs api public

ajbura:


robertlong:
Looks good, thanks for pushing this forward!

PR #693, opened 2022-02-26
Make the SDK friendly to locally link and develop on

MadLittleMods:
Make the `hydrogen-view-sdk` friendly to locally link and develop on. Can now simply `yarn watch:sdk` to watch and build on changes and use in CommonJS and ESM environments.

Also adds `yarn test:sdk` to test whether the SDK works in an ESM Vite build and when using `require` in Node.js (CommonJS).

Fix https://github.com/vector-im/hydrogen-web/issues/686
Fix https://github.com/vector-im/hydrogen-web/issues/682
Fix https://github.com/vector-im/hydrogen-web/issues/722

Instead of deleting the whole `target/` directory, leave it alone so the symlink driving the `npm link`/`yarn link` stays in tact.

Leave Vite builds in their build directories (`/lib-build`/`/asset-build`) so you can `vite build --watch` to build on local changes and still have a consisent place to reference in the `package.json` `exports`. Previously, everything relied on `build.sh` which does a bunch of moving and renaming and made it hard to quickly rebuild on changes.

Add back support for CommonJS (adding the `package.json` `exports`).

The last piece is making sure the `?url` imports (ex. `import workerPath from 'hydrogen-view-sdk/main.js?url';`) still work. ~~It looks like this may have just been solved via https://github.com/vitejs/vite/issues/6725 -> https://github.com/vitejs/vite/pull/7073 (literally 2 days ago) and we just need to wait for the next Vite release~~ Was solved by https://github.com/vitejs/vite/pull/7098 and is available in the latest Vite releases 🎉

Also updating `vite` because it includes fixes for https://github.com/vector-im/hydrogen-web/issues/722 which were added in https://github.com/vitejs/vite/issues/7822 ->https://github.com/vitejs/vite/pull/7827

## Dev notes

 - `package.json` and `exports` docs: https://nodejs.org/docs/latest-v17.x/api/packages.html
 - https://vitejs.dev/config/#build-assetsdir

#### Use the latest Vite from GitHub source (before release)

How can I try out the latest Vite without waiting for a release?

See https://github.com/vitejs/vite/blob/main/CONTRIBUTING.md#repo-setup to build locally and `npm link` it

---

How to yarn install a package from a GitHub Monorepo? This doesn't work for Vite as it requires more build stuff.

See

 - https://github.com/yarnpkg/yarn/issues/4725#issuecomment-761528734
 - https://github.com/yarnpkg/yarn/issues/1570

```
[...]

PR #746, opened 2022-05-31
Import SDK assets from the `assets/` directory

MadLittleMods:
Import SDK assets from the `assets/` directory

> Will be easier towards the future when adding more assets. Probably best to keep style.css for now for backwards compat though.
>
> *-- @bwindels, https://github.com/vector-im/hydrogen-web/pull/693#discussion_r853844282*

---

```sh
$ yarn test:sdk
...
SDK works in Vite build ✅
SDK works in CommonJS ✅
```

### DIFF
--- a/scripts/sdk/base-manifest.json
+++ b/scripts/sdk/base-manifest.json
@@ -10,8 +10,6 @@
         },
         "./paths/vite": "./paths/vite.js",
         "./style.css": "./asset-build/assets/theme-element-light.css",
-        "./theme-element-light.css": "./asset-build/assets/theme-element-light.css",
-        "./theme-element-dark.css": "./asset-build/assets/theme-element-dark.css",
         "./main.js": "./asset-build/assets/main.js",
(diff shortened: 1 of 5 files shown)


scripts/sdk/base-manifest.json
```diff
@@ -10,8 +10,6 @@
         },
         "./paths/vite": "./paths/vite.js",
         "./style.css": "./asset-build/assets/theme-element-light.css",
-        "./theme-element-light.css": "./asset-build/assets/theme-element-light.css",
```
bwindels: Could we keep these for backwards compatibility? Do keep the docs updated though.

bwindels (approved):
lgtm, thanks!

scripts/sdk/base-manifest.json
MadLittleMods: Added them back ⏩

MadLittleMods:
Thanks for the review and merge @bwindels 🐒

PR #749, opened 2022-06-08
More simple lightbox setup for SDK and compatible with custom `History`

MadLittleMods:
More simple lightbox setup for SDK and compatible with custom `History`

Changes to go along with https://github.com/matrix-org/matrix-public-archive/pull/12

## Todo

 - [x] Test this out in the real Hydrogen to make sure everything still works


src/domain/session/SessionViewModel.js
```diff
@@ -81,12 +81,12 @@ export class SessionViewModel extends ViewModel {
...
-            this._updateLightbox(eventId);
-        }));
-        this._updateLightbox(lightbox.get());
-
+        setupLightboxNavigation(this, 'lightboxViewModel', (eventId) => {
+            return {
+                room,
+                eventId,
+            };
+        });
```
MadLittleMods, line 89: Replacing all of the previous boilerplate to setup the lightbox with this composable function that external consumers can also use in the SDK.

This makes setting up the lightbox in your project that uses the SDK as simple as a `setupLightboxNavigation(...)` function call. See the usage in https://github.com/matrix-org/matrix-public-archive/pull/12

src/domain/session/room/LightboxViewModel.js
```diff
@@ -92,6 +96,6 @@ export class LightboxViewModel extends ViewModel {
     }
 
     close() {
-        this.platform.history.pushUrl(this.closeUrl);
+        this.history.pushUrl(this.closeUrl);
```
MadLittleMods, line 99: Since we're using a custom `History` implementation in https://github.com/matrix-org/matrix-public-archive/pull/12 to get hashes relative to the room, we need to use that custom `History` over the default one from `platform`.

The only other place we use `platform.history` is in [`src/platform/web/main.js#L38`](https://github.com/vector-im/hydrogen-web/blob/8b2299852eea7a2bc0e247de6ccdb2676b3e4665/src/platform/web/main.js#L38)

src/domain/ViewModel.ts
```diff
@@ -142,4 +144,8 @@ export class ViewModel<O extends Options = Options> extends EventEmitter<{change
     get navigation(): Navigation {
         return this._options.navigation;
     }
+
+    get history(): History {
+        return this._options.history;
+    }
```
MadLittleMods, line 150: See https://github.com/vector-im/hydrogen-web/pull/749#discussion_r891899396 for why we're doing this now.

src/domain/session/room/LightboxViewModel.js
```diff
@@ -19,21 +19,25 @@ import {ViewModel} from "../../ViewModel";
...
     }
 
     _subscribeToEvent(room, eventId) {
-        const eventObservable = room.observeEvent(eventId);
-        this.track(eventObservable.subscribe(eventEntry => {
-            this._loadEvent(room, eventEntry);
-        }));
-        this._loadEvent(room, eventObservable.get());
+        let event = this._eventEntry;
+        if (!this._eventEntry) {
```
MadLittleMods: Added the option of passing in `eventEntry` directly since we have access to all of the events already in [`matrix-public-archive`](https://github.com/matrix-org/matrix-public-archive/pull/12) and we don't need or want to load more events.

Previously, this only allowed passing `eventId` which loaded in the full event.

src/platform/web/dom/History.js
```diff
@@ -30,10 +30,10 @@ export class History extends BaseObservableValue {
         But for SSO, we need to handle <root>/?loginToken=<TOKEN>
         Handle that as a special case for now.
         */
-        if (document.location.search.includes("loginToken")) {
+        if (document?.location?.search.includes("loginToken")) {
             return document.location.search;
         }
-        return document.location.hash;
+        return document?.location?.hash;
```
MadLittleMods, line 36: This is to allow this function to run when server-side rendered in `linkedom`. It's ok for this function to return `undefined`. See [`shared/lib/archive-history.js`](https://github.com/matrix-org/matrix-public-archive/pull/12/files#diff-511a3e74302f787f8e9c5b9a97c92ecee3ef3706a917b3134a2c5b5c5033b561R1) for how this is integrated.

If this isn't desired, I can always wrap `super.get()` in my custom `History` instance and default that way.

src/domain/session/room/LightboxViewModel.js
bwindels: I think it would be better to inject your custom history through `platform.history`.

src/domain/session/room/LightboxViewModel.js
bwindels: Let's make it clear that this `if` needs `event` to be defined as it assigns it.
```suggestion
        if (!event) {
```

src/platform/web/dom/History.js
bwindels: I think it'd be better if you add your custom logic in your custom instance. It's going to be hard to not have regressions for some arcane DOM impl we never test for.

src/domain/session/room/lightbox-navigation.js
```diff
@@ -0,0 +1,69 @@
...
+// it. This acts like a private field on the class since no one else has the
+// symbol to look it up.
+let lightboxViewModelSymbol = Symbol('lightboxViewModel');
+
+/**
+ * Destroys and creates a new the `LightboxViewModel` depending if
+ * `lightboxChildOptions.eventEntry` or `lightboxChildOptions.eventId` are
+ * provided.
+ */
+function updateLightboxViewModel(vm, fieldName, lightboxChildOptions) {
```
bwindels, line 30: Hmmm, I think this kind of dynamic mixin programming makes the code as a whole a lot less readable. Also, this will prevent us to make the `track` and `disposeTracked` method protected once we're fully converted to typescript. 

I can see a case for having a generic helper method on ViewModel to setup a child view model based on a navigation observable, but I'd do it differently, something like this:
```js
class ViewModel {
	bindChildToNavigationSegment(segmentName, setter, getter, vmMapper, publicPropName) {
		const update = value => {
			if (getter()) {
	            setter(this.disposeTracked(getter()));
	        }
	        if (value) {
	            setter(this.track(vmMapper(value)));
	        }
	        this.emitChange(publicPropName);
		}

		const observable = this.navigation.observe(segmentName);
        this.track(observable.subscribe(update));
        update(observable.get());
	}
}
```

MadLittleMods: To make it make sense it my head, here would be the usage for `lightbox` (adapted to using object parameters since there is so many flying around here):

```js
this.bindChildToNavigationSegment({
  segmentName: 'lightbox',
  setter: (newValue) => {
    this._lightboxViewModel = newValue;
  },
  getter: () = {
    return this._lightboxViewModel;
  },
  vmMapper: (eventId) = {
    return new LightboxViewModel(this.childOptions({eventId, room}));
  },
  publicPropName: 'lightboxViewModel'
});
```


`bindChildToNavigationSegment` does make a lot of sense to capture this abstraction because it's so common 👍

The usage seems like too much boilerplate for what a SDK consumer cares about though.


---

I think we could boil it down to:

```js
this.bindChildToNavigationSegment({
  segmentName: 'lightbox',
  vmMapper: (eventId) = {
    return new LightboxViewModel(this.childOptions({eventId, room}));
  },
  publicPropName: 'lightboxViewModel'
});
```

```js
class ViewModel {
  bindChildToNavigationSegment({ segmentName, vmMapper, publicPropName }) {
        // Store the `ViewModel` under a symbol so no one else can tamper with
        // it. This acts like a private field on the class since no one else has the
        // symbol to look it up.
      let viewModelSymbol = Symbol(publicPropName);

      // On the given `vm`, create a getter at `publicPropName` that the
        // `ViewModel` is exposed at for usage in the view.
        Object.defineProperty(vm, publicPropName, {
            get: function() {
                return vm[viewModelSymbol];
            }
        });
        
    const update = value => {
      if (vm[lightboxViewModelSymbol]) {
              vm[lightboxViewModelSymbol] = this.disposeTracked(getter());
          }
          if (value) {
              vm[lightboxViewModelSymbol] = this.track(vmMapper(value));
          }
          this.emitChange(publicPropName);
    }

    const observable = this.navigation.observe(segmentName);
        this.track(observable.subscribe(update));
        update(observable.get());
  }
}
```

If some use case needs `getter`/`setter` we could have those as extra options and store it for them by default as above ^

MadLittleMods: @bwindels Does the boiled down version look reasonable?

PR #752, opened 2022-06-15
Create a type for theme-manifest

MidhunSureshR:


src/platform/types/theme.ts
```diff
@@ -0,0 +1,67 @@
...
+         * initialized) and 'variant' is the key used to specify the variant details in the values
+         * section below.
+         */
+        "built-asset": Record<string, string>;
+        // Location of css file that will be used for themes derived from this theme.
+        "runtime-asset": string;
+        // Array of derived-variables
+        "derived-variables": Array<string>;
+    };
+    values: {
```
bwindels, line 46: Should all keys in the manifest be optional so we always are pushed to validate the contents by the type system?

bwindels (approved):
looks good, just one question

src/platform/types/theme.ts
MidhunSureshR: Sounds like a good idea! We can always revert if it's too annoying. ([34eac94](https://github.com/vector-im/hydrogen-web/pull/752/commits/34eac94da3d2c5cf4c700f76ab0dc294b61bd02a))

bwindels (approved):

PR #756, opened 2022-06-16
Improve code quality in css-url-variables plugin

### LINKED ISSUE #745: Scope variables in Custom PostCSS plugins to each file processed
*As discovered in https://github.com/vector-im/hydrogen-web/issues/722 and suggested we should that we should still put some effort into fixing:*

> [...] could you still have a look to move the scope of those variables down? I think it's a good thing to do regardless of it not being an immediate problem anymore now.
>
> *-- @bwindels, https://github.com/vector-im/hydrogen-web/issues/722#issuecomment-1127534936*

---

Scope variables in Custom PostCSS plugins to each file processed instead of in the global plugin instance.

> We should move the map variables inside the file processing (`Once`) scope so they aren't shared across plugin runs. [...] We should probably also pass them around directly to avoid concurrent parallel processing problems.

Example: https://github.com/vector-im/hydrogen-web/blob/480c5c1584c43c0f2d1ad82054e33b62a442a730/scripts/postcss/css-compile-variables.js#L136-L139

Currently, this doesn't cause any issues in the codebase. But this is a bit of technical debt that we should get to at some point.

MidhunSureshR:
- Fixes a typo in `vite.common-config.js` which was preventing the `icons` section in the manifest from being populated.
- Refactor away global variables. This was causing variables to be repeated in the css file.

### DIFF
--- a/scripts/postcss/css-url-to-variables.js
+++ b/scripts/postcss/css-url-to-variables.js
@@ -20,11 +20,9 @@ const valueParser = require("postcss-value-parser");
  * This plugin extracts content inside url() into css variables and adds the variables to the root section.
  * This plugin is used in conjunction with css-url-processor plugin to colorize svg icons.
  */
-let counter;
-let urlVariables;
 const idToPrepend = "icon-url";
 
-function findAndReplaceUrl(decl) {
+function findAndReplaceUrl(decl, urlVariables, counter) {
     const value = decl.value;
     const parsed = valueParser(value);
     parsed.walk(node => {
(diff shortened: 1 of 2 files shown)


bwindels (approved):
Thanks, looking good!

scripts/postcss/css-url-to-variables.js
```diff
@@ -20,11 +20,9 @@ const valueParser = require("postcss-value-parser");
  * This plugin extracts content inside url() into css variables and adds the variables to the root section.
```
MadLittleMods, line 20: Related to https://github.com/vector-im/hydrogen-web/issues/745

PR #760, opened 2022-06-20
Refactor theme builder plugin

MidhunSureshR:
- [x] Do not use theme name in manifest file name; instead use the theme-key (`theme-element.json` vs `theme-Element.json`)
- [x] Runtime theme chunks should also be aggregated into a map (because there will be more than one runtime asset when we have multiple theme collections)
- [x] Locations in `built-asset`section should be relative to manifest file ([see here](https://github.com/vector-im/hydrogen-web/pull/742#discussion_r892253075))
- [x] Split `parseBundle` function into smaller functions
- [x] Theme Collection id ("element") should come from within the manifest and not from the key in `vite.config.js` i.e add an `id` field to manifest.

### DIFF
--- a/src/platform/web/ThemeLoader.ts
+++ b/src/platform/web/ThemeLoader.ts
@@ -76,11 +76,16 @@ export class ThemeLoader {
             const themeName = manifest.name;
             let defaultDarkVariant: any = {}, defaultLightVariant: any = {};
             for (let [themeId, cssLocation] of Object.entries(builtAssets)) {
-                /**
-                 * This cssLocation is relative to the location of the manifest file.
-                 * So we first need to resolve it relative to the root of this hydrogen instance.
-                 */
-                cssLocation = new URL(cssLocation, new URL(manifestLocation, window.location.origin)).href;
+                try {
+                    /**
+                     * This cssLocation is relative to the location of the manifest file.
+                     * So we first need to resolve it relative to the root of this hydrogen instance.
+                     */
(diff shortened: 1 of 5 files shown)


src/platform/web/ThemeLoader.ts
```diff
@@ -75,7 +75,12 @@ export class ThemeLoader {
             const builtAssets: Record<string, string> = manifest.source?.["built-assets"];
             const themeName = manifest.name;
             let defaultDarkVariant: any = {}, defaultLightVariant: any = {};
-            for (const [themeId, cssLocation] of Object.entries(builtAssets)) {
+            for (let [themeId, cssLocation] of Object.entries(builtAssets)) {
+                /**
+                 * This cssLocation is relative to the location of the manifest file.
+                 * So we first need to resolve it relative to the root of this hydrogen instance.
+                 */
+                cssLocation = new URL(cssLocation, new URL(manifestLocation, window.location.origin)).href;
```
bwindels: I think this can throw, we should probably still try any next items in the builtAssets when that happens?

bwindels (approved):
apart from one comment, lgtm!

src/platform/web/ThemeLoader.ts
MidhunSureshR: Fixed by https://github.com/vector-im/hydrogen-web/pull/760/commits/6718198d9cc9e5aac8d9e3c0ca5534440b53bdcd

PR #761, opened 2022-06-24
Require node 15+

Half-Shot:
We use replaceAll in scripts/postcss/svg-colorizer.js which is a ES2021 feature. https://node.green/#ES2021-features--String-prototype-replaceAll.

I ran into this when trying to build on Node 14

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,9 @@
   "directories": {
     "doc": "doc"
   },
+  "enginesStrict": {
+    "node": ">=15"
+  },
   "scripts": {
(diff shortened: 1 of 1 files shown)


bwindels:
Looks good, thanks!

bwindels (approved):

jaller94:
[NodeJS 15 is no longer maintained.](https://nodejs.org/en/about/releases/)
We may as well raise the required version to the closest LTS Node 16, before someone asks for support.

PR #764, opened 2022-06-25
Add date headers in timeline

### LINKED ISSUE #489: Add date separator in timeline
It'll be a bit tricky, but we should really do the proper thing and have date separator be represented by their own tile in TilesCollection, rather than a flag on other tiles (which would be messy to implement in the view).

Tiles could signal that a tile should be inserted when setting the sibling, OR, that the tile itself should be removed. This way, a DateTile can be inserted where we go from one day to another, and the DateTile can remove itself if it stops having siblings of different dates.

bwindels:
Fixes https://github.com/vector-im/hydrogen-web/issues/489
Currently looks like:
image

Missing:
 - [ ] jumps when at top and gap is filled
 - [ ] date separators look too much like announcement views
 - [ ] seems not stay at the bottom anymore when sending new messages
 - [ ] create shared date formatter in tile options
 	- does relative time
 	- use Int.DateFormatter

### DIFF
--- a/src/domain/session/room/timeline/TilesCollection.js
+++ b/src/domain/session/room/timeline/TilesCollection.js
@@ -130,15 +130,11 @@ export class TilesCollection extends BaseObservableList {
 
         const newTile = this._createTile(entry);
         if (newTile) {
-            if (prevTile) {
-                prevTile.updateNextSibling(newTile);
-                // this emits an update while the add hasn't been emitted yet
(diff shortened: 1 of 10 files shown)


bwindels:
Started a second stab at this at https://github.com/vector-im/hydrogen-web/pull/938

bwindels:
Obsoleted by #938, which is merged now.

PR #769, opened 2022-06-28
Support for derived themes

MidhunSureshR:


package.json
```diff
@@ -50,6 +50,7 @@
     "postcss-flexbugs-fixes": "^5.0.2",
     "postcss-value-parser": "^4.2.0",
     "regenerator-runtime": "^0.13.7",
+    "svgo": "^2.8.0",
```
MidhunSureshR, line 53: See https://github.com/svg/svgo; tldr: optimizes svg, removes whitespace etc..

scripts/postcss/css-url-to-variables.js
```diff
@@ -81,7 +81,8 @@ module.exports = (opts = {}) => {
             const urlVariables = new Map();
             const counter = createCounter();
             root.walkDecls(decl => findAndReplaceUrl(decl, urlVariables, counter));
-            if (urlVariables.size) {
+            const cssFileLocation = root.source.input.from;
+            if (urlVariables.size && !cssFileLocation.includes("type=runtime")) {
```
MidhunSureshR, line 85: For runtime themes, we shouldn't add the extracted icon variables to the `:root` section because we'll be adding them at runtime.

scripts/test-theme.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/zsh
```
MidhunSureshR, line 1: This is just a script to make testing the code easier. It copies `theme.json` to `target/assets/theme-customer.json` and adds it as a theme to `config.json`. `jq` needs to be installed for this to work.
`yarn build && ./scripts/test-theme.sh` if you want to the run the PR.

src/platform/types/theme.ts
```diff
@@ -22,6 +22,13 @@ export type ThemeManifest = Partial<{
     version: number;
     // A user-facing string that is the name for this theme-collection.
     name: string;
+    // An identifier for this theme
```
MidhunSureshR, line 25: Adding some fields I missed in the documentation PR.

theme.json
```diff
@@ -0,0 +1,51 @@
+{
```
MidhunSureshR, line 1: This is a test theme; can be removed along with `test-theme.sh`.

MidhunSureshR (commented):
Some clarifications:

scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -43,6 +43,37 @@ function addThemesToConfig(bundle, manifestLocations, defaultThemes) {
...
+ * @param {*} icons Object where keys are css variable names and values are locations of the svg
+ * @param {*} manifestLocation Location of manifest used for resolving path 
+ */
+async function generateIconSourceMap(icons, manifestLocation) {
+    const sources = {};
+    const fs = require("fs").promises;
+    for (const icon of Object.values(icons)) {
+        const [location] = icon.split("?");
+        const resolvedLocation = path.resolve(__dirname, "../../", manifestLocation, location);
+        const iconData = await fs.readFile(resolvedLocation);
```
MidhunSureshR: Will remove the await within loop here

scripts/build-plugins/rollup-plugin-build-themes.js
```diff
@@ -43,6 +43,44 @@ function addThemesToConfig(bundle, manifestLocations, defaultThemes) {
...
+ * @param {*} manifestLocation Location of manifest used for resolving path 
+ */
+async function generateIconSourceMap(icons, manifestLocation) {
+    const sources = {};
+    const fileNames = [];
+    const promises = [];
+    const fs = require("fs").promises;
+    for (const icon of Object.values(icons)) {
+        const [location] = icon.split("?");
+        const resolvedLocation = path.resolve(__dirname, "../../", manifestLocation, location);
```
bwindels: can you add a comment what the ../.. is about?

src/platform/web/theming/shared/color.mjs
```diff
@@ -13,10 +13,10 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+import pkg from 'off-color';
+const offColor = pkg.offColor;
```
bwindels, line 17: are you doing a default import because this is a commonjs package? No way to import only what we need I suppose?

src/platform/web/theming/ThemeLoader.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    async init(manifestLocations: string[], log?: ILogItem): Promise<void> {
+        await this._platform.logger.wrapOrRun(log, "ThemeLoader.init", async (log) => {
+            const results = await Promise.all(
+                manifestLocations.map(location => this._platform.request(location, { method: "GET", format: "json", cache: true, }).response())
+            );
+            const runtimeThemeParser = new RuntimeThemeParser(this._platform, this.preferredColorScheme);
+            const builtThemeParser = new BuiltThemeParser(this.preferredColorScheme);
+            const runtimeThemePromises: Promise<void>[] = [];
+            for (let i = 0; i < results.length; ++i) {
```
bwindels, line 41: Not needed in this PR, but it would be nice to at some point only load the theme we're going to use it (either the one at startup, or the one we're switching to in the settings). I think we already discussed this?

tsconfig.json
```diff
@@ -4,7 +4,7 @@
         "noImplicitAny": false,
         "noEmit": true,
         "target": "ES2020",
-        "module": "ES2020",
+        "module": "Node16",
```
bwindels: I wonder if we can restrict this to the only the directory where we need this? Most of the code can not use the node api ...

bwindels (approved):
Just a few small things, but looks great!

scripts/build-plugins/rollup-plugin-build-themes.js
MidhunSureshR: I just realized that `manifestLocation` is already an absolute path, so we don't need this. Fixed in https://github.com/vector-im/hydrogen-web/commit/708637e390250db305e11bbc2c7b5bf8cc30479e.

src/platform/web/theming/shared/color.mjs
MidhunSureshR: Yep, `yarn test` fails otherwise.

tsconfig.json
MidhunSureshR: I'm trying to think of why I made this change ...

MidhunSureshR: I was getting this [error](https://github.com/vector-im/hydrogen-web/runs/7423907806?check_suite_focus=true) but I no longer get it when I revert this change.

MidhunSureshR: Reverted in https://github.com/vector-im/hydrogen-web/pull/769/commits/39817dc36bb8952aa7ce55045c7ead6f506c776f

PR #770, opened 2022-06-28
Ignore macOS house-keeping metadata `.DS_Store`

MadLittleMods:
Ignore macOS house-keeping metadata `.DS_Store`


MadLittleMods:
Merging as this seems like an uncontroversial small house-keeping thing that every repo has to add at some point.

PR #773, opened 2022-06-29
Add a couple consistent selectors to reference in tests

MadLittleMods:
Add a couple consistent selectors to reference in tests, example: https://github.com/matrix-org/matrix-public-archive/pull/29#discussion_r909492946

Related to https://github.com/vector-im/hydrogen-web/pull/690

Using `data-testid` because it seems generic out of the list from the following Cypress links. I'm not using Cypress but just trying to use something "standard". Feel free to suggest something better.

 - https://docs.cypress.io/guides/core-concepts/cypress-app#Uniqueness
 - https://docs.cypress.io/guides/references/best-practices#How-It-Works

### DIFF
--- a/src/platform/web/ui/avatar.js
+++ b/src/platform/web/ui/avatar.js
@@ -31,7 +31,7 @@ export function renderStaticAvatar(vm, size, extraClasses = undefined) {
         avatarClasses += ` ${extraClasses}`;
     }
     const avatarContent = hasAvatar ? renderImg(vm, size) : text(vm.avatarLetter);
-    const avatar = tag.div({className: avatarClasses}, [avatarContent]);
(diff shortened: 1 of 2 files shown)


bwindels (approved):
Not an enormous fan of having these attributes in the production build, but can't think of anything better either without adding a lot of build-time tooling to remove or add them at build-time, so let's go with this.

I guess we could always add an AST filter to remove "data-testid" from all object literals if this becomes an issue :)

MadLittleMods:
Thanks for the review and merge @bwindels 🐸

I'd be in favor of the dumb solution of keeping them around in all environments. If we just don't want them in the DOM, we could filter them from in the templating but then it's harder to compare/reference your Hydrogen vs the test Hydrogen.

PR #774, opened 2022-06-29
Add more homeserver api

ajbura:
This adds methods on HomeServerApi to search for a user profile and invite a user into a room. It also adds the `type` property to the room creation data and exposes that on BaseRoom.

### DIFF
--- a/src/matrix/room/BaseRoom.js
+++ b/src/matrix/room/BaseRoom.js
@@ -415,7 +415,7 @@ export class BaseRoom extends EventEmitter {
     }
 
     get type() {
-        return this._summary.data.type;
+        return this._summary.data.type ?? undefined;
     }
 
     get lastMessageTimestamp() {
--- a/src/matrix/room/timeline/persistence/GapWriter.js
+++ b/src/matrix/room/timeline/persistence/GapWriter.js
@@ -163,7 +163,7 @@ export class GapWriter {
         if (!Array.isArray(chunk)) {
             throw new Error("Invalid chunk in response");
         }
-        if (typeof end !== "string") {
+        if (typeof end !== "string" && typeof end !== "undefined") {
             throw new Error("Invalid end token in response");
         }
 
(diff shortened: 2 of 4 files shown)


src/matrix/room/BaseRoom.js
```diff
@@ -415,7 +415,7 @@ export class BaseRoom extends EventEmitter {
     }
 
     get type() {
-        return this._summary.data.type;
+        return this._summary.data.type ?? undefined;
```
bwindels, line 418: what's this for?

src/matrix/room/timeline/persistence/GapWriter.js
```diff
@@ -163,7 +163,7 @@ export class GapWriter {
         if (!Array.isArray(chunk)) {
             throw new Error("Invalid chunk in response");
         }
-        if (typeof end !== "string") {
+        if (typeof end !== "string" && typeof end !== "undefined") {
```
bwindels, line 166: This was already added on master, you can just merge it in now.

bwindels (approved):
looks good, just one question about the ?? undefined

src/matrix/room/BaseRoom.js
ajbura: Fallback to `this._summary.data.type` is null when [processingStateEvent](https://github.com/vector-im/hydrogen-web/blob/bebbbd9db027ba8200e1f20be237dc63659de486/src/matrix/room/RoomSummary.js#L85) but it was returning undefined because of [this](https://github.com/vector-im/hydrogen-web/blob/bebbbd9db027ba8200e1f20be237dc63659de486/src/matrix/room/RoomSummary.js#L171) where I am unsure if copy.type is string, undefined or null. so added the ?? undefined to make sure it always return string or undefined at public endpoint.

src/matrix/room/timeline/persistence/GapWriter.js
ajbura: This PR is on `thirdroom/dev` so I was wondering is it safe to merge master?

robertlong:
Going to merge this into thirdroom/dev. We still need to sync up on what's left to merge group calls into main and then get our various changes split up and merged.

PR #777, opened 2022-07-01
TS conversion for `LoginViewModel`

ibeckermayer:


.eslintrc.js
```diff
@@ -12,7 +12,8 @@ module.exports = {
         "no-console": "off",
         "no-empty": "off",
         "no-prototype-builtins": "off",
-        "no-unused-vars": "warn"
+        "no-unused-vars": "warn",
+        "semi": ["error", "always"]
```
ibeckermayer: I noticed this was the de facto standard so I made it de jure.

src/domain/ViewModel.ts
```diff
@@ -115,7 +115,7 @@ export class ViewModel<O extends Options = Options> extends EventEmitter<{change
         return result;
     }
 
-    emitChange(changedProps: any): void {
+    emitChange(changedProps?: any): void {
```
ibeckermayer: I made this change in order to get [this line](https://github.com/vector-im/hydrogen-web) to compile.

ibeckermayer:
Also of note:

Unfortunately when a variable is typed as a Javascript class like [`private _passwordLoginViewModel?: PasswordLoginViewModel;`](https://github.com/vector-im/hydrogen-web/pull/777/files#diff-fdb2dad635bedc87b24b90bdff4fa083c7f69bd7431fe60f0c0e4d1ff92de1d0R37), the typescript compiler isn't well equipped to deal with it and falls back to an `any`:
```
Could not find a declaration file for module './PasswordLoginViewModel.js'. './src/domain/login/PasswordLoginViewModel.js' implicitly has an 'any' type.ts(7016)
```

In the case of these optional types (`?:`), this means that we can use them in a way that would normally cause the compiler to complain about us not checking for `undefined`. Hypothetically I could manually identify and solve for all these now, however I decided that because this code is already battle tested, we can safely ignore this issue for the moment. Once we get around to converting `PasswordLoginViewModel.js` to `PasswordLoginViewModel.ts`, the compiler will begin complaining and we can restructure things then.

src/domain/ViewModel.ts
```diff
@@ -37,7 +37,7 @@ export type Options = {
     emitChange?: (params: any) => void
 }
 
-export class ViewModel<O extends Options> extends EventEmitter<{change: never}> {
+export class ViewModel<O extends Options = Options> extends EventEmitter<{change: never}> {
```
bwindels: can you remove this commit together with the one that introduced the change (and force push)?

src/domain/ViewModel.ts
bwindels: which line? the link just points to the repo

src/domain/login/LoginViewModel.ts
```diff
@@ -15,34 +15,51 @@ limitations under the License.
...
+import {Options as BaseOptions, ViewModel} from "../ViewModel";
 import {PasswordLoginViewModel} from "./PasswordLoginViewModel.js";
 import {StartSSOLoginViewModel} from "./StartSSOLoginViewModel.js";
 import {CompleteSSOLoginViewModel} from "./CompleteSSOLoginViewModel.js";
 import {LoadStatus} from "../../matrix/Client.js";
 import {SessionLoadViewModel} from "../SessionLoadViewModel.js";
 
-export class LoginViewModel extends ViewModel {
-    constructor(options) {
+// TODO(isaiah): make these default exported
```
bwindels: Please don't use default exports anywhere, see https://github.com/vector-im/hydrogen-web/blob/master/codestyle.md

src/domain/login/LoginViewModel.ts
```diff
@@ -15,34 +15,51 @@ limitations under the License.
...
+import {PasswordLoginMethod} from "../../matrix/login/PasswordLoginMethod";
```
bwindels: These can all be `import type` I imagine

src/domain/login/LoginViewModel.ts
```diff
@@ -15,34 +15,51 @@ limitations under the License.
...
-        this._loadViewModelSubscription = null;
         this._homeserver = defaultHomeserver;
-        this._queriedHomeserver = null;
-        this._errorMessage = "";
-        this._hideHomeserver = false;
-        this._isBusy = false;
-        this._abortHomeserverQueryTimeout = null;
-        this._abortQueryOperation = null;
-        this._initViewModels();
+        void this._initViewModels();
```
bwindels: OOI, why do you do this? Because it's an async method?

src/domain/login/LoginViewModel.ts
```diff
@@ -93,23 +110,23 @@ export class LoginViewModel extends ViewModel {
...
-    _setBusy(status) {
+    _setBusy(status: boolean) {
         this._isBusy = status;
         this._passwordLoginViewModel?.setBusy(status);
         this._startSSOLoginViewModel?.setBusy(status);
         this.emitChange("isBusy");
     }
 
-    async attemptLogin(loginMethod) {
+    async attemptLogin(loginMethod: PasswordLoginMethod) {
```
bwindels: Should this be `ILoginMethod` rather?

src/domain/login/LoginViewModel.ts
```diff
@@ -152,7 +169,7 @@ export class LoginViewModel extends ViewModel {
     }
 
     _disposeViewModels() {
-        this._startSSOLoginViewModel = this.disposeTracked(this._ssoLoginViewModel);
+        this._startSSOLoginViewModel = this.disposeTracked(this._startSSOLoginViewModel);
```
bwindels, line 197: good catch, we were leaking `_startSSOLoginViewModel` so far then?

src/domain/login/LoginViewModel.ts
```diff
@@ -15,34 +15,51 @@ limitations under the License.
...
-        this._errorMessage = "";
-        this._hideHomeserver = false;
-        this._isBusy = false;
-        this._abortHomeserverQueryTimeout = null;
-        this._abortQueryOperation = null;
-        this._initViewModels();
+        void this._initViewModels();
     }
 
     get passwordLoginViewModel() { return this._passwordLoginViewModel; }
```
bwindels: can you add return types to all the getters and methods as well please?

src/domain/login/LoginViewModel.ts
```diff
@@ -15,34 +15,51 @@ limitations under the License.
...
         this._client = new Client(this.platform);
-        this._loginOptions = null;
-        this._passwordLoginViewModel = null;
-        this._startSSOLoginViewModel = null;
-        this._completeSSOLoginViewModel = null;
-        this._loadViewModel = null;
-        this._loadViewModelSubscription = null;
         this._homeserver = defaultHomeserver;
-        this._queriedHomeserver = null;
-        this._errorMessage = "";
```
bwindels, line 40: usually we add the type annotation for props anyway, regardless of whether typescript can infer it or not, as it's a bit easier to parse for the humans ...

bwindels (commented):
This looks great, thank you for your PR. Just a few nits about our (not always documented) habits.

src/domain/ViewModel.ts
ibeckermayer: Oops. Meant to link [here](https://github.com/ibeckermayer/hydrogen-web/blob/5c5c6007047e10cf0dc00c9d9c5be1b77db99eae/src/domain/login/LoginViewModel.ts#L182).

src/domain/login/LoginViewModel.ts
ibeckermayer: Right, that was more of a note to self and not using proper terminology (although I used it in the commit message, which isn't ideal). What I meant was add them to `index.ts` so that I didn't need to import from each file separately, see

https://github.com/vector-im/hydrogen-web/pull/777/commits/5c5c6007047e10cf0dc00c9d9c5be1b77db99eae

src/domain/login/LoginViewModel.ts
ibeckermayer: Yes, I've been using the `.ts-eslintrc.js` which was complaining at me
```
Promises must be handled appropriately or explicitly marked as ignored with the `void` operator.eslint[@typescript-eslint/no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/v4.29.2/packages/eslint-plugin/docs/rules/no-floating-promises.md)
```

Normally I would mark the function `async` and `await` it, but it's a constructor so I just `void`ed it.

src/domain/login/LoginViewModel.ts
ibeckermayer: It seems so, you were effectively calling
```javascript
        this._startSSOLoginViewModel = this.disposeTracked(undefined);
```

src/domain/login/LoginViewModel.ts
ibeckermayer: Ah yes, makes sense. Double check me here -- I think [this](https://github.com/ibeckermayer/hydrogen-web/blob/f8c47059447cbc12962f12a6373b49cf634cb721/src/domain/login/LoginViewModel.ts#L112) is meant to be `TokenLoginMethod` and then [this](https://github.com/ibeckermayer/hydrogen-web/blob/f8c47059447cbc12962f12a6373b49cf634cb721/src/domain/login/LoginViewModel.ts#L126) is `PasswordLoginMethod`.

src/domain/RootViewModel.js
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 import {Client} from "../matrix/Client.js";
 import {SessionViewModel} from "./session/SessionViewModel.js";
 import {SessionLoadViewModel} from "./SessionLoadViewModel.js";
-import {LoginViewModel} from "./login/LoginViewModel.js";
+import {LoginViewModel} from "./login/LoginViewModel.ts";
```
MidhunSureshR:
```suggestion
import {LoginViewModel} from "./login/LoginViewModel";
```

src/domain/login/LoginViewModel.ts
MidhunSureshR: That is correct :+1:

src/domain/login/LoginViewModel.ts
```diff
@@ -181,9 +223,9 @@ export class LoginViewModel extends ViewModel {
             }
         }
         this._abortHomeserverQueryTimeout = this.disposeTracked(this._abortHomeserverQueryTimeout);
-        this.queryHomeserver();
+        await this.queryHomeserver();
```
MidhunSureshR: Why `await` here and in other places? The intent is to not block further execution unless there's a reason to.

src/domain/login/LoginViewModel.ts
```diff
@@ -221,19 +263,29 @@ export class LoginViewModel extends ViewModel {
...
+
+type ReadyFn = (client: Client) => void;
+
+// TODO: move to Client.js when its converted to typescript.
+type LoginOptions = {
+    homeserver: string;
+    password?: (username: string, password: string) => PasswordLoginMethod;
+    sso?: SSOLoginHelper;
+    token?: (loginToken: string) => TokenLoginMethod;
+};
```
MidhunSureshR: Can you add a newline here?

src/domain/ViewModel.ts
MidhunSureshR: The argument passed to the `emitChange` method does nothing as of now but it will serve some purpose if Hydrogen is ported to another platform. So we'd like to have typescript shout if we forget to pass an argument to the method.

For [the call here](https://github.com/ibeckermayer/hydrogen-web/blob/5c5c6007047e10cf0dc00c9d9c5be1b77db99eae/src/domain/login/LoginViewModel.ts#L182), you can do `emitChange("loginViewModels")`.

MidhunSureshR (commented):
Thank you, leaving some comments here.

You'll also need to [sign-off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off) before we can merge this PR.

src/domain/login/LoginViewModel.ts
ibeckermayer: I wasn't thinking clearly, removed these in 40276298041a70058f5447795d591f48af83ac96

ibeckermayer:
https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off

Signed-off-by: Isaiah Becker-Mayer isaiah@becker-mayer.com

src/domain/login/LoginViewModel.ts
```diff
@@ -15,101 +15,143 @@ limitations under the License.
...
+    get isFetchingLoginOptions(): boolean {
+        return !!this._abortQueryOperation;
+    }
+
+    goBack(): void {
         this.navigation.push("session");
     }
 
-    async _initViewModels() {
+    _initViewModels(): void {
```
MidhunSureshR: Can you mark methods starting with _ as private?

MidhunSureshR (approved):
Just one more thing:

MidhunSureshR:
Looks like the linter is picking up lots of errors due to the new rules. Might make sense to add these rules in a separate PR which also fixes all these errors.

ibeckermayer:
@MidhunSureshR added those `private`s and reverted all the eslint changes for now. Looks ready to merge to me (you will have to do so since I don't have write access to the repo).

PR #785, opened 2022-07-06
Log the error when we can't get storage access

Half-Shot:
This is quite useful when debugging why a session isn't working properly.

### DIFF
--- a/src/matrix/storage/idb/StorageFactory.ts
+++ b/src/matrix/storage/idb/StorageFactory.ts
@@ -42,6 +42,7 @@ async function requestPersistedStorage(): Promise<boolean> {
             await glob.document.requestStorageAccess();
             return true;
         } catch (err) {
+            console.warn("requestStorageAccess threw an error:", err);
             return false;
         }
     } else {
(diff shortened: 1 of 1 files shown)


src/matrix/storage/idb/StorageFactory.ts
```diff
@@ -42,6 +42,7 @@ async function requestPersistedStorage(): Promise<boolean> {
             await glob.document.requestStorageAccess();
             return true;
         } catch (err) {
+            console.warn("requestStorageAccess threw an error:", err);
```
bwindels, line 45: Ideally we'd use the [logitem object](https://github.com/vector-im/hydrogen-web/blob/master/src/logging/LogItem.ts) in here we use elsewhere rather than the console logs. You can pass it in from StorageFactory.create.

src/matrix/storage/idb/StorageFactory.ts
```diff
@@ -65,10 +67,10 @@ export class StorageFactory {
 
     async create(sessionId: string, log: ILogItem): Promise<Storage> {
         await this._serviceWorkerHandler?.preventConcurrentSessionAccess(sessionId);
-        requestPersistedStorage().then(persisted => {
+        requestPersistedStorage(log).then(persisted => {
             // Firefox lies here though, and returns true even if the user denied the request
             if (!persisted) {
-                console.warn("no persisted storage, database can be evicted by browser");
+                log.log("no persisted storage, database can be evicted by browser:", LogLevel.Warn);
```
bwindels: you can use `log.level.Warn` fwiw, so you don't need the import.

bwindels (approved):

PR #788, opened 2022-07-07
Disable composer when user lacks powerlevel needed to send messages

MidhunSureshR:


src/domain/session/room/RoomViewModel.js
```diff
@@ -67,6 +67,29 @@ export class RoomViewModel extends ViewModel {
         this._clearUnreadAfterDelay();
     }
 
+    async _recreateComposerOnPowerLevelChange() {
+        const powerLevelObservable = await this._room.observePowerLevels();
+        let oldCanSendMessage = powerLevelObservable.get().canSendType("m.room.message");
```
bwindels:
```suggestion
        const canSendMessage = () => powerLevelObservable.get().canSendType("m.room.message");
        let oldCanSendMessage = canSendMessage();
```
And reuse that below as well?

src/domain/session/room/RoomViewModel.js
```diff
@@ -365,3 +388,13 @@ class ArchivedViewModel extends ViewModel {
...
     }
 }
+
+class LowerPowerLevelViewModel extends ViewModel {
+    get description() {
+        return this.i18n`You do not have the powerlevel necessary to send messages`;
+    }
+
+    get kind() {
+        return "low-powerlevel";
```
bwindels: why not use the same kind (e.g. "disabled") for both `ArchivedViewModel` and this vm? the `kind` property is usually signalling how a vm should be rendered, so makes sense to use the same value if they are rendered in the same way.

bwindels (approved):
look great, just a few cleanups

PR #793, opened 2022-07-09
Typescript conversion for `src/observable/map`

ibeckermayer:
The conversion of `domain/session/leftpanel` was being blocked by typescript's inability to pick up on the methods of `BaseObservableMap` when they were defined like the following in order to avoid circular dependencies:

```javascript
// avoid circular dependency between these classes
// and BaseObservableMap (as they extend it)
Object.assign(BaseObservableMap.prototype, {
    sortValues(comparator) {
        return new SortedMapList(this, comparator);
    },

    mapValues(mapper, updater) {
        return new MappedMap(this, mapper, updater);
    },

    filterValues(filter) {
        return new FilteredMap(this, filter);
    },

    join(...otherMaps) {
        return new JoinedMap([this].concat(otherMaps));
    }
});
```

This PR fixes that problem by adding those functions as abstract functions in `BaseObservableMap`, and adding their default implementations in a new class called `BaseObservableMapDefaults`.

Each class that inherits `BaseObservableMap<K, V>` is then initialized with a `BaseObservableMapDefaults` field like:
https://github.com/ibeckermayer/hydrogen-web/blob/f0ebcfcbfe4f2a007670bafae9489df413909d68/src/observable/map/ObservableMap.ts#L25-L26

and then some boilerplate is added to call the default implementations:
https://github.com/ibeckermayer/hydrogen-web/blob/f0ebcfcbfe4f2a007670bafae9489df413909d68/src/observable/map/ObservableMap.ts#L102-L116

 and some boilerplate which calls the default implementation, which is now defined in `config.ts`.

## Update 1

Found a way to limit the boilerplate: https://github.com/vector-im/hydrogen-web/commit/c6efdc7513d04e788585affbb639f2e4ae7e8787


.ts-eslintrc.js
```diff
@@ -19,6 +19,11 @@ module.exports = {
     ],
     rules: {
         "@typescript-eslint/no-floating-promises": 2,
-        "@typescript-eslint/no-misused-promises": 2
+        "@typescript-eslint/no-misused-promises": 2,
+        "no-unused-vars": "off",
+        "@typescript-eslint/no-unused-vars": ["warn"],
```
ibeckermayer, line 24: See https://github.com/import-js/eslint-plugin-import/issues/653#issuecomment-840228881 (I was getting the error in the OP of that thread when trying to lint)

.ts-eslintrc.js
```diff
@@ -19,6 +19,11 @@ module.exports = {
...
+        "no-undef": "off",
```
ibeckermayer, line 25: Recommended by ESlint's official guide: https://typescript-eslint.io/docs/linting/troubleshooting/#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors

.ts-eslintrc.js
```diff
@@ -19,6 +19,11 @@ module.exports = {
...
+        "semi": ["error", "always"],
+        "@typescript-eslint/explicit-function-return-type": ["error"]
```
ibeckermayer, line 27: Added based on feedback in my first PR.

ibeckermayer:
Also: I tried for the life of me to convert `SortedMapList.js`, but couldn't define a plausible type system that fits how its currently written.

src/observable/map/ApplyMap.ts
```diff
@@ -0,0 +1,125 @@
...
+import {MappedMap} from "./MappedMap";
+import {FilteredMap} from "./FilteredMap";
+import {SortedMapList} from "../list/SortedMapList.js";
+
+
+export class ApplyMap<K, V> extends BaseObservableMap<K, V> {
+    private _source: BaseObservableMap<K, V>;
+    private _subscription?: SubscriptionHandle;
+    private _apply?: Apply<K, V>;
+    private _config: BaseObservableMapConfig<K, V>;
```
ibeckermayer, line 30: All classes which `extend` `BaseObservableMap` get this private config variable.

MidhunSureshR:
I wonder if we can use a mixin of some sort here. In `index.ts` where you import and export the classes, you define a function:
```js
const NewSortedMapList = mapCreator(SortedMapList);
export { NewSortedMapList as SortedMapList };

function mapCreator(mapClass) {
    return class extends mapClass {
        sortValues(comparator) {
            return new SortedMapList(this, comparator);
        }

        mapValues(mapper, updater) {
            return new MappedMap(this, mapper, updater);
        }

        filterValues(filter) {
            return new FilteredMap(this, filter);
        }

        join(...otherMaps) {
            return new JoinedMap([this].concat(otherMaps));
        }
    };
}
```
Haven't tried it out but it's worth a try.

ibeckermayer:
@MidhunSureshR I wasn't able to figure out a good way to do it using mixins, however I did find a way to eliminate a good amount of the boilerplate:

c6efdc7513d04e788585affbb639f2e4ae7e8787

ibeckermayer:
I just noticed that there's a [constrained mixin](https://www.typescriptlang.org/docs/handbook/mixins.html#constrained-mixins) pattern that might work for us, I'll play around with that.

ibeckermayer:
#### Couldn't get [mixin](https://www.typescriptlang.org/docs/handbook/mixins.html#constrained-mixins) to work

I did some messing around today, long story short is that I didn't find a good way use mixins, even when using constraints. 

A core issue is that mixins don't play nicely with `abstract` (which our `BaseObservableMap` is):

```typescript
type GConstructor<T = {}> = new (...args: any[]) => T;
type IsABaseObservableMap<K, V> = GConstructor<BaseObservableMap<K, V>>;

function makeTransformable<K, V, TBase extends IsABaseObservableMap<K, V>>(Base: TBase) {
    return class extends Base {
        join(...otherMaps: Array<BaseObservableMap<K, V>>): JoinedMap<K, V> {
            return new JoinedMap([this].concat(otherMaps));
        }

        mapValues(mapper: Mapper<V>, updater?: Updater<V>): MappedMap<K, V> {
            return new MappedMap(this, mapper, updater);
        }

        sortValues(comparator: Comparator<V>): SortedMapList {
            return new SortedMapList(this, comparator);
        }

        filterValues(filter: Filter<K, V>): FilteredMap<K, V> {
            return new FilteredMap(this, filter);
        }
    }
}
```

The returned class complains 

image

I also ran into another typing error which I wasn't able to decipher, wherein the type system seems to get tripped up on `join`, complaining about the `otherMaps` parameter:

image

It's as if it doesn't realize that `this` and `otherMaps[n]` are the same type, seems to me like a bug in the type system. 

#### What I did instead

I changed `BaseObservableMapDefaults` to the more descriptive `BaseObservableMapTransformers`, and added [a more comprehensive description](https://github.com/vector-im/hydrogen-web/pull/793/commits/05622d9204d2ba02d731dc557e387bce009f07db#diff-7bf13fde365b2c587e733dc86ad014bf3d2bd8d548ceb38e98548f7163e479e6R23-R42) of why we're doing it this way. 

IMO it would be nicer if we could get mixins to work in a manner similar to what I tried above, but given the issues with the type system, this is good enough. (Naturally, feel free to show me how its done if you think you see a path I missed).

I'm going to leave those merge conflicts unresolved until we figure out what's going on [here](https://matrix.to/#/!LFSvgabwCbgYyABWyM:matrix.org/$BjokrR9Pt8Ut3bD_Zpj83xubizrTI_XbyZkkvQ7fyCA?via=matrix.org&via=envs.net&via=element.io).

@MidhunSureshR

MidhunSureshR:
```ts
import { SortedMapList as _SortedMapList } from "./list/SortedMapList.js";
import { MappedMap as _MappedMap } from "./map/MappedMap";
import { FilteredMap as _FilteredMap } from "./map/FilteredMap";
import { JoinedMap as _JoinedMap } from "./map/JoinedMap";
import { BaseObservableMap as _BaseObservableMap } from "./map/BaseObservableMap.js";

// re-export "root" (of chain) collection
export { ObservableMap } from "./map/ObservableMap";
export { ObservableArray } from "./list/ObservableArray";
export { SortedArray } from "./list/SortedArray";
export { MappedList } from "./list/MappedList";
export { AsyncMappedList } from "./list/AsyncMappedList";
export { ConcatList } from "./list/ConcatList";

type GConstructor<T = {}> = abstract new (...args: any[]) => T;
type MapConstructor<K, V> = GConstructor<_BaseObservableMap<K, V>>;

function addTransformers<K, V, T extends MapConstructor<K, V>>(MyClass: T) {
    abstract class ClassWithTransformers extends MyClass {
        sortValues(comparator) {
            return new _SortedMapList(this, comparator);
        }

        mapValues(mapper, updater) {
            return new _MappedMap(this, mapper, updater);
        }

        filterValues(filter) {
            return new _FilteredMap(this, filter);
        }

        join(...otherMaps) {
            return new _JoinedMap([this].concat(otherMaps));
        }
    }
    return ClassWithTransformers;
}

export const SortedMapList = addTransformers(_SortedMapList);
export const MappedMap = addTransformers(_MappedMap);
export const FilteredMap = addTransformers(_FilteredMap);
export const JoinedMap = addTransformers(_JoinedMap);
```
@ibeckermayer would this work?

ibeckermayer:
@MidhunSureshR Clever, it appears to work: 7d27d4676fb12be5eff60950ff06db5f65155fcd

The primary downsides I see to this approach are:

1. It messes with the type system in a couple of ways. 
 
One way is that from a straightforward reading of the code, it seems like we are instantiating `abstract` classes directly (anything that's been passed through `addTransformers`). Nevertheless we can get away with it, maybe because according to my editor the return type of `addTransformers is
```typescript
function addTransformers<K, V, T extends MapConstructor<K, V>>(MyClass: T): ((abstract new (...args: any[]) => ClassWithTransformers) & {
    prototype: addTransformers<any, any, any>.ClassWithTransformers;
}) & T
```

which is something, but I'm not exactly clear what.

As an example `export const ObservableMap = addTransformers(_ObservableMap);` gives us a type like

```typescript
const ObservableMap: ((abstract new (...args: any[]) => addTransformers<unknown, unknown, typeof _ObservableMap>.ClassWithTransformers) & {
    prototype: addTransformers<any, any, any>.ClassWithTransformers;
}) & typeof _ObservableMap
```

It's not exactly clear to me why the type system allows me to instantiate that, but the overarching point is that the types of these classes are now more or less inscrutable.

2. We aren't expressing that all `BaseObservableMap` are transformable by `sortValues`, `mapValues`, etc. 

In other words, there's a semantic difference between this solution and the previous one -- previously we were saying that all `BaseObservableMap` have these transformers, now we're saying that only special ones which are passed through `addTransformers` do.

In both cases the code around this part gets a bit "weird". I'm partial to my solution from before, because outside of `observable/map`, all the "weirdness" goes away, in that the types just look like normal types again, and we're not playing any games with instantiating what appears to be an abstract class. I also think that "all `BaseObservableMap` are transformable by the following methods" is the original intent, and that is better captured by the previous solution.

ibeckermayer:
@MidhunSureshR Any thoughts on my argument above?

ibeckermayer:
https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off

Signed-off-by: Isaiah Becker-Mayer [isaiah@becker-mayer.com](mailto:isaiah@becker-mayer.com)

MidhunSureshR:
Sorry, it took me a bit of time to come back to this. One issue with both the approaches is that we're making non trivial changes to the code to get static typing to work. Can we instead do some thing like [this](https://www.typescriptlang.org/play?#code/PTAECkHsEsDsGFIBskFMDGAXalagIawAmoAsvgA6Ioba6gCuAzqgaNALYWQBOmomAJ4VWAC1Q9WmSAQBuMEumg90DJPh7suvTACg4mCQDN86VgHkRPfNJ5NQAb12gXoAFYxYACg4AuUEyYPHAA5gCU-lBw1GhYOLAA3M6uHJReEWSUMbTxSQC+urro6kz2AEL4LNlx9E6uoEaQkOn+sAwcAEYSjqCSmAw8eACsCaAFBUUl9lEIyLF0eKgAHobE5ZWo1QuOyS7ouIE8DFi8XhTBsjassK3tXTwANKDn0JeGoByopfghqP6HoTCO3q9SYDCs6SS9QK9Q6GhaoDanW6Dl6qH6gwEomgTAAdIkxroJsVKvZyFQ5jl6MtVkR1lVKTU8HVXHCAF5shFI+49PoDPAAdlG40K5g6blouNJ0BC3gqDJoTNx50g0iEIieLJcHjgXhuASCgMini28V56P5iNQAHcICbGQsvAAWABMT1gYWFD12HzSGXJptqaIxeFgNsyFMVjs9hLyBHslgkNl4TE9hXVrAACgwOkhoOgZoHYAAVYSsAC8VttXlxtYoGnwHCY-kQsEOx1smYbn0MdgAPBnIEY7dEHfEAHxA8vjkezKNmgBkoET1lsTCS+zb-GzufzhbH9H8O7zBft89wpZEoEr+-PeEqoGPe7P83il9QSV0y24fEcT9Po53vGs5Fnkn6boEDRNNeVaPjmJ63q+uDOm6oAAES5jmaFphB-CpBQMGNJAuL4ZCRQHMgqC4kgkAhD4lC4uynJhJ6QA)? Thank you for working on this.

ibeckermayer:
For reference, I found a bug in the `MappedMap` here while working on `rightpanel`, which I addressed here: 92ed503700e9c8350849e62067f295eead66865d (on a separate branch for now).

> Sorry, it took me a bit of time to come back to this. One issue with both the approaches is that we're making non trivial changes to the code to get static typing to work. Can we instead do some thing like [this](https://www.typescriptlang.org/play?#code/PTAECkHsEsDsGFIBskFMDGAXalagIawAmoAsvgA6Ioba6gCuAzqgaNALYWQBOmomAJ4VWAC1Q9WmSAQBuMEumg90DJPh7suvTACg4mCQDN86VgHkRPfNJ5NQAb12gXoAFYxYACg4AuUEyYPHAA5gCU-lBw1GhYOLAA3M6uHJReEWSUMbTxSQC+urro6kz2AEL4LNlx9E6uoEaQkOn+sAwcAEYSjqCSmAw8eACsCaAFBUUl9lEIyLF0eKgAHobE5ZWo1QuOyS7ouIE8DFi8XhTBsjassK3tXTwANKDn0JeGoByopfghqP6HoTCO3q9SYDCs6SS9QK9Q6GhaoDanW6Dl6qH6gwEomgTAAdIkxroJsVKvZyFQ5jl6MtVkR1lVKTU8HVXHCAF5shFI+49PoDPAAdlG40K5g6blouNJ0BC3gqDJoTNx50g0iEIieLJcHjgXhuASCgMini28V56P5iNQAHcICbGQsvAAWABMT1gYWFD12HzSGXJptqaIxeFgNsyFMVjs9hLyBHslgkNl4TE9hXVrAACgwOkhoOgZoHYAAVYSsAC8VttXlxtYoGnwHCY-kQsEOx1smYbn0MdgAPBnIEY7dEHfEAHxA8vjkezKNmgBkoET1lsTCS+zb-GzufzhbH9H8O7zBft89wpZEoEr+-PeEqoGPe7P83il9QSV0y24fEcT9Po53vGs5Fnkn6boEDRNNeVaPjmJ63q+uDOm6oAAES5jmaFphB-CpBQMGNJAuL4ZCRQHMgqC4kgkAhD4lC4uynJhJ6QA)? Thank you for working on this.

No worries, I will take a look in detail and see if we can get it to work. At first glance I'm skeptical, because this contains an `Object.assign` that looks similar to the original solution that was causing Typescript to falter (it wasn't smart enough to recognize the `Object.assign` would give us the assigned methods, and so wouldn't let us call them later, elsewhere). Granted I've only glanced at your proposal, so perhaps you've addressed this and I haven't understood it yet.

It appears that the general problem we're trying to solve is that we have a `BaseClass` for which we want to define functions which return `ChildClass`es that themselves extend that `BaseClass`, without circular imports. Perhaps there's a summary of prior art on this topic that I can find. (Just thinking out loud a bit here).

ibeckermayer:
@MidhunSureshR 

> Sorry, it took me a bit of time to come back to this. One issue with both the approaches is that we're making non trivial changes to the code to get static typing to work. Can we instead do some thing like [this](https://www.typescriptlang.org/play?#code/PTAECkHsEsDsGFIBskFMDGAXalagIawAmoAsvgA6Ioba6gCuAzqgaNALYWQBOmomAJ4VWAC1Q9WmSAQBuMEumg90DJPh7suvTACg4mCQDN86VgHkRPfNJ5NQAb12gXoAFYxYACg4AuUEyYPHAA5gCU-lBw1GhYOLAA3M6uHJReEWSUMbTxSQC+urro6kz2AEL4LNlx9E6uoEaQkOn+sAwcAEYSjqCSmAw8eACsCaAFBUUl9lEIyLF0eKgAHobE5ZWo1QuOyS7ouIE8DFi8XhTBsjassK3tXTwANKDn0JeGoByopfghqP6HoTCO3q9SYDCs6SS9QK9Q6GhaoDanW6Dl6qH6gwEomgTAAdIkxroJsVKvZyFQ5jl6MtVkR1lVKTU8HVXHCAF5shFI+49PoDPAAdlG40K5g6blouNJ0BC3gqDJoTNx50g0iEIieLJcHjgXhuASCgMini28V56P5iNQAHcICbGQsvAAWABMT1gYWFD12HzSGXJptqaIxeFgNsyFMVjs9hLyBHslgkNl4TE9hXVrAACgwOkhoOgZoHYAAVYSsAC8VttXlxtYoGnwHCY-kQsEOx1smYbn0MdgAPBnIEY7dEHfEAHxA8vjkezKNmgBkoET1lsTCS+zb-GzufzhbH9H8O7zBft89wpZEoEr+-PeEqoGPe7P83il9QSV0y24fEcT9Po53vGs5Fnkn6boEDRNNeVaPjmJ63q+uDOm6oAAES5jmaFphB-CpBQMGNJAuL4ZCRQHMgqC4kgkAhD4lC4uynJhJ6QA)? Thank you for working on this.

The example is close to what we [currently have](https://github.com/ibeckermayer/hydrogen-web/blob/869978517efe39e00326f9fec23f376464f53df8/src/observable/index.js#L30-L49) and causes the same problem. Namely, classes that extend `BaseObservableMap` don't have the `Object.assign`ed methods according to Typescript.

ibeckermayer:
Alright, turns out we aren't the first to encounter this general problem. I was able to solve it by arranging the export order in `src/observable/map/index.ts` and then always importing these classes from that file only. Based on [this blog post](https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de), and have added [a comment](https://github.com/ibeckermayer/hydrogen-web/blob/ebd8c0751a11e63e877d74a68dd836cac4fcfa5c/src/observable/map/index.ts#L1-L9) in `index.ts` and this [warning label](https://github.com/ibeckermayer/hydrogen-web/blob/ebd8c0751a11e63e877d74a68dd836cac4fcfa5c/src/observable/map/ApplyMap.ts#L21-L25) on all the classes.

There's a lot of commits in here now from me going down various paths, if it's not breaking a project policy then it might be best to squash and merge this one. If that's not an option let me know, and I can try and work some gitfu to clean the history up.

src/observable/list/SortedArray.ts
```diff
@@ -127,6 +128,7 @@ class Iterator<T> {
         this._current = null;
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
MidhunSureshR: You can avoid having to disable ts here by re-implementing the Iterator class as follows:
```ts
class Iterator<T> {
    private _sortedArray: SortedArray<T>;
    private _current: T | null | undefined;
    private _consumed: boolean = false;

    constructor(sortedArray: SortedArray<T>) {
        this._sortedArray = sortedArray;
        this._current = null;
    }

    next(): IteratorResult<T> {
        if (this._consumed) {
            return {value: undefined, done: true};
        }
        this._current = this._current? this._sortedArray._getNext(this._current): this._sortedArray.get(0);
        if (!this._current) {
            this._consumed = true;
        }
        return { value: this._current, done: this._consumed } as IteratorResult<T>;
    }
}
```

src/observable/list/SortedArray.ts
```diff
@@ -112,6 +112,7 @@ export class SortedArray<T> extends BaseObservableList<T> {
         return this._items.length;
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
     [Symbol.iterator]() {
```
MidhunSureshR:
```suggestion
    [Symbol.iterator](): Iterator<T> {
```

src/observable/list/ObservableArray.ts
```diff
@@ -75,6 +75,7 @@ export class ObservableArray<T> extends BaseObservableList<T> {
         return this._items.length;
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
     [Symbol.iterator]() {
```
MidhunSureshR:
```suggestion
    [Symbol.iterator](): IterableIterator<T> {
```

src/observable/list/ConcatList.ts
```diff
@@ -86,10 +86,12 @@ export class ConcatList<T> extends BaseObservableList<T> implements IListObserve
         return len;
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
     [Symbol.iterator]() {
         let sourceListIdx = 0;
         let it = this._sourceLists[0][Symbol.iterator]();
         return {
+            // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
MidhunSureshR: Can we avoid the eslint disable here as well?
```ts
    [Symbol.iterator](): IterableIterator<T> {
        let sourceListIdx = 0;
        let it = this._sourceLists[0][Symbol.iterator]();
        return {
            next: () => {
                let result = it.next();
                while (result.done) {
                    sourceListIdx += 1;
                    if (sourceListIdx >= this._sourceLists.length) {
                        return result;  //done
                    }
                    it = this._sourceLists[sourceListIdx][Symbol.iterator]();
                    result = it.next();
                }
                return result;
            }
        } as IterableIterator<T>;
    }
}
```

src/observable/list/BaseMappedList.ts
```diff
@@ -37,14 +37,15 @@ export class BaseMappedList<F,T,R = T> extends BaseObservableList<T> {
...
+    findAndUpdate(predicate: (value: T) => boolean, updater: (value: T) => any | false): boolean {
         return findAndUpdateInArray(predicate, this._mappedValues!, this, updater);
     }
 
-    get length() {
+    get length(): number {
         return this._mappedValues!.length;
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
MidhunSureshR: same here

src/observable/map/ApplyMap.ts
```diff
@@ -14,80 +14,93 @@ See the License for the specific language governing permissions and
...
             this.applyOnce(this._apply);
         }
         super.onSubscribeFirst();
     }
 
-    onUnsubscribeLast() {
+    onUnsubscribeLast(): void {
         super.onUnsubscribeLast();
-        this._subscription = this._subscription();
+        if (this._subscription) this._subscription = this._subscription();
```
MidhunSureshR:
```suggestion
        if (this._subscription) {
         this._subscription = this._subscription();
         }
```

src/observable/map/ApplyMap.ts
```diff
@@ -14,80 +14,93 @@ See the License for the specific language governing permissions and
...
 
-    onReset() {
+    onReset(): void {
         if (this._apply) {
             this.applyOnce(this._apply);
         }
         this.emitReset();
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
MidhunSureshR: same here

src/observable/map/BaseObservableMap.ts
```diff
@@ -23,33 +28,70 @@ export interface IMapObserver<K, V> {
...
     }
 
-    emitRemove(key, value) {
+    emitRemove(key: K, value: V): void {
         for(let h of this._handlers) {
             h.onRemove(key, value);
         }
     }
 
+    join(...otherMaps: Array<typeof this>): JoinedMap<K, V> {
```
MidhunSureshR, line 67: I get an error here when I run `yarn run tsc`:
```
src/observable/map/BaseObservableMap.ts:67:37 - error TS2304: Cannot find name 'this'.
```

src/observable/map/FilteredMap.ts
```diff
@@ -106,51 +123,57 @@ export class FilteredMap extends BaseObservableMap {
...
+        }
     }
 
-    onReset() {
+    onReset(): void {
         this._reapplyFilter();
         this.emitReset();
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
MidhunSureshR: same

src/observable/map/FilteredMap.ts
```diff
@@ -106,51 +123,57 @@ export class FilteredMap extends BaseObservableMap {
...
+        this._included?.forEach(included => {
             if (included) {
                 count += 1;
             }
         });
         return count;
     }
 
-    get(key) {
+    get(key): V | undefined{
```
MidhunSureshR:
```suggestion
    get(key: K): V | undefined {
```

src/observable/map/FilteredMap.ts
```diff
@@ -106,51 +123,57 @@ export class FilteredMap extends BaseObservableMap {
...
-        this._included = _included;
+class FilterIterator<K, V> {
+    private _included?: Map<K, boolean>
+    private _sourceIterator: Iterator<[K, V], any, undefined>
+    constructor(map: BaseObservableMap<K, V>, included?: Map<K, boolean>) {
+        this._included = included;
         this._sourceIterator = map[Symbol.iterator]();
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
MidhunSureshR: same

src/observable/map/LogMap.ts
```diff
@@ -0,0 +1,87 @@
...
+        if (this._subscription) this._subscription = this._subscription();
+        this.log("unsubscribeLast");
+    }
+
+    onReset(): void {
+        this.log("reset");
+        this.emitReset();
+    }
+
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
MidhunSureshR: same here

src/observable/map/MappedMap.ts
```diff
@@ -14,55 +14,74 @@ See the License for the specific language governing permissions and
...
         this._mappedValues.set(key, mappedValue);
         this.emitAdd(key, mappedValue);
     }
 
-    onRemove(key/*, _value*/) {
+    onRemove(key: K/*, _value*/): void {
         const mappedValue = this._mappedValues.get(key);
         if (this._mappedValues.delete(key)) {
-            this.emitRemove(key, mappedValue);
+            if (mappedValue) this.emitRemove(key, mappedValue);
```
MidhunSureshR:
```suggestion
            if (mappedValue) {
             this.emitRemove(key, mappedValue);
             }
```

src/observable/map/MappedMap.ts
```diff
@@ -72,26 +91,27 @@ export class MappedMap extends BaseObservableMap {
...
         this._mappedValues.clear();
     }
 
-    onReset() {
+    onReset(): void {
         this._mappedValues.clear();
         this.emitReset();
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
MidhunSureshR: same

src/observable/map/ObservableMap.ts
```diff
@@ -91,11 +97,12 @@ export class ObservableMap<K, V> extends BaseObservableMap<K, V> {
     keys(): Iterator<K> {
         return this._values.keys();
     }
-}
+};
```
MidhunSureshR:
```suggestion
}
```

MidhunSureshR (commented):
Taking an initial look at the PR:

src/observable/map/BaseObservableMap.ts
ibeckermayer: Hmm, interesting, that's not happening for me. Maybe try it again after a `yarn install`?

src/observable/list/ConcatList.ts
ibeckermayer: I think this is more accurate:

```typescript
[Symbol.iterator](): Iterator<T> {
        let sourceListIdx = 0;
        let it = this._sourceLists[0][Symbol.iterator]();
        return {
            next: (): IteratorResult<T> => {
                let result = it.next();
                while (result.done) {
                    sourceListIdx += 1;
                    if (sourceListIdx >= this._sourceLists.length) {
                        return result;  //done
                    }
                    it = this._sourceLists[sourceListIdx][Symbol.iterator]();
                    result = it.next();
                }
                return result;
            }
        };
    }
```

See https://github.com/vector-im/hydrogen-web/commit/d025c1111e98ac481dfdf685ed54b2ca9a30e09c#diff-5333ee9971207c195c9074602d2e629227255e77a8ec4d7395b98a2b41cec3f1R89-R96

src/observable/map/BaseObservableMap.ts
MidhunSureshR: That fixed it!

ibeckermayer:
@MidhunSureshR I think this one is very close to ready.

MidhunSureshR (approved):
Thanks :tada:

PR #794, opened 2022-07-11
Convert color.js and svg-colorizer.js to es6 modules

MidhunSureshR:
This allows us to reuse this code when implementing runtime themes.


MidhunSureshR:
Closing because change is made redundant in #769

PR #803, opened 2022-07-19
Adding the /join and others commands into the hydrogen project

Kaki-In:
I am actually adding the /join command, that is not yet read by Hydrogen.


Kaki-In:
# Last update

## Added the `/join` command. 

You can now join a room using the `/join` command (e.g. `/join #a:b.c`). Some problems can occur for the moments. If you find one, please make know.

## Improved the commands experimentation.

 - Making now the whole parsing in a separated function
 - Parsing with a split and a `switch` instruction.

Kaki-In:
Next last update : adding the functionnality to connect with the element id (e.g. @a:b.c) without having to precise the webserver.

PR #809, opened 2022-07-21
Implemented /join

### LINKED ISSUE #792: The /join command
**Is your feature request related to a problem? Please describe.**

I cannot figure out how to join a room as it's not simple `/join #hydrogen:matrix.org` which is universal outside of Element iOS.

**Describe the solution you'd like**

I would like to have a `/join` command accepting the room alias and if the room is join rule `knock` or `knock_restricted` (and the user doesn't have access through allowed room) prompt for a knocking message and knock the room instead. I guess a separate /knock command would be nice or accepting join reason which could then be used as a knock reason?

**Describe alternatives you've considered**

None regarding /join, for /knock see above

**Additional context**

I found https://github.com/vector-im/hydrogen-web/pull/358 which implies joining is somehow supported, but I don't know how. I am using Hydrogen as a PWA, so I don't see the address bar.

Ref: #733 #775

Kaki-In:
# Implemented the /join command. 

## Change that have been added 

 - add a method with a `switch` syntax to improve the commands manager.
 - add the /join command - we can now join a room using /join #roomname:server.addr
 - improve the error message display (css)

Should fix [this issue (#792)](https://github.com/vector-im/hydrogen-web/issues/792).

### DIFF
--- a/src/domain/session/room/RoomViewModel.js
+++ b/src/domain/session/room/RoomViewModel.js
@@ -214,9 +214,9 @@ export class RoomViewModel extends ViewModel {
                            const internalId = await this._options.client.session.joinRoom(roomName);
                            this.navigation.push("room", internalId);
                        } catch (exc) {
-                           if (exc.statusCode ?? exc.status === 400) {
+                           if ((exc.statusCode ?? exc.status) === 400) {
                                this._sendError = new Error(`/join : '${roomName}' was not legal room ID or room alias`);
-                           } else if ((exc.statusCode ?? exc.status === 404) || (exc.statusCode ?? exc.status === 502) || (exc.message == "Internal Server Error")) {
+                           } else if ((exc.statusCode ?? exc.status) === 404 || (exc.statusCode ?? exc.status) === 502 || exc.message == "Internal Server Error") {
                                this._sendError = new Error(`/join : room '${roomName}' not found`);
                            } else {
                                this._sendError = new Error("join syntax: /join <room-id>");
--- a/src/domain/session/room/RoomViewModel.js
+++ b/src/domain/session/room/RoomViewModel.js
@@ -244,15 +244,15 @@ export class RoomViewModel extends ViewModel {
                 msgtype = "m.text";
                 break;
             case "tableflip":
-                message="(╯°□°）╯︵ ┻━┻ " + args.join(" ");
+                message = "(╯°□°）╯︵ ┻━┻ " + args.join(" ");
                 msgtype = "m.text";
                 break;
             case "unflip":
-                message="┬──┬ ノ( ゜-゜ノ) " + args.join(" ");
+                message = "┬──┬ ノ( ゜-゜ノ) " + args.join(" ");
                 msgtype = "m.text";
                 break;
             case "lenny":
-                message="( ͡° ͜ʖ ͡°) " + args.join(" ");
+                message = "( ͡° ͜ʖ ͡°) " + args.join(" ");
                 msgtype = "m.text";
                 break;
             default:
--- a/src/domain/session/room/RoomViewModel.js
+++ b/src/domain/session/room/RoomViewModel.js
@@ -230,7 +230,7 @@ export class RoomViewModel extends ViewModel {
                 if (args.length == 1) {
                     const roomName = args[0];
                     const exc = await this.joinRoom(roomName);
-                    if (exc!==true) {
+                    if (exc !== true) {
                         if (exc && exc.stack && exc.message) {
                             this._sendError = exc;
                         } else {
@@ -273,21 +273,21 @@ export class RoomViewModel extends ViewModel {
     
     async _sendMessage(message, replyingTo) {
         if (!this._room.isArchived && message) {
-            let messinfo = {msgtype : "m.text", message : message};
+            let messinfo = {type : "m.text", message : message};
             if (message.startsWith("//")) {
                 messinfo.message = message.substring(1).trim();
             } else if (message.startsWith("/")) {
                 messinfo = await this._processCommand(message);
             }
             try {
-               const msgtype = messinfo.type;
-               const message = messinfo.message;
-               if (msgtype && message) {
-                   if (replyingTo) {
-                           await replyingTo.reply(msgtype, message);
-                   } else {
-                           await this._room.sendEvent("m.room.message", {msgtype, body: message});
-                   }
+                const msgtype = messinfo.type;
+                const message = messinfo.message;
+                if (msgtype && message) {
+                    if (replyingTo) {
+                        await replyingTo.reply(msgtype, message);
+                    } else {
+                        await this._room.sendEvent("m.room.message", {msgtype, body: message});
+                    }
                 }
             } catch (err) {
                 console.error(`room.sendMessage(): ${err.message}:\n${err.stack}`);
--- a/src/domain/session/room/RoomViewModel.js
+++ b/src/domain/session/room/RoomViewModel.js
@@ -23,6 +23,7 @@ import {imageToInfo} from "../common.js";
 // TODO: remove fallback so default isn't included in bundle for SDK users that have their custom tileClassForEntry
 // this is a breaking SDK change though to make this option mandatory
 import {tileClassForEntry as defaultTileClassForEntry} from "./timeline/tiles/index";
+import {RoomStatus} from "../../../matrix/room/common";
 
 export class RoomViewModel extends ViewModel {
     constructor(options) {
@@ -210,7 +211,7 @@ export class RoomViewModel extends ViewModel {
                     let roomName = args[0];
                     try {
                         const roomId = await this._options.client.session.joinRoom(roomName);
-                        await session.observeRoomStatus(roomId).waitFor(status === RoomStatus.Joined);
+                        await (await this._options.client.session.observeRoomStatus(roomId)).waitFor(status => status === RoomStatus.Joined);
                         this.navigation.push("room", roomId);
                     } catch (exc) {
                         if ((exc.statusCode ?? exc.status) === 400) {
@@ -220,7 +221,7 @@ export class RoomViewModel extends ViewModel {
                         } else if ((exc.statusCode ?? exc.status) === 403) {
                             this._sendError = new Error(`/join : you're not invited to join '${roomName}'`);
                         } else {
-                            this._sendError = new Error("join syntax: /join <room-id>");
+                            this._sendError = exc;
                         }
                         this._timelineError = null;
                         this.emitChange("error");
--- a/src/domain/session/room/RoomViewModel.js
+++ b/src/domain/session/room/RoomViewModel.js
@@ -230,7 +230,7 @@ export class RoomViewModel extends ViewModel {
                 msgtype = "m.emote";
                 break;
             case "join":
-                if (args.length == 1) {
+                if (args.length === 1) {
                     const roomName = args[0];
                     await this._processCommandJoin(roomName);
                 } else {
--- a/src/domain/session/room/RoomViewModel.js
+++ b/src/domain/session/room/RoomViewModel.js
@@ -210,8 +210,8 @@ export class RoomViewModel extends ViewModel {
                    if (args.length == 1) {
                        let roomName = args[0];
                        try {
-                           await this._options.client.session.joinRoom(roomName);
                            const internalId = await this._options.client.session.joinRoom(roomName);
+                           await this._options.client.session.waitForRoomFromSync(internalId);
                            this.navigation.push("room", internalId);
                        } catch (exc) {
                            if ((exc.statusCode ?? exc.status) === 400) {
--- a/src/domain/session/room/RoomViewModel.js
+++ b/src/domain/session/room/RoomViewModel.js
@@ -416,6 +416,11 @@ export class RoomViewModel extends ViewModel {
             this._composerVM.setReplyingTo(entry);
         }
     }
+    
+    dismissError(evt) {
+        this._sendError = null;
+        this.emitChange("error");
+    }
 }
 
 function videoToInfo(video) {
--- a/src/domain/session/room/RoomViewModel.js
+++ b/src/domain/session/room/RoomViewModel.js
@@ -216,9 +216,9 @@ export class RoomViewModel extends ViewModel {
                        } catch (exc) {
                            if ((exc.statusCode ?? exc.status) === 400) {
                                this._sendError = new Error(`/join : '${roomName}' was not legal room ID or room alias`);
-                           } else if (exc.statusCode ?? exc.status) === 404 || (exc.statusCode ?? exc.status) === 502 || exc.message == "Internal Server Error") {
+                           } else if ((exc.statusCode ?? exc.status) === 404 || (exc.statusCode ?? exc.status) === 502 || exc.message == "Internal Server Error") {
                                this._sendError = new Error(`/join : room '${roomName}' not found`);
-                           } else if (exc.statusCode ?? exc.status) === 403) {
+                           } else if ((exc.statusCode ?? exc.status) === 403) {
                                this._sendError = new Error(`/join : you're not invited to join '${roomName}'`);
                            } else {
                                this._sendError = new Error("join syntax: /join <room-id>");
--- a/src/domain/session/room/RoomViewModel.js
+++ b/src/domain/session/room/RoomViewModel.js
@@ -198,22 +198,21 @@ export class RoomViewModel extends ViewModel {
         }
     }
     
-    async joinRoom(roomName) {
-    }
-    
     async _processCommandJoin(roomName) {
         try {
             const roomId = await this._options.client.session.joinRoom(roomName);
             const roomStatusObserver = await this._options.client.session.observeRoomStatus(roomId);
             await roomStatusObserver.waitFor(status => status === RoomStatus.Joined);
             this.navigation.push("room", roomId);
-        } catch (exc) {
+        } catch (err) {
             if ((exc.statusCode ?? exc.status) === 400) {
-                exc = new Error(`/join : '${roomName}' was not legal room ID or room alias`);
+                const exc = new Error(`/join : '${roomName}' was not legal room ID or room alias`);
             } else if ((exc.statusCode ?? exc.status) === 404 || (exc.statusCode ?? exc.status) === 502 || exc.message == "Internal Server Error") {
-                exc = new Error(`/join : room '${roomName}' not found`);
+                const exc = new Error(`/join : room '${roomName}' not found`);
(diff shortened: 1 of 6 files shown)


src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,80 @@ export class RoomViewModel extends ViewModel {
         }
     }
     
+    async _getMessageInformations (message) {
+        let msgtype = "m.text";
+        if (message.startsWith("/")) {
+            let command=message.substring(1).split(" ");
```
MidhunSureshR: You can use [array destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#array_destructuring) here to make this a bit more readable:
```suggestion
            const [command, ...args] = message.substring(1).split(" ");
```
This way you don't have to use `command[0]` everywhere.

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,80 @@ export class RoomViewModel extends ViewModel {
...
+            switch (command[0]) {
+                case "me":
+                   message = message.substring(4).trim();
```
MidhunSureshR: With the above suggestion, to get the rest of the message excluding the command, you can use:
```suggestion
                   message = args.join(" ").trim();
```
This is so you don't need to call `substring` with different argument for different commands.

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,80 @@ export class RoomViewModel extends ViewModel {
...
+                   msgtype = "m.emote";
+                   break;
+               case "join":
+                   if (command.length==2) {
+                       let roomname = message.substring(5).trim();
+                       try {
```
MidhunSureshR: A bit more simpler:
```suggestion
                   const [roomName] = args;
                   if (roomName) {
                       try {
```

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,80 @@ export class RoomViewModel extends ViewModel {
...
+                           await this._options.client.session.joinRoom(roomname);
+                       } catch (exc) {
+                           if (exc.statusCode??exc.status==400) {
```
MidhunSureshR: We always add a space before and after operators:
```suggestion
                           if (exc.statusCode ?? exc.status === 400) {
```
Can you make this change throughout? It should be possible to setup your IDE/editor to do this automatically.

MidhunSureshR: Also you probably meant to do `===` instead of `==`?

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,80 @@ export class RoomViewModel extends ViewModel {
...
+            switch (command[0]) {
+                case "me":
+                   message = message.substring(4).trim();
+                   msgtype = "m.emote";
+                   break;
+               case "join":
+                   if (command.length==2) {
+                       let roomname = message.substring(5).trim();
+                       try {
+                           await this._options.client.session.joinRoom(roomname);
```
MidhunSureshR: After joining the room, we should switch over to the new room:
```suggestion
                           const internalId = await this._options.client.session.joinRoom(roomName);
                           this.navigation.push("room", internalId);
```

MidhunSureshR: Also wanted to mention that we use camel casing for variables, so prefer `roomName` over `roomname`.

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,80 @@ export class RoomViewModel extends ViewModel {
...
+                    break;
+                case "lenny":
+                    message="( ͡° ͜ʖ ͡°) "+message.substring(7);
+                    break;
+                default:
+                    if (command[0][0]=="/") {
+                        message = message.substring(1).trim();
+                        break;
+                    } else {
+                        this._sendError=new Error("no command name '"+command[0]+"'");
```
MidhunSureshR: You can use a template string here if you want:
```suggestion
                        this._sendError=new Error(`no command named '${command[0]}'`);
```
We may also want to mention that you can use double slash `//` if you actually intend to send a message starting with `/`

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,80 @@ export class RoomViewModel extends ViewModel {
...
+                        this._timelineError = null;
+                        this.emitChange("error");
+                        msgtype=undefined;
+                        message=undefined;
+                   }
+           }
+       }
+       return [msgtype, message];
```
MidhunSureshR: Do you think it would make more sense to return a object `{msgType, message}` rather than an array?

src/domain/session/room/RoomViewModel.js
MidhunSureshR: It would also be nice to have a button to dismiss the error.

MidhunSureshR (commented):
Looking good, leaving some comments here:

src/domain/session/room/RoomViewModel.js
Kaki-In: Why should we need to increase the verification level?

MidhunSureshR: `===` is strict equality check, see https://stackoverflow.com/a/359509

Kaki-In: Yes I know... But if the status is `"404"` instead of being `404` (and I am not sure that it is possible) what's the problem?

MidhunSureshR: It's a code quality thing i.e if you have the choice of choosing between `===` or `==`, you should always choose `===`.

Kaki-In: Ok, thanks

src/platform/web/ui/session/room/RoomView.js
```diff
@@ -46,7 +46,13 @@ export class RoomView extends TemplateView {
...
             ]),
             t.div({className: "RoomView_body"}, [
-                t.div({className: "RoomView_error"}, vm => vm.error),
+                t.div({className: "RoomView_error"}, [
+					t.if(vm => vm.error, t => t.p({}, vm => vm.error)),
+					t.if(vm => vm.error, t => t.button({
+						className: "RoomView_error_closerButton",
+						onClick: evt => vm.dismissError(evt)
+					}))
+                ]),
```
MidhunSureshR: You can use one binding:
```js
t.div({className: "RoomView_error"}, [
	t.if(vm => vm.error, t => t.div(
		[
			t.p({}, vm => vm.error),
			t.button({ className: "RoomView_error_closerButton", onClick: evt => vm.dismissError(evt) })
		]
	)),
```

src/matrix/Session.js
```diff
@@ -948,6 +948,23 @@ export class Session {
             return body.room_id;
         });
     }
+
+    waitForRoomFromSync(roomId) {
```
MidhunSureshR: Can you integrate this with the `joinRoom` method so that it returns a Promise that eventually resolves to the joined room?

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,83 @@ export class RoomViewModel extends ViewModel {
...
-                let msgtype = "m.text";
-                if (message.startsWith("/me ")) {
-                    message = message.substr(4).trim();
-                    msgtype = "m.emote";
-                }
-                if (replyingTo) {
-                    await replyingTo.reply(msgtype, message);
-                } else {
-                    await this._room.sendEvent("m.room.message", {msgtype, body: message});
+               let msgtype = messinfo.type;
```
MidhunSureshR: Can you use `const` if you don't need to redefine the variables?

MidhunSureshR (commented):
Some more comments:

src/domain/session/room/RoomViewModel.js
bwindels: Basically, we always use === unless there is a good reason to use == (which hasn't happened yet I think?), so we are less likely to get it wrong when it does matter :)

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,83 @@ export class RoomViewModel extends ViewModel {
...
+        let msgtype = "m.text";
+        if (message.startsWith("/")) {
+            const [commandName, ...args] = message.substring(1).split(" ");
+            switch (commandName) {
+                case "me":
+                   message = message.substring(4).trim();
+                   msgtype = "m.emote";
+                   break;
+               case "join":
+                   if (args.length == 1) {
```
bwindels: We strive to have methods not much longer than 15-20 lines (definitely with some exceptions); the code for handling the join command seems like a good candidate to extract in a separate method. Then you won't need the `               if (msgtype && message) {` on line 271 which on the face of it is a bit strange, why wouldn't you have a message if you just typed one?

src/matrix/Session.js
```diff
@@ -948,6 +948,23 @@ export class Session {
...
+        let resolve;
+        const promise = new Promise(r => { resolve = r; })
+        const subscription = {
```
bwindels: This is a bit of a hacky way to do it, because you're not implementing the full IMapObserver contract, and it's a bit overkill anyway. You can use `session.observeRoomStatus` instead which is meant for this use case:
```js
await session.observeRoomStatus(roomId).waitFor(status === RoomStatus.Joined);
```

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,83 @@ export class RoomViewModel extends ViewModel {
         }
     }
     
+    async _getMessageInformations (message) {
+        let msgtype = "m.text";
+        if (message.startsWith("/")) {
```
bwindels: This method is more general than it needs to be. We don't do anything currently to messages apart from processing commands. You could put this `if` in the parent method and change this method to be `_processCommand` called from within that `if`.

src/matrix/Session.js
bwindels: Given the onliner below, not sure it is worth it actually for now. It's true that the API is not super obvious, but I'd rather improve that separately.

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +197,83 @@ export class RoomViewModel extends ViewModel {
...
+            switch (commandName) {
+                case "me":
+                   message = message.substring(4).trim();
+                   msgtype = "m.emote";
+                   break;
+               case "join":
+                   if (args.length == 1) {
+                       let roomName = args[0];
+                       try {
+                           const internalId = await this._options.client.session.joinRoom(roomName);
```
bwindels: I'd just call this `roomId`, `internalId` suggests this is something else

src/platform/web/ui/session/room/RoomView.js
Kaki-In: I need to modify the theme.css then.

src/domain/session/room/RoomViewModel.js
Kaki-In: if you type /join @roomid:site.location, you don't want to send an undefined object to everybody (it really does), so I had to precise that, if the message or the message type goes to be undefined, the message won't be sent

src/matrix/Session.js
MidhunSureshR: True, resolving this in favor of https://github.com/vector-im/hydrogen-web/pull/809#discussion_r928833506

src/domain/session/room/RoomViewModel.js
MidhunSureshR: > why wouldn't you have a message if you just typed one?

There's nothing to send if an invalid command was entered, for eg: sending "/foobar" from the composer should not send any message.

Kaki-In: For /join, there is also nothing to send, even if the room id was right

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +198,83 @@ export class RoomViewModel extends ViewModel {
...
+                            this._sendError = exc;
+                        }
+                        this._timelineError = null;
+                        this.emitChange("error");
+                    }
+                } else {
+                    this._sendError = new Error("join syntax: /join <room-id>");
+                    this._timelineError = null;
+                    this.emitChange("error");
+                }
```
MidhunSureshR: Can you extract this code into a separate method?

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +198,87 @@ export class RoomViewModel extends ViewModel {
         }
     }
     
+    async executeJoinCommand(args) {
```
MidhunSureshR: It's a bit more cleaner to just pass the room id to this method.
so
```js
   case "join":
                const roomId = args[0];
                if(roomId) {
                   await this.executeJoinCommand(roomId);                
                }
                break;
```

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +198,87 @@ export class RoomViewModel extends ViewModel {
...
+        if (args.length == 1) {
+            let roomName = args[0];
+            try {
+                const roomId = await this._options.client.session.joinRoom(roomName);
+                await (await this._options.client.session.observeRoomStatus(roomId)).waitFor(status => status === RoomStatus.Joined);
```
MidhunSureshR: You don't need the double await here.

Kaki-In: I need an await to get the `this._options.client.session.observeRoomStatus(roomId)` object (because he do have the .waitFor) and an await to wait the waitFor to be ended. 

I already tried without the second await, it throws a Type Error because `this._options.client.session.observeRoomStatus(roomId)` returns a Promise.

Kaki-In: Is this code better? :  
```javascript
            const roomId = await this._options.client.session.joinRoom(roomName);
            const roomStatusObserver = await this._options.client.session.observeRoomStatus(roomId);
            await roomStatusObserver.waitFor(status => status === RoomStatus.Joined);
```

MidhunSureshR: > Is this code better? :
> 
> ```js
>             const roomId = await this._options.client.session.joinRoom(roomName);
>             const roomStatusObserver = await this._options.client.session.observeRoomStatus(roomId);
>             await roomStatusObserver.waitFor(status => status === RoomStatus.Joined);
> ```


ah yes, you do need the double await since it's not a promise that resolves to another promise. This code looks good! :+1:

Kaki-In:
I cannot send any message with this implement, it doesn't throw any error, I can see the `ensureMessageKeyIsShared` message on the console log, but nothing is sent... 
With master it works successfully. 

I often have this error : 
``` A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received```

```
:3000/#/session/1692…bGbqHQR:discu.top:1 Uncaught (in promise) {message: 'A listener indicated an asynchronous response by r…age channel closed before a response was received'}
Promise.then (async)
b @ content_script_bundle.js:100
js @ content_script_bundle.js:56
t.runContentScripts @ content_script_bundle.js:163
(anonymous) @ content_script_bundle.js:71
Promise.then (async)
t.default @ content_script_bundle.js:71
1366 @ content_script_bundle.js:9
n @ content_script_bundle.js:1
1365 @ content_script_bundle.js:1
n @ content_script_bundle.js:1
1364 @ content_script_bundle.js:1
n @ content_script_bundle.js:1
1364 @ content_script_bundle.js:1
(anonymous) @ content_script_bundle.js:1
```
I absolutely don't know why this probleme occurs, I have read all my code again and if the first conditions are not entered by a simple message, I don't understand why the message isn't sent.

File: src/domain/session/room/RoomViewModel.js [hunk 18]
```diff
@@ -197,18 +198,100 @@ export class RoomViewModel extends ViewModel {
...
+                return `room '${roomName}' not found`;
+            } else if ((exc.statusCode ?? exc.status) === 403) {
+                return `you're not invited to join '${roomName}'`;
+            } else {
+                return exc;
+            }
+        }
+    }
+    
+    async _processCommandJoin(roomName) {
```
MidhunSureshR: I think this method can be merged with `joinRoom` method.

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +198,100 @@ export class RoomViewModel extends ViewModel {
...
+    async _processCommand (message) {
+        let msgtype = undefined;
+        const [commandName, ...args] = message.substring(1).split(" ");
+        switch (commandName) {
+            case "me":
+                message = args.join(" ");
+                msgtype = "m.emote";
+                break;
+            case "join":
+                if (args.length == 1) {
```
MidhunSureshR:
```suggestion
                if (args.length === 1) {
```

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +198,100 @@ export class RoomViewModel extends ViewModel {
...
+            } else {
+                this._sendError = new Error("/join : " + exc);
+            }
+            this._timelineError = null;
+            this.emitChange("error");
+        }
+    } 
+
+    async _processCommand (message) {
+        let msgtype = undefined;
```
MidhunSureshR:
```suggestion
        let msgtype;
```

src/domain/session/room/RoomViewModel.js
```diff
@@ -197,18 +198,100 @@ export class RoomViewModel extends ViewModel {
...
+                break;
+            case "lenny":
+                message="( ͡° ͜ʖ ͡°) " + args.join(" ");
+                msgtype = "m.text";
+                break;
+            default:
+                this._sendError = new Error(`no command name "${commandName}". To send the message instead of executing, please type "/${message}"`);
+                this._timelineError = null;
+                this.emitChange("error");
+                msgtype = undefined;
```
MidhunSureshR: `msgtype` would undefined anyway, so can remove this line.

src/platform/web/ui/css/themes/element/theme.css
```diff
@@ -521,6 +521,62 @@ a {
...
+    overflow : hidden;
+    height: 100%;
+    width: 100%;
+    position : relative;
+    display : flex;
+    align-items : center;
+}
+
+.RoomView_error:not(:empty) {
+    height : 40px;
```
MidhunSureshR: Currently the errors are not fully shown in mobile devices.
```suggestion
    height : auto;
```

MidhunSureshR (approved):
Last couple of changes:

src/domain/session/room/RoomViewModel.js
bwindels: Yeah, you need the msgtype && message because some commands do want to send something, like tableflip. Missed that bit.

PR #810, opened 2022-07-21
Detect home server from matrix id

### LINKED ISSUE #479: Support .well-known in the way the spec says
Hydrogen has now implemented [Element's newish style of doing .well-known](https://github.com/matrix-org/matrix-react-sdk/pull/5474) which is great but that isn't how the spec says .well-known works. It would be good to support both methods.

The spec says .well-known works by having the user enter their full username (as in @user:server.com) and then the client sends the .well-known request to the server name that appears after the colon. https://spec.matrix.org/unstable/client-server-api/#well-known-uri

Kaki-In:
# Detect home server from matrix id

## Added functionnality : 
 - you can now enter the user matrix id (e.g. @username:server.local) in the login input. 
 - if the home server that you've entered into youre id is not the same that those into the home server input, it will be use instead to connect to the default matrix server.

Should fix [this issue (#479)](https://github.com/vector-im/hydrogen-web/issues/479).


src/domain/login/PasswordLoginViewModel.js
```diff
@@ -39,10 +41,35 @@ export class PasswordLoginViewModel extends ViewModel {
...
+	
+	setEnabled(enabled) {
+		this._enabled = enabled;
+        this.emitChange("isEnabled");
+	}
 
+    async parseUsernameLogin (login) {
+        if ((login.match(/@/g) || []).length == 1 && (login.match(/:/g) || []).length == 1 && (login.match(/\./g) || []).length) {
+            if (login.indexOf("@") == 0 && login.indexOf("@") < login.indexOf(":") && login.indexOf(":") < login.indexOf(".")) {
+				if (this._loginOptions.homeserver != login.substring(login.indexOf(":") + 1)) {await await this._options.setHomeserver(login.substring(login.indexOf(":") + 1));}
```
bwindels: You added the double await because you're now returning a promise from an async method?

src/domain/login/PasswordLoginViewModel.js
```diff
@@ -39,10 +41,35 @@ export class PasswordLoginViewModel extends ViewModel {
...
+		this._loginOptions = loginOptions;
+	}
+	
+	setEnabled(enabled) {
+		this._enabled = enabled;
+        this.emitChange("isEnabled");
+	}
 
+    async parseUsernameLogin (login) {
+        if ((login.match(/@/g) || []).length == 1 && (login.match(/:/g) || []).length == 1 && (login.match(/\./g) || []).length) {
```
bwindels: Consider using just one regex to parse the mxid and extract the server name : `@.*:(.+)`

bwindels (commented):
Just a few comments, haven't done a full review yet.

src/domain/login/PasswordLoginViewModel.js
Kaki-In: Yes, exactly. I am not sure to have to put the await inside the `setHomeServer` method.

src/domain/login/PasswordLoginViewModel.js
Kaki-In: So I replace the 55th line's condition by `login.match(/@.*:(.+)/)`?

src/platform/web/ui/login/PasswordLoginView.js
```diff
@@ -32,15 +32,15 @@ export class PasswordLoginView extends TemplateView {
             disabled
         });
         
-        return t.div({className: "PasswordLoginView form"}, [
+        return t.div({className: "PasswordLoginView form" + (disabled ? " disabled" : "")}, [
```
MidhunSureshR: You can do this without string manipulation:
```suggestion
return t.div({
	className: {
		PasswordLoginView: true,
		form: true,
		disabled
	}
}, [
```

src/platform/web/ui/login/PasswordLoginView.js
```diff
@@ -18,7 +18,7 @@ import {TemplateView} from "../general/TemplateView";
 
 export class PasswordLoginView extends TemplateView {
     render(t, vm) {
-        const disabled = vm => !!vm.isBusy;
+        const disabled = vm => (!!vm.isBusy) && (!vm.isEnabled);
```
MidhunSureshR: Not a change you made but we don't really need the double bang (`!!`) here since `vm.isBusy` is already a boolean.
So this can be:
```suggestion
        const disabled = vm => vm.isBusy && !vm.isEnabled;
```

src/domain/login/PasswordLoginViewModel.js
```diff
@@ -39,10 +41,36 @@ export class PasswordLoginViewModel extends ViewModel {
...
+    setLoginOptions(loginOptions) {
+        this._loginOptions = loginOptions;
+    }
+    
+    setEnabled(enabled) {
+        this._enabled = enabled;
+        this.emitChange("isEnabled");
+    }
 
+    async parseUsernameLogin (login) {
```
MidhunSureshR:
```suggestion
    async parseUsernameLogin (username) {
```

src/domain/login/PasswordLoginViewModel.js
MidhunSureshR: When you do `const foo = await something...`, foo will never be a promise!
See https://jsfiddle.net/e3tpmufg/

src/domain/login/PasswordLoginViewModel.js
MidhunSureshR: You can do something like:
```js
findHomeserverFromUsername (login) {
    const matches = login.match(/@.*:(.+)/);
    if (!matches) {
        return;
    }
    const [, homeserverInMxid] = matches;
    return homeserverInMxid;
}

async changeHomeserverFromUsername(username) {
    const currentHomeserver = this._loginOptions.homeserver;
    const newHomeserver = this.findHomeserverFromUsername(username);
    if (newHomeserver && currentHomeserver !== newHomeserver) {
        await this._options.setHomeserver(newHomeserver);
        return true;
    }
    return false;
}
```

src/domain/login/PasswordLoginViewModel.js
```diff
@@ -57,7 +85,13 @@ export class PasswordLoginViewModel extends ViewModel {
                 break;
         }
         if (error) {
-            this._showError(error);
+            if (usernameChanged) {
+                console.log(error);
+                this._options.showConnectionError(error);
```
MidhunSureshR: Why do we have two different ways of showing errors here?

MidhunSureshR (commented):
Some comments and questions:

src/domain/login/PasswordLoginViewModel.js
Kaki-In: Surely a miss from a previous commit

src/platform/web/ui/login/PasswordLoginView.js
```diff
@@ -18,7 +18,7 @@ import {TemplateView} from "../general/TemplateView";
 
 export class PasswordLoginView extends TemplateView {
     render(t, vm) {
-        const disabled = vm => !!vm.isBusy;
+        const disabled = vm => vm.isBusy && vm.isEnabled;
```
bwindels: This seems weird. We disable the form when we're both busy and the view model says we should be enabled rather than disabled? Do we even need the two fields?

src/matrix/common.js
```diff
@@ -22,16 +22,16 @@ export function makeTxnId() {
 }
 
 export function isTxnId(txnId) {
-	return txnId.startsWith("t") && txnId.length === 15;
+    return txnId.startsWith("t") && txnId.length === 15;
 }
 
 export function tests() {
-	return {
```
bwindels: Please don't change the indenting when not making changes.

src/platform/web/ui/login/PasswordLoginView.js
Kaki-In: Yes, it should be `vm.isBusy || !vm.isEnabled;` sorry ^^'

src/matrix/common.js
Kaki-In: Do I have to cancel those indenting changes or it should stay like that?

src/domain/login/LoginViewModel.ts
```diff
@@ -193,20 +206,17 @@ export class LoginViewModel extends ViewModel<Options> {
         );
     }
 
-    private _disposeViewModels(): void {
-        this._startSSOLoginViewModel = this.disposeTracked(this._startSSOLoginViewModel);
-        this._passwordLoginViewModel = this.disposeTracked(this._passwordLoginViewModel);
-        this._completeSSOLoginViewModel = this.disposeTracked(this._completeSSOLoginViewModel);
-        this.emitChange("disposeViewModels");
+    updateLoginOptions(): void {
```
MidhunSureshR:
```suggestion
    private _updateLoginOptions(): void {
```
Don't forget to add the "_" to where this method is called.

src/domain/login/LoginViewModel.ts
```diff
@@ -123,26 +124,38 @@ export class LoginViewModel extends ViewModel<Options> {
         this._passwordLoginViewModel = this.track(new PasswordLoginViewModel(
             this.childOptions({
                 loginOptions: this._loginOptions,
-                attemptLogin: (loginMethod: PasswordLoginMethod) => this.attemptLogin(loginMethod)
-        })));
+                attemptLogin: loginMethod => this.attemptLogin(loginMethod),
+                setHomeserver: homeserver => this.setHomeserver(homeserver)
```
MidhunSureshR:
```suggestion
                attemptLogin: (loginMethod: PasswordLoginMethod) => this.attemptLogin(loginMethod),
                setHomeserver: (homeserver: string) => this.setHomeserver(homeserver)
```

MidhunSureshR (commented):
Some more ts changes:

PR #811, opened 2022-07-22
Key sharing based on room history visibility

### LINKED ISSUE #189: Send room keys to invited members if m.room.history_visibility=invited
Read from state events when calling `roomEncryption.writeMemberChanges`

When `m.room.history_visibility` is changed, we'll need to reevaluate which roomIds should be on a userIdentity? We will need to do this anyway for the migration.

bwindels:
Fixes #189 

Before we didn't take the history visibility into account when sharing keys, we only shared the key on a join and discarded our current key when somebody left (so a new one will be created on the next message we sent).

This loads the history visibility when needed in the RoomEncryption object and updates it from sync, and passes it into the device tracker.

The device tracker writes entries in the userIdentities store when starting to track a room. The idea is that userIdentities only have the roomIds in them for the rooms where they should receive keys. Before again it just took joined members. Now we need to rewrite them when the history visibility changes, so that was added.

I've also take the decision out of the RoomEncryption object to detect when a key should be shared or discarded and delegate it to the DeviceTracker, which now returns removed and added userIds. This avoids duplication of the same logic in both classes.

The way key sharing works is by first writing an operation to the operation store, so the intent of sharing the key does not get lost if we can't send it right now.

So we share the key in these cases:
 - we are sending a message and need to rotate the key
 - some membership change happened and according to the current room history visibility the member should now receive the keys
 - the history visibility was changed and some members might now start or stop receiving the keys. If a member should no longer receive the key, we need to discard the current key so next time we create a new one.

### DIFF
--- a/src/matrix/e2ee/DeviceTracker.js
+++ b/src/matrix/e2ee/DeviceTracker.js
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 import {verifyEd25519Signature, SIGNATURE_ALGORITHM} from "./common.js";
+import {shouldShareKey} from "./common.js";
 
 const TRACKING_STATUS_OUTDATED = 0;
 const TRACKING_STATUS_UPTODATE = 1;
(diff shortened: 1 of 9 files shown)


src/matrix/e2ee/DeviceTracker.js
```diff
@@ -112,21 +143,44 @@ export class DeviceTracker {
...
-                await this._writeMember(member, txn);
-            }
-        }));
+    async writeHistoryVisibility(room, historyVisibility, syncTxn, log) {
+        const added = [];
+        const removed = [];
+        if (room.isTrackingMembers && room.isEncrypted) {
+            await log.wrap("rewriting userIdentities", async log => {
+                // don't use room.loadMemberList here because we want to use the syncTxn to load the members
+                const memberList = await room.loadMemberList(syncTxn, log);
```
MidhunSureshR: Is this comment no longer relevant?

MidhunSureshR (approved):
LGTM :+1:

PR #812, opened 2022-07-25
[Documentation] - Add type for config options

MidhunSureshR:
This PR adds a ts type for the config options supported by Hydrogen.

### DIFF
--- a/src/platform/types/config.ts
+++ b/src/platform/types/config.ts
@@ -50,6 +50,7 @@ export type Config = {
     /**
      * Configuration for push notifications.
      * See https://spec.matrix.org/latest/client-server-api/#post_matrixclientv3pushersset
+     * and https://github.com/matrix-org/sygnal/blob/main/docs/applications.md#webpush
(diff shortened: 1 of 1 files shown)


src/platform/types/config.ts
```diff
@@ -0,0 +1,63 @@
...
+     */
+    defaultTheme?: {
+        // id of light theme
+        light: string;
+        // id of dark theme
+        dark: string;
+    };
+    /**
+     * Configuration for push notifications.
+     * See https://spec.matrix.org/latest/client-server-api/#post_matrixclientv3pushersset
```
bwindels, line 52: Can you also add a link to https://github.com/matrix-org/sygnal/blob/main/docs/applications.md#webpush here? :pray:

bwindels (approved):

src/platform/types/config.ts
bwindels: Adding it myself so it can get merged :)

PR #816, opened 2022-07-26
Opening the last opened room at start

### LINKED ISSUE #435: Last location is not restored when opening the app
Often I just get the session picker screen, rather than opening the session and room I was previously in.

Kaki-In:
The session now directly opens at the last opened room when openening the Hydrogen page. 

---
Should fix [this issue (#435)](https://github.com/vector-im/hydrogen-web/issues/435)

### DIFF
--- a/src/matrix/common.js
+++ b/src/matrix/common.js
@@ -22,16 +22,16 @@ export function makeTxnId() {
 }
 
 export function isTxnId(txnId) {
-	return txnId.startsWith("t") && txnId.length === 15;
+    return txnId.startsWith("t") && txnId.length === 15;
 }
 
 export function tests() {
-	return {
-		"isTxnId succeeds on result of makeTxnId": assert => {
-			assert(isTxnId(makeTxnId()));
-		},
-		"isTxnId fails on event id": assert => {
-			assert(!isTxnId("$yS_n5n3cIO2aTtek0_2ZSlv-7g4YYR2zKrk2mFCW_rm"));
-		},
-	}
+    return {
+        "isTxnId succeeds on result of makeTxnId": assert => {
+            assert(isTxnId(makeTxnId()));
+        },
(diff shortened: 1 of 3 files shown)


bwindels:
Fixes #435

src/matrix/common.js
```diff
@@ -22,16 +22,16 @@ export function makeTxnId() {
 }
 
 export function isTxnId(txnId) {
-	return txnId.startsWith("t") && txnId.length === 15;
```
bwindels: Please avoid making unrelated whitespace changes. It makes it harder to see what is being changed.

src/platform/web/dom/History.js
```diff
@@ -65,6 +65,7 @@ export class History extends BaseObservableValue {
     }
 
     onSubscribeFirst() {
+        this._lastSessionHash = window.localStorage?.getItem("hydrogen_last_url_hash");
```
bwindels, line 74: Consider initializing _lastSessionHash to undefined in the constructor to not break [hidden classes](https://draft.li/blog/2016/12/22/javascript-engines-hidden-classes/).

bwindels (commented):
Thanks, just a few details to fix.

src/platform/web/dom/History.js
```diff
@@ -17,6 +17,12 @@ limitations under the License.
 import {BaseObservableValue} from "../../../observable/ObservableValue";
 
 export class History extends BaseObservableValue {
+    
+    constructor() {
+        super();
+        this._lastSessionHash = null;
```
bwindels: Thanks for the change, we try to use `undefined` going forward though as it's nicer to type in typescript.
```suggestion
        this._lastSessionHash = undefined;
```

bwindels (approved):
Looks good, happy to have this finally fixed!

PR #821, opened 2022-07-29
Implement emojis

Kaki-In:
# Implemented emojis.

Current emojis are  : 
 - grinning : 😀
 - smiley : 😃
 - smile : 😄
 - grin : 😁
 - laughing: 😆
 - sweat smile : 😅
 - rolling on the floor laughing : 🤣
 - joy : 😂
 - slightly smilling face : 🙂
 - upside face : 🙃
 - malting face : 🫠
 - wink : 😉. 

In the message composer, tape `joy :joy:` and then press enter into any room.


bwindels:
Hey, as discussed before, I'm afraid this UI-impacting feature would need further thinking of how we want to add it to the app, but for now this isn't heading towards something we would want to merge, mainly because it is not discoverable by the user.

PR #824, opened 2022-07-29
Fix develop server breaking due to import syntax

MidhunSureshR:


src/platform/web/theming/DerivedVariables.ts
```diff
@@ -98,8 +98,8 @@ export class DerivedVariables {
     }
 }
 
-import pkg from "off-color";
-const {offColor} = pkg;
+import * as pkg from "off-color";
+const offColor = pkg.offColor ?? pkg.default.offColor;
```
bwindels, line 103: I wonder if this would fix the tsc error?
```suggestion
const offColor = pkg.offColor ?? pkg.default?.offColor;
```

bwindels (commented):
Can we also document this in doc/ somewhere? A document how to add a dependency from npm ...

bwindels (approved):

PR #825, opened 2022-07-29
Fix sdk build failing after derived theme implementation

MidhunSureshR:
Forgot to update the vite config

### DIFF
--- a/vite.sdk-assets-config.js
+++ b/vite.sdk-assets-config.js
@@ -36,7 +36,7 @@ export default mergeOptions(commonOptions, {
     plugins: [
         themeBuilder({
             themeConfig: {
-                themes: { element: "./src/platform/web/ui/css/themes/element" },
+                themes: ["./src/platform/web/ui/css/themes/element"],
                 default: "element",
             },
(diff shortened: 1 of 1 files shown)


vite.sdk-assets-config.js
```diff
@@ -36,7 +36,7 @@ export default mergeOptions(commonOptions, {
     plugins: [
         themeBuilder({
             themeConfig: {
-                themes: { element: "./src/platform/web/ui/css/themes/element" },
+                themes: ["./src/platform/web/ui/css/themes/element"],
```
MidhunSureshR, line 39: I wonder if this should go in `vite.common-config.js` since it's common. Will have to think on Monday.

bwindels: I guess the problem then would be that it also would add the plugin for the sdk library build?

bwindels (approved):

vite.sdk-assets-config.js
MidhunSureshR: Ah yes!

PR #834, opened 2022-08-05
Gap fill from HS should halt on network disconnection

### LINKED ISSUE #627: Gap filling is tight-looped when offline and need timeline back-filling.
As is visible from the logs (lots of fillGap entries with only request time in between)

MidhunSureshR:
and should resume when connection is regained
This PR:
- Wait for network reconnection before retrying gap fill
- UI Improvements:
   - Errors can optionally show a spinner (eg when waiting for reconnection)
   - Other errors show no spinner
   - Show either "Not Loading" or the error itself but not both
   - Not Loading should not show a spinner

### DIFF
--- a/src/domain/session/room/timeline/tiles/GapTile.js
+++ b/src/domain/session/room/timeline/tiles/GapTile.js
@@ -38,6 +38,7 @@ export class GapTile extends SimpleTile {
             } catch (err) {
                 console.error(`room.fillGap(): ${err.message}:\n${err.stack}`);
                 this._error = err;
+                this._loading = false;
                 this.emitChange("error");
                 if (err instanceof ConnectionError) {
                     /*
--- a/src/domain/session/room/timeline/tiles/GapTile.js
+++ b/src/domain/session/room/timeline/tiles/GapTile.js
@@ -26,21 +26,22 @@ export class GapTile extends SimpleTile {
         this._error = null;
         this._isAtTop = true;
         this._siblingChanged = false;
+        this._showSpinner = false;
     }
 
     async fill() {
         if (!this._loading && !this._entry.edgeReached) {
             this._loading = true;
             this._error = null;
+            this._showSpinner = true;
             this.emitChange("isLoading");
             try {
                 await this._room.fillGap(this._entry, 10);
             } catch (err) {
                 console.error(`room.fillGap(): ${err.message}:\n${err.stack}`);
                 this._error = err;
-                this._loading = false;
-                this.emitChange("error");
                 if (err instanceof ConnectionError) {
+                    this.emitChange("error");
                     /*
                     We need to wait for reconnection here rather than in
                     notifyVisible() because when we return/throw here
@@ -49,13 +50,13 @@ export class GapTile extends SimpleTile {
                     resulting in multiple error entries in logs and elsewhere!
                     */
                     await this._waitForReconnection();
-                    return true;
                 }
                 // rethrow so caller of this method
                 // knows not to keep calling this for now
                 throw err;
             } finally {
                 this._loading = false;
+                this._showSpinner = false;
                 this.emitChange("isLoading");
             }
                 return true;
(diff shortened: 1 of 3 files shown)


src/domain/session/room/timeline/tiles/GapTile.js
```diff
@@ -100,8 +130,11 @@ export class GapTile extends SimpleTile {
 
     get error() {
         if (this._error) {
+            if (this._error instanceof ConnectionError) {
```
bwindels, line 139: We don't usually return anything but primitive types from view model properties to keep the view simple. How about also implementing this logic in the isLoading property and just return a string here?

src/platform/web/ui/session/room/timeline/GapView.js
```diff
@@ -29,10 +29,25 @@ export class GapView extends TemplateView {
             isLoading: vm => vm.isLoading,
             isAtTop: vm => vm.isAtTop,
         };
-        return t.li({className}, [
-            spinner(t),
-            t.div(vm => vm.isLoading ? vm.i18n`Loading more messages …` : vm.i18n`Not loading!`),
-            t.if(vm => vm.error, t => t.strong(vm => vm.error))
+        return t.li({ className }, [
```
bwindels, line 32: I think this could probably be simplified to have a binding on showing the spinner (e.g. a hidden class binding) and have a single description property on the vm which returns the message to be display (either or or normal operation) and bind to that here.

src/domain/session/room/timeline/tiles/GapTile.js
```diff
@@ -55,7 +70,18 @@ export class GapTile extends SimpleTile {
         let canFillMore;
         this._siblingChanged = false;
         do {
-            canFillMore = await this.fill();
```
bwindels, line 58: if you scroll while offline, this will retry the fill while the previous invocation of fill is still waiting for the connection to come back online. notifyVisible should probably exit early if we're waiting for the connection to come back online.

src/domain/session/room/timeline/tiles/GapTile.js
```diff
@@ -55,7 +70,18 @@ export class GapTile extends SimpleTile {
...
+            try {
+                canFillMore = await this.fill();
+            }
+            catch (e) {
+                if (e instanceof ConnectionError) {
```
bwindels, line 78: how could we ever have a connection error here if we catch that and drop the exception in `fill` already?

src/platform/web/ui/session/room/timeline/GapView.js
```diff
@@ -30,24 +30,18 @@ export class GapView extends TemplateView {
...
-                        elements = [spinner(t), t.span(vm.i18n`Loading more messages …`)];
-                    }
-                    else {
-                        elements = t.span(vm.i18n`Not loading!`);
-                    }
-                    return t.div({ className: "GapView__container" }, elements);
-                })
+            t.if(vm => vm.showSpinner, (t) => spinner(t)),
+            t.span(vm => {
+                if (vm.error) {
```
bwindels: I'd usually put this kind of logic into a property on the vm rather than in the view.

src/domain/session/room/timeline/tiles/GapTile.js
bwindels: @MidhunSureshR Could you add a word why your resolved this?

src/domain/session/room/timeline/tiles/GapTile.js
bwindels: @MidhunSureshR Could you add a word why your resolved this?

bwindels (commented):
Thanks for the changes, a few more comments.

src/domain/session/room/timeline/tiles/GapTile.js
MidhunSureshR: While we wait for reconnection, `this._isLoading` is still true, so further calls to fill method will return early.

src/domain/session/room/timeline/tiles/GapTile.js
MidhunSureshR: We rethrow the same exception after handling it?

src/platform/web/ui/session/room/timeline/GapView.js
MidhunSureshR: Resolved in https://github.com/vector-im/hydrogen-web/pull/834/commits/e6f43d6f4f32717153eda313a77a04ff970d2bbd

src/domain/session/room/timeline/tiles/GapTile.js
bwindels: Right, for some reason I thought the `throw err` was in an else case :facepalm:

bwindels (approved):

PR #836, opened 2022-08-09
Clear keys from local storage when logging out

### LINKED ISSUE #779: Remove olm account backup from localstorage after logout

MidhunSureshR:
Fixes #779

### DIFF
--- a/src/matrix/Client.js
+++ b/src/matrix/Client.js
@@ -451,6 +451,12 @@ export class Client {
 
     async deleteSession(log) {
         if (this._sessionId) {
+            await log.wrap("sessionStore", async () => {
+                const storage = this._storage ?? await this._platform.storageFactory.create(this._sessionId, log);
+                const txn = await storage.readWriteTxn([storage.storeNames.session]);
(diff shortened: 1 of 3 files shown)


src/matrix/Client.js
```diff
@@ -451,6 +451,12 @@ export class Client {
 
     async deleteSession(log) {
         if (this._sessionId) {
+            await log.wrap("sessionStore", async () => {
+                const storage = this._storage ?? await this._platform.storageFactory.create(this._sessionId, log);
+                const txn = await storage.readWriteTxn([storage.storeNames.session]);
```
bwindels: we don't actually need the txn as this will just remove localstorage rather than manipulate indexeddb. How about putting the delete method in SessionStore in a separate function (not on that class but in the same file) that takes the IDOMStorage and the database name, and you call that from StorageFactory.delete?

You will have to do the same for _localStorageKeyPrefix though, passing in the database name... and call that from delete as you do now already.

bwindels (approved):
Thanks for the changes!

PR #839, opened 2022-08-10
Improve error handling in the UI and logs

MidhunSureshR:
The following cases are handled:
- Error is shown in the UI when all the theme manifests failed to load (in addition to them being logged).
- If some manifests failed to load but at least one manifest is available, the errors are not shown in UI but all such errors are logged.
- If after parsing all the themes, no themes are available (for eg due to them all missing the "name" field) then an error saying as much is shown in the UI.
- If some themes failed to parse but at least one is available, no errors are shown in the UI although they are still logged.
- If when setting a particular theme, the css file added using the `link` element failed to load, an error is shown in the UI. This is limited to when the theme is set by `Platform.init` method.

### DIFF
--- a/src/platform/web/Platform.js
+++ b/src/platform/web/Platform.js
@@ -334,8 +334,7 @@ export class Platform {
 
     async replaceStylesheet(newPath, log) {
         const error = await this.logger.wrapOrRun(log, { l: "replaceStylesheet", location: newPath, }, async (l) => {
-            let resolve, error;
-            const promise = new Promise(r => resolve = r);
+            let error;
             const head = document.querySelector("head");
             // remove default theme 
             document.querySelectorAll(".theme").forEach(e => e.remove());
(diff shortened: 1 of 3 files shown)


src/platform/web/dom/request/fetch.js
```diff
@@ -119,10 +119,10 @@ export function createFetchRequest(createTimeout, serviceWorkerHandler) {
...
                 }
             } catch (err) {
-                // some error pages return html instead of json, ignore error
-                if (!(err.name === "SyntaxError" && status >= 400)) {
-                    throw err;
+                if (err.name === "SyntaxError" && status >= 400) {
+                    throw new ConnectionError(`${method} ${url}: Failed to fetch JSON file!`);
                 }
+                throw err;
             }
```
MidhunSureshR: For JSON files that fail to load, the previous code would fail silently without throwing.

src/platform/web/Platform.js
```diff
@@ -332,17 +332,34 @@ export class Platform {
...
-        const styleTag = document.createElement("link");
-        styleTag.href = newPath;
-        styleTag.rel = "stylesheet";
-        styleTag.type = "text/css";
-        styleTag.className = "theme";
-        head.appendChild(styleTag);
+    async replaceStylesheet(newPath, log) {
+        const error = await this.logger.wrapOrRun(log, { l: "replaceStylesheet", location: newPath, }, async (l) => {
+            let resolve, error;
+            const promise = new Promise(r => resolve = r);
```
bwindels: if you move this down to where you'll use it (e.g. just above onerror) and move those two event handlers into the promise callback where you receive `r`, it might be a bit easier to follow what is going on.

bwindels (commented):
otherwise looks good!

bwindels (approved):
Thanks!

PR #844, opened 2022-08-14
Fix undetectable race condition in open timeline

ajbura:
This PR fix the bug where [observePowerLevels promise used when creating new timeline](https://github.com/vector-im/hydrogen-web/blob/076a899b6634c55341c70dea77ffdd074d42dae4/src/matrix/room/BaseRoom.js#L550,L565) was generating the case of undetectable race condition.


robertlong:
Looks good and seems to work on my machine! Thanks for fixing this!

PR #848, opened 2022-08-18
Fix/dynamic OIDC registration

ajbura:
Adds OIDC accountManagmentUrl to session.sessionInfo so it will available in `Session`.

Map statically registered redirect uris to oidc client_id so client dynamic registration doesn't always pick single client_id for same issuer.
And moves oidc clientConfigs to app configs so we don't need to update sdk to change these configs.

### DIFF
--- a/src/platform/web/assets/config.json
+++ b/src/platform/web/assets/config.json
@@ -10,7 +10,7 @@
     "clientConfigs": {
       "https://id.thirdroom.io/realms/thirdroom/": {
         "client_id": "thirdroom",
-        "uris": ["http:localhost:3000", "https://thirdroom.io"]
+        "uris": ["http://localhost:3000", "https://thirdroom.io"]
       }
     }
   }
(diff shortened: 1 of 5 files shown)


src/platform/web/assets/config.json
```diff
@@ -5,5 +5,13 @@
     "applicationServerKey": "BC-gpSdVHEXhvHSHS0AzzWrQoukv2BE7KzpoPO_FfPacqOo3l1pdqz7rSgmB04pZCWaHPz7XRe6fjLaC-WPDopM"
   },
   "defaultHomeServer": "matrix.org",
-  "bugReportEndpointUrl": "https://element.io/bugreports/submit"
+  "bugReportEndpointUrl": "https://element.io/bugreports/submit",
+  "oidc": {
```
robertlong, line 9: We'll eventually need a way to not hardcode this. Ideally passing it in to the client constructor.

robertlong (approved):

src/platform/web/assets/config.json
ajbura: This is what i done here moved the hardcoded client config to main config.json file which is being passed to client. These are here so we can test the sdk.

Thirdroom have its own configration which is passed to sdk constructor.

PR #850, opened 2022-08-22
Perform forced logout when access token is invalidated 

### LINKED ISSUE #798: Logout automatically when access token is invalid
To support remote wiping of keys

MidhunSureshR:
When the access token is invalid, we force the user to logout. "Forced" here refers to deleting the session data without the users confirmation.

### DIFF
--- a/src/domain/ForcedLogoutViewModel.ts
+++ b/src/domain/ForcedLogoutViewModel.ts
@@ -30,6 +30,7 @@ export class ForcedLogoutViewModel extends ViewModel<SegmentType, Options> {
     constructor(options: Options) {
         super(options);
         this._sessionId = options.sessionId;
+        // Start the logout process immediately without any user interaction
         this._logoutPromise = this.forceLogout();
     }
 
@@ -40,6 +41,7 @@ export class ForcedLogoutViewModel extends ViewModel<SegmentType, Options> {
         }
         catch (err) {
             this._error = err;
+            // Show the error in the UI 
             this._showSpinner = false;
             this._showStatus = true;
             this.emitChange("error");
@@ -47,10 +49,15 @@ export class ForcedLogoutViewModel extends ViewModel<SegmentType, Options> {
     }
 
     async proceed(): Promise<void> {
+        /**
+         * The logout should already be completed because we started it from the ctor.
+         * In case the logout is still proceeding, we will show a message with a spinner. 
+         */
         this._showSpinner = true;
         this._showStatus = true;
         this.emitChange("showStatus");
         await this._logoutPromise;
+        // At this point, the logout is completed for sure.
         if (!this._error) {
             this.navigation.push("session", true);
         }
--- a/src/domain/ForcedLogoutViewModel.ts
+++ b/src/domain/ForcedLogoutViewModel.ts
@@ -1,0 +1,55 @@
+/*
+Copyright 2021 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


src/platform/web/ui/ForcedLogoutView.js
```diff
@@ -0,0 +1,46 @@
...
+                t.div({ className: "button-row" }, [
+                    t.button({
+                        className: "button-action primary",
+                        type: "submit",
+                        onClick: () => vm.proceed(),
+                    }, vm.i18n`Proceed`)
+                ]),
+            ]);
+        });
+        const progressView = new InlineTemplateView(vm, t => {
```
bwindels: you can just use `t.map(vm => vm.showStatus, (showStatus, t, vm) => {})` below.

src/domain/ForcedLogoutViewModel.ts
```diff
@@ -0,0 +1,74 @@
...
+    private _showSpinner: boolean = false;
+
+    constructor(options: Options) {
+        super(options);
+        this._sessionId = options.sessionId;
+        this._logoutPromise = this.forceLogout();
+    }
+
+    async forceLogout(): Promise<void> {
+        try {
```
bwindels, line 38: shouldn't we show the spinner and status here? They seem to default to not being shown.

src/domain/ForcedLogoutViewModel.ts
```diff
@@ -0,0 +1,74 @@
...
+        }
+        catch (err) {
+            this._error = err;
+            this._showSpinner = false;
+            this._showStatus = true;
+            this.emitChange("error");
+        }
+    }
+
+    async proceed(): Promise<void> {
```
bwindels, line 51: this method is basically to proceed to the login/session picker screen once the logout has done its thing, right? I'm unsure why you would want to show the status and or spinner here. Either the logout is done already and this will just be changing the url hash, or either it is still in progress and we should probably not do anything here (e.g. by checking that showSpinner is true, but for that we'd have to set it initially in the method above as commented there). Perhaps I'm misunderstanding how this view model is supposed to be used; I also expected the proceed button to be hidden while we're logging out (e.g. showing a spinner).

src/domain/ForcedLogoutViewModel.ts
```diff
@@ -0,0 +1,74 @@
...
+            this.navigation.push("session", true);
+        }
+    }
+
+    get status(): string {
+        if (this._error) {
+            return this.i18n`Could not log out of device: ${this._error.message}`;
+        } else {
+            return this.i18n`Logging out… Please don't close the app.`;
+        }
```
bwindels, line 71: Should we have another message here saying we're done with logout, you can now proceed to reauthenticate?

src/domain/ForcedLogoutViewModel.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    async proceed(): Promise<void> {
+        this._showSpinner = true;
+        this._showStatus = true;
+        this.emitChange("showStatus");
+        await this._logoutPromise;
+        if (!this._error) {
+            this.navigation.push("session", true);
```
bwindels: I wonder if we should redirect to the login screen here? Even if you have another session, it's probably not for the same user that you wanted to work with now anyway?

src/platform/web/ui/ForcedLogoutView.js
```diff
@@ -0,0 +1,46 @@
...
+import {TemplateView, InlineTemplateView} from "./general/TemplateView";
+import {spinner} from "./common.js";
+
+export class ForcedLogoutView extends TemplateView {
+    render(t, vm) {
+        const proceedView = new InlineTemplateView(vm, t => {
+            return t.div([
+                t.p("Your access token is no longer valid! You can reauthenticate in the next screen."),
+                t.div({ className: "button-row" }, [
+                    t.button({
```
bwindels: Why show the button while logout is still in progress? It's probably lightning fast to delete the cached data so you won't notice, but in case it is slow on some machine, it wouldn't make much sense?

bwindels (approved):

PR #860, opened 2022-08-30
Find relations to remote events within the new remote events

MadLittleMods:
Find relations to remote events within the new remote events. This allows the reactions to appear in the [Matrix public archive](https://github.com/matrix-org/matrix-public-archive) since we provide one big list of events to display.

### Dev notes

```
yarn impunity --entry-point src/matrix/room/timeline/Timeline.js --force-esm-dirs lib/ src/
```

Can't run a single test: https://github.com/bwindels/impunity/issues/17

### DIFF
--- a/src/matrix/room/timeline/Timeline.js
+++ b/src/matrix/room/timeline/Timeline.js
@@ -198,8 +198,10 @@ export class Timeline {
         if (!this._localEntries?.hasSubscriptions) {
             return;
         }
-        // find any local relations to this new remote event
-        for (const pee of this._localEntries) {
+        // find any local relations to these new remote events or maybe these
+        // new remote events reference one of the other new remote events we have.
+        const entryList = new ConcatList(entries, this._localEntries);
+        for (const pee of entryList) {
             // this will work because we set relatedEventId when removing remote echos
             if (pee.relatedEventId) {
                 const relationTarget = entries.find(e => e.id === pee.relatedEventId);
--- a/src/matrix/room/timeline/Timeline.js
+++ b/src/matrix/room/timeline/Timeline.js
@@ -666,7 +666,7 @@ export function tests() {
                 relatedEventId: reactionEntry.id
             }}));
             await poll(() => timeline.entries.length >= 3);
-            assert.equal(messageEntry.pendingAnnotations.get("👋").count, -1);
+            assert.equal(messageEntry.pendingAnnotations.get("👋").count, 0);
         },
         "local reaction gets applied after remote echo is added to timeline": async assert => {
             const messageEntry = new EventEntry({event: withTextBody("hi bob!", withSender(alice, createEvent("m.room.message", "!abc"))),


src/matrix/room/timeline/Timeline.js
```diff
@@ -198,8 +198,10 @@ export class Timeline {
         if (!this._localEntries?.hasSubscriptions) {
             return;
         }
-        // find any local relations to this new remote event
-        for (const pee of this._localEntries) {
+        // find any local relations to these new remote events or maybe these
+        // new remote events reference one of the other new remote events we have.
+        const entryList = new ConcatList(entries, this._localEntries);
+        for (const pee of entryList) {
```
MadLittleMods, line 204: Failure in CI, https://github.com/vector-im/hydrogen-web/runs/8104935984?check_suite_focus=true#step:7:167

```
 * src::matrix::room::timeline::Timeline::remove remote reaction ... failed
no persisted storage, database can be evicted by browser
AssertionError [ERR_ASSERTION]: 0 == -1
    at remove remote reaction (file:///home/runner/work/hydrogen-web/hydrogen-web/src/matrix/room/timeline/Timeline.js:669:20)
    at async runTests (file:///home/runner/work/hydrogen-web/hydrogen-web/node_modules/impunity/src/runner.mjs:44:5)
    at async main (file:///home/runner/work/hydrogen-web/hydrogen-web/node_modules/impunity/src/main.mjs:21:19)
```

src/matrix/room/timeline/Timeline.js
```diff
@@ -664,7 +666,7 @@ export function tests() {
                 relatedEventId: reactionEntry.id
             }}));
             await poll(() => timeline.entries.length >= 3);
-            assert.equal(messageEntry.pendingAnnotations.get("👋").count, -1);
+            assert.equal(messageEntry.pendingAnnotations.get("👋").count, 0);
```
MadLittleMods, line 669: AFAICT, the important part of this test is that there are no relations on the message which both `0` or `-1` indicates.

The negative `pendingAnnotations` seems a bit dubious to me but that's a separate issue. Do we really want them to ever be negative?

src/matrix/room/timeline/Timeline.js
MadLittleMods: Fixed by adjust the value, see https://github.com/vector-im/hydrogen-web/pull/860#discussion_r959095048

src/matrix/room/timeline/Timeline.js
bwindels: Yeah, the [total annotation count](https://github.com/vector-im/hydrogen-web/blob/master/src/domain/session/room/timeline/ReactionsViewModel.js#L116) is calculated by the remote count + the local, pending count. When removing a remote annotation, the remote count stays 1 until we receive the remote echo. During this time, the pending count is -1 so the total count is 0.

bwindels (changes requested):
As discussed on our pairing session on Friday, I think the http mock approach is better, so if that works out as we prototyped, we won't need this PR anymore.

bwindels:
I'll go ahead and close this PR, given we seem to be committed to the http mock approach.

PR #861, opened 2022-08-30
Clarify that we no-op because if it's already aborted, that's what we wanted to do anyway

MadLittleMods:
Clarify that we no-op because if it's already aborted, that's what we wanted to do anyway


bwindels (approved):

MadLittleMods:
Thanks for the review @bwindels 🦟

PR #862, opened 2022-08-30
Remove unused conflicting generic styles

MadLittleMods:
Remove unused conflicting generic styles.

These styles were conflicting with some in the Matrix public archive and instead of adjusting in that project, they seem unused here anyway so we can simplify and remove here.

For reference, the style here conflicts with the `.room-header-change-dates-button` styles in https://github.com/matrix-org/matrix-public-archive/pull/53

### DIFF
--- a/src/platform/web/ui/css/room.css
+++ b/src/platform/web/ui/css/room.css
@@ -23,10 +23,6 @@ limitations under the License.
     flex: 1;
 }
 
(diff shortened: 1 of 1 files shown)


bwindels (approved):

MadLittleMods:
Thanks for the review @bwindels 🦍

PR #863, opened 2022-08-31
Make no `content` PL fallback to `0`

MadLittleMods:
Make no `content` PL fallback to `0`. A valid PL event will have `content` but when you're just stubbing a PL event with an empty object `{}`, it doesn't need to exist and we can just as easily fallback to `0` PL level.

Changes necessary for https://github.com/matrix-org/matrix-public-archive/pull/57


bwindels (approved):

MadLittleMods:
Thanks for the review @bwindels 🐤

PR #873, opened 2022-09-16
Log errors when mounting views to the console for easier reference

MadLittleMods:
Also log errors when mounting views to the console for easier reference instead of just having it in the DOM at the component error boundary.

From the console, I can click the source references in the stack trace to jump to the spot in the code where things are going wrong. Looking a the console also gives an easy way to check for errors anywhere on the page.

It also helps with the problem of the error not having enough space to be read in some components since it overflows off-screen.

Error overflowing the component and hard to read

---


MidhunSureshR (approved):

MadLittleMods:
Thanks for the review @MidhunSureshR 🦌

PR #880, opened 2022-09-26
Implement TURN

ara4n:
Call the /voip/turnServer API every TTL seconds to ensure our calls use the right TURN server, rather than the hardcoded default stun:turn.matrix.org.  Otherwise we will never be able to set up relay candidates, and never be able to traverse firewalls reliably.

### DIFF
--- a/src/matrix/Session.js
+++ b/src/matrix/Session.js
@@ -561,14 +561,14 @@ export class Session {
     }
 
     async _updateTurnServers() {
-        const turnServersData = await this._hsApi.getTurnServers().response();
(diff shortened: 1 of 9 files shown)


ara4n:
i’m not sure this catches all the places where optiond.turnServers should be set, as Travis had no relay candidates when i tested with him. I did though, so the call connected

bwindels:
Have rewritten things a bit, to use ObservableValues, do the polling only when necessary, and to cleanup properly when the session gets disposed.

bwindels (approved):

PR #851, opened 2022-08-25
Typescript conversion for `IGridItemViewModel`

ibeckermayer:
Conversion for all the objects that according to the README are meant to extend `IGridItemViewModel`

PR #885, opened 2022-10-06
Fix sync error

### LINKED ISSUE #866: Sync failed because of TypeError: Cannot read properties of null (reading 'emitChange')
happened while I was trying to send messages during catchup sync right after startup. A restart fixed the problem without data loss.

MidhunSureshR:
Fixes #866 

For a non-archived room, the composer vm is created via an async method (`async _recreateComposerOnPowerLevelChange()`) that is called from the ctor. It is possible that a room change event occurs before this async method finishes which would call `emitChange` on a composer that is yet to be created.


MidhunSureshR:
https://github.com/vector-im/hydrogen-web/pull/851 already includes this change.

bwindels (approved):

PR #889, opened 2022-10-11
Add abiity to setup session immediately after registration without using /login

### LINKED ISSUE #888: Ability to login via registration 200 response without having to use /login again
After successful registration, we should already have the access_token in the 200 response. Using `/login` with the username/password after registration is redundant.

MidhunSureshR:
Fixes #888

### DIFF
--- a/src/matrix/Client.js
+++ b/src/matrix/Client.js
@@ -200,40 +200,38 @@ export class Client {
     }
 
     async _createSessionAfterAuth({deviceId, userId, accessToken, homeserver}, inspectAccountSetup, log) {
-        await log.wrap("_createSessionAfterAuth", async (l) => {
-            const id = this.createNewSessionId();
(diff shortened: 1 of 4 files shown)


src/matrix/Client.js
```diff
@@ -191,9 +200,26 @@ export class Client {
                 }
                 return;
             }
+            await this.startWithSessionInfo(sessionInfo, inspectAccountSetup, log);
+        });
+    }
+
+    async startWithSessionInfo({deviceId, userId, accessToken, homeserver}, inspectAccountSetup, log) {
```
bwindels: perhaps call this `_createSessionAfterAuth` or something? Would definitely mark it private with `_`.

bwindels (approved):
just one comment, otherwise looks good!

PR #890, opened 2022-10-12
Typescript conversion for `leftpanel` and `rightpanel`

ibeckermayer:
Divergent git histories have made it a bit of a mind bender to try and create fork PR's. On my fork, this branch appears to have gotten the changes for converting `leftpanel` and `rightpanel` to typescript. 

If you're ok with it: we can merge all these changes at once (via a squash and merge). It will be a big commit, but since it's just a typescript conversion, and so nothing is likely to break and therefore need to be reverted in a more high-resolution manner, I think its acceptable. However its not my repo, and if you insist that I go in and preform some git surgery then I will do so.


ibeckermayer:
> If you're ok with it: we can merge all these changes at once (via a squash and merge). It will be a big commit, but since it's just a typescript conversion, and so nothing is likely to break and therefore need to be reverted in a more high-resolution manner, I think its acceptable. However its not my repo, and if you insist that I go in and preform some git surgery then I will do so.

Scratch that, I discovered a relatively simple way to correct this. Closing this and will re-open separate PR's for `leftpanel` and `rightpanel` later today.

PR #907, opened 2022-10-18
Fix isMemberExpired check and typechecking

robertlong:
If the expires at key is undefined the member should be ignored / disconnected. Also fixed some typechecking errors that happen on `yarn build:sdk`


bwindels (approved):
Thanks, looks good to me!

PR #913, opened 2022-10-28
Verify unfetched roomkey senders

bwindels:


bwindels:
Should be somewhat complete, but needs testing

bwindels:
after testing:
 - [x] sync afterSyncCompleted gets triggered but doesn't fetch the keys
 - [x] updating the UI does not work

PR #915, opened 2022-11-02
Add a way to toggle `scrollToBottom` so you can start scroll from the top as well and continue reading

MadLittleMods:
Add a way to toggle `scrollToBottom` so you can start scroll from the top as well and continue reading.

See `?continue=top` in https://github.com/matrix-org/matrix-public-archive/pull/114 for how this is specifically used. When jumping to the next page of activity, you want to start from the top to continue reading where you left off.

### Future notes

For Matrix Public Archive we want:

 - Scrolled to bottom (normal or when jumping to previous page of activity)
 - Scrolled to top (when jumping to next page of activity), https://github.com/matrix-org/matrix-public-archive/pull/114
 - Event highlighted and focused in the middle, https://github.com/matrix-org/matrix-public-archive/issues/4
 - Event with date header scroll aligned with top of viewport, https://github.com/matrix-org/matrix-public-archive/issues/73

 
 ### Potential alternatives

Add a way to disable all scroll controls from Hydrogen itself and have the downstream API consumer control everything.

I could see Hydrogen supporting most of those scroll behaviors though (except scrolled to top).

### DIFF
--- a/src/platform/web/ui/session/room/TimelineView.ts
+++ b/src/platform/web/ui/session/room/TimelineView.ts
@@ -57,6 +57,10 @@ function findFirstNodeIndexAtOrBelow(tiles: HTMLElement, top: number, startIndex
     return 0;
 }
 
+interface TimelineViewOpts {
+    viewClassForTile: ViewClassForEntryFn
+    stickToBottom: boolean
+}
 export class TimelineView extends TemplateView<TimelineViewModel> {
 
     private anchoredNode?: HTMLElement;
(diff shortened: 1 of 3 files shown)


src/platform/web/ui/session/room/TimelineView.ts
```diff
@@ -57,16 +57,27 @@ function findFirstNodeIndexAtOrBelow(tiles: HTMLElement, top: number, startIndex
...
     private stickToBottom: boolean = true;
     private tilesView?: TilesListView;
     private resizeObserver?: ResizeObserver;
+    private viewClassForTile: ViewClassForEntryFn;
 
-    constructor(vm: TimelineViewModel, private readonly viewClassForTile: ViewClassForEntryFn) {
+    constructor(vm: TimelineViewModel, {
+        viewClassForTile,
+        stickToBottom = true
+    }: TimelineViewOpts) {
```
MadLittleMods, line 76: I've made this an options object because I forsee more options here (see future notes in PR description)

src/platform/web/ui/session/room/TimelineView.ts
```diff
@@ -57,16 +57,27 @@ function findFirstNodeIndexAtOrBelow(tiles: HTMLElement, top: number, startIndex
     return 0;
 }
```
MadLittleMods, line 59: Closing (alternative incoming)

PR #917, opened 2022-11-09
Add full MXID tooltip to message avatar

MadLittleMods:
Add full MXID tooltip to message avatar to be able to disambiguate users without opening the member in the right-panel


bwindels (approved):
makes sense, thanks!

MadLittleMods:
Thanks for the review and merge @bwindels 🐟

Issue #677, opened 2022-02-17
URL generation/routing is only compatible with Hydrogen itself


MadLittleMods:
## Problem

If you use `hydrogen-view-sdk` to embed some Hydrogen pieces in your own app, the URL's throughout the Hydrogen UI only use a specific routing scheme and can't be changed.

Even if you supply your own `urlCreator`/`urlRouter` and `Navigation`, that only controls the base part of the URL and the rest of the path is just string interpolated on top, ex.

https://github.com/vector-im/hydrogen-web/blob/d9acc831826c14f311ff65bdb57756584448aab7/src/domain/session/room/timeline/tiles/BaseMessageTile.js#L61

## Expected

In regards to my use case with https://github.com/matrix-org/matrix-public-archive

 - For usernames/avatars: Either `matrix.to` or completely remove the link because I'm not interested in having a right-toolbar show the member. 
 - For the message timestamp: I want to replace them with `matrix.to` links or my own scheme for permalinking and highlighting. Also want to change the timezone to UTC but that's a slightly different problem.
 - For the lightbox: ~~Remove URL and pop the lightbox with no URL change~~
    - Solved via https://github.com/matrix-org/matrix-public-archive/pull/12 and https://github.com/vector-im/hydrogen-web/pull/749

## Potential solutions

To be determined.

MadLittleMods:
*Spawning from a quick call with @bwindels. Discussed some ideas but our plan is for me to propose something and @bwindels can put some more thought and taste-making to make it Hydrogen.*

---

We could change the signature of `URLRouter` (also known as `urlCreator` in the views) to have a method for each type of URL (`urlCreator.memberLink(matrixId)`, `urlCreator.messagePermalink(matrixId)`) similar to what we are already doing in [`matrix-public-archive` for its own links](https://github.com/matrix-org/matrix-public-archive/blob/940c73868ff53422cc71b47ca577aa5d35a8d317/shared/lib/url-creator.js#L10-L16). This allows SDK consumers to define their own `URLRouter` and override the methods they want to change throughout the app.

This doesn't solve all cases where you want to contextually replace some links but not others based on the component they are in or even the parent context (think avatar re-used in many different places). Say I want to replace the member links in the timeline, but not pills for example. We could introduce an extra `context` parameter which is just a string enum of discrete UI locations, `urlCreator.memberLink(matrixId, { context: UILocation.Pill })`

Better idea please 🙏 

---

[...]

PR #918, opened 2022-11-09
Rename `urlRouter` usage to `urlRouter` instead of `urlCreator`

MadLittleMods:
Rename `urlRouter` usage to `urlRouter` instead of `urlCreator`.

This is an initial step to https://github.com/vector-im/hydrogen-web/issues/677 where we probably want to introduce a separate `URLCreator` which will be its own thing.

### DIFF
--- a/doc/SDK.md
+++ b/doc/SDK.md
@@ -85,7 +85,7 @@ async function main() {
             room,
             ownUserId: session.userId,
             platform,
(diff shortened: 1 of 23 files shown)


bwindels (approved):
looks good, thanks!

MadLittleMods:
Thanks for the review and merge @bwindels 🐮

PR #921, opened 2022-11-11
Add "Copy `matrix.to` permalink" message action

MadLittleMods:
Add "Copy `matrix.to` permalink" message action

---

Part of https://github.com/matrix-org/matrix-public-archive/issues/2


src/platform/web/dom/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+            element.removeEventListener("error", handleError);
+        };
+        element.addEventListener(successEvent, handleSuccess);
+        element.addEventListener("error", handleError);
+    });
+}
+
+// Copies the given text to clipboard and returns a boolean of whether the action was
+// successful
+export async function copyPlaintext(text: string): Promise<boolean> {
```
MadLittleMods, line 39: Based on copy text utility from `matrix-react-sdk`, see https://github.com/matrix-org/matrix-react-sdk/blob/ee13e23b156fbad9369d6a656c827b6444343d4f/src/utils/strings.ts#L25-L56

src/matrix/room/PowerLevels.js
```diff
@@ -66,10 +66,11 @@ export class PowerLevels {
 
     /** @param {string} action either "invite", "kick", "ban" or "redact". */
     _getActionLevel(action) {
-        const level = this._plEvent?.content[action];
+        const level = this._plEvent?.content?.[action];
         if (typeof level === "number") {
             return level;
         } else {
+            // TODO: Why does this default to 50?
             return 50;
```
MadLittleMods: Why does this default to 50?

Seems like the default powerlevel should be `0` if we have no frame of reference.

src/platform/web/ui/session/room/timeline/BaseMessageView.js
```diff
@@ -120,6 +120,7 @@ export class BaseMessageView extends TemplateView {
         } else if (vm.canRedact) {
             options.push(Menu.option(vm.i18n`Delete`, () => vm.redact()).setDestructive());
         }
+        options.push(Menu.option(vm.i18n`Copy permalink`, () => vm.copyPermalink()));
```
MadLittleMods: Should it say `matrix.to` in it?

```suggestion
        options.push(Menu.option(vm.i18n`Copy matrix.to permalink`, () => vm.copyPermalink()));
```

MadLittleMods: I reckon it should say "Copy `matrix.to` permalink". Updating to this ✅

src/matrix/room/PowerLevels.js
bwindels: See https://spec.matrix.org/latest/client-server-api/#mroompower_levels, `kick`, `ban` and `redact` default to `50` rather than `0`. E.g. by default you need to be a moderator to be able to do these actions.

src/domain/session/room/timeline/tiles/BaseMessageTile.js
```diff
@@ -45,6 +47,10 @@ export class BaseMessageTile extends SimpleTile {
         return `https://matrix.to/#/${encodeURIComponent(this._room.id)}/${encodeURIComponent(this._entry.id)}`;
     }
 
+    copyPermalink() {
+        copyPlaintext(this.permaLink);
```
bwindels: Since this code is in the domain layer, we can't import DOM code straight away here. Instead, you can expose the copy function on the platform object that is available here through `this.platform`.

bwindels (commented):
Apart from exposing copyPlaintext through platform, lgtm. Thanks!

src/matrix/room/PowerLevels.js
MadLittleMods: Thanks @bwindels, I've updated this magic number to be a constant `STATE_DEFAULT_POWER_LEVEL`

bwindels (approved):
thanks, lgtm!

MadLittleMods:
Thanks for the review and merge @bwindels 👽

PR #922, opened 2022-11-11
Add full MXID tooltip to message sender display name

MadLittleMods:
Add full MXID tooltip to message sender display name to disambiguate users without opening the member in the right-panel

Follow-up to https://github.com/vector-im/hydrogen-web/pull/917


bwindels (approved):

MadLittleMods:
Thanks for the review and merge @bwindels 🐦

PR #925, opened 2022-11-13
typescriptifying `RootViewModel`

ibeckermayer:


src/domain/RootViewModel.ts
```diff
@@ -117,9 +118,10 @@ export class RootViewModel extends ViewModel {
         }
     }
 
-    _showLogin(loginToken) {
+    _showLogin(loginToken?: string) {
         this._setSection(() => {
             this._loginViewModel = new LoginViewModel(this.childOptions({
+                ...this.options,
```
ibeckermayer, line 124: I can't for the life of me figure out why this is required, but typescript complains otherwise. Let me know if anybody else understands it...

PR #926, opened 2022-11-13
Typescriptifying `Client`

ibeckermayer:


src/matrix/Client.ts
```diff
@@ -123,24 +140,31 @@ export class Client {
...
 
-    queryLogin(homeserver) {
+    queryLogin(homeserver: string): AbortableOperation<Promise<LoginOptions>, void> {
         return new AbortableOperation(async setAbortable => {
             homeserver = await lookupHomeserver(homeserver, (url, options) => {
                 return setAbortable(this._platform.request(url, options));
             });
             const hsApi = new HomeServerApi({homeserver, request: this._platform.request});
-            const response = await setAbortable(hsApi.getLoginFlows()).response();
+            const response = await (setAbortable(await hsApi.getLoginFlows()) as IHomeServerRequest).response();
```
ibeckermayer, line 149: I think I get what you're trying to do with `setAbortable`'s typing, however typescript doesn't want to work that way. `SetAbortableFn` looks like:

```typescript
export type SetAbortableFn = (a: IAbortable) => typeof a;
```

which typescript interprets to mean 

```typescript
export type SetAbortableFn = (a: IAbortable) => IAbortable;
```

even though what you're really trying to say is more like

```typescript
export type SetAbortableFn = (a: IAbortable) => extends IAbortable;
```

I don't know if there's a way to express this in typescript without using generics, however then you're limited to passing one type of object into `SetAbortableFn` when really you want to be able to pass multiple different `IAbortable`s.

src/matrix/Client.ts
```diff
@@ -485,33 +514,51 @@ export class Client {
...
-        this._loginFailure = null;
+        this._error = undefined;
+        this._loginFailure = undefined;
     }
 }
 
 class AccountSetup {
-    constructor(encryptedDehydratedDevice, finishStage) {
+    _encryptedDehydratedDevice: EncryptedDehydratedDevice;
+    _dehydratedDevice?: any;
```
ibeckermayer, line 533: This appears to only get set by `finish(dehydratedDevice: any)` down below, but afaict `finish` never gets called. Maybe worth a cleanup here, but I don't understand the intention well enough to do so myself.

psrpinto:
Would be cool to get this merged. @ibeckermayer would you have availability to fix merge conflicts? If not, I would be up for doing it myself, but don't think I can push to this branch (I'm not a maintainer). I guess I could open a new PR but wanted to check here first.

PR #938, opened 2022-11-21
Add date headers in timeline (second stab)

### LINKED ISSUE #489: Add date separator in timeline

bwindels:
Fixes https://github.com/vector-im/hydrogen-web/issues/489

This is the second stab at this, with proper support in the model for date separators. The first stab at it was #764 .
This approach looks more promising, as the code is cleaner and there don't seem to be any jumps in the timeline when scrolling. Needs cleanup and more testing (e.g. how it responds to local echo when sending a message).

The sticky part is just css, but to be able to have transparent and/or variable width date headers, we'll need a JS-based approach using intersectionobserver.

What it looks like for now:

[Screencast from 18-11-22 17:17:00.webm](https://user-images.githubusercontent.com/274386/203011179-e5549347-32ad-4869-9c7e-9d67a355052e.webm)

### DIFF
--- a/src/platform/web/dom/TimeFormatter.ts
+++ b/src/platform/web/dom/TimeFormatter.ts
@@ -1,0 +1,73 @@
+/*
+Copyright 2022 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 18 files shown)


src/platform/web/dom/TimeFormatter.ts
```diff
@@ -0,0 +1,83 @@
...
+export class TimeFormatter implements ITimeFormatter {
+
+    private todayMidnight: Date;
+    private relativeDayFormatter: Intl.RelativeTimeFormat;
+    private weekdayFormatter: Intl.DateTimeFormat;
+    private currentYearFormatter: Intl.DateTimeFormat;
+    private otherYearFormatter: Intl.DateTimeFormat;
+    private timeFormatter: Intl.DateTimeFormat;
+
+    constructor(private clock: Clock) {
```
ibeckermayer, line 34: It looks like this `clock` is never used

PR #939, opened 2022-11-25
Fix '/' in room-id or mxid breaking navigation

### LINKED ISSUE #451: Can't open rooms with slashes in the ID
**To Reproduce**
1. Join a room whose room ID contains a slash (`/`), e.g. [#test:maunium.net](https://matrix.to/#/#test:maunium.net) (room ID `!test/v7 🐈️:maunium.net`)
2. Try to open the room in Hydrogen
3. Hydrogen says "You are not currently in !test. Want to join it?"

**Expected behavior**
The room is opened and the user can send messages

**Screenshots**
image

**Smartphone (please complete the following information):**
 - Device: Nokia 8.3 5G
 - OS: Android 11
 - Browser: Firefox 91.0.0-beta.6
 - Version: 0.2.5

MidhunSureshR:
Fixes #451

### DIFF
--- a/src/domain/navigation/index.ts
+++ b/src/domain/navigation/index.ts
@@ -137,7 +137,7 @@ export function parseUrlPath(urlPath: string, currentNavPath: Path<SegmentType>,
         if (type === "rooms") {
             const roomsValue = iterator.next().value;
             if (roomsValue === undefined) { break; }
-            const roomIds = roomsValue.split(",");
+            const roomIds = decodeURIComponent(roomsValue).split(",");
             segments.push(new Segment(type, roomIds));
             const selectedIndex = parseInt(iterator.next().value || "0", 10);
             const roomId = roomIds[selectedIndex];
--- a/src/domain/navigation/index.ts
+++ b/src/domain/navigation/index.ts
@@ -147,7 +147,7 @@ export function parseUrlPath(urlPath: string, currentNavPath: Path<SegmentType>,
                 segments.push(new Segment("empty-grid-tile", selectedIndex));
             }
         } else if (type === "open-room") {
-            const roomId = iterator.next().value;
+            const roomId = decodeURIComponent(iterator.next().value);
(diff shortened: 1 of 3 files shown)


src/domain/navigation/index.ts
```diff
@@ -176,7 +176,7 @@ export function parseUrlPath(urlPath: string, currentNavPath: Path<SegmentType>,
         } else if (type === "details" || type === "members") {
             pushRightPanelSegment(segments, type);
         } else if (type === "member") {
-            const userId = iterator.next().value;
+            const userId = decodeURIComponent(iterator.next().value);
```
bwindels: `userId` can be `undefined` here, which `decodeURIComponent` helpfully converts into the truthy `"undefined"` value, so should convert after this check.

src/domain/navigation/index.ts
```diff
@@ -147,7 +147,7 @@ export function parseUrlPath(urlPath: string, currentNavPath: Path<SegmentType>,
                 segments.push(new Segment("empty-grid-tile", selectedIndex));
             }
         } else if (type === "open-room") {
-            const roomId = iterator.next().value;
+            const roomId = decodeURIComponent(iterator.next().value);
```
bwindels: same here as for `userId`

src/domain/navigation/index.ts
```diff
@@ -217,15 +222,28 @@ export function stringifyPath(path: Path<SegmentType>): string {
...
     return urlPath;
 }
 
+function encodeSegmentValue(value: SegmentType[keyof SegmentType]) {
+    if (typeof value === "boolean") {
+        // Nothing to encode for boolean
+        return value;
+    }
+    else if (Array.isArray(value)) {
+        return value.map(v => encodeURIComponent(v));
```
bwindels: could do the `.join(",")` straight here, as it's part of the encoding it to a string process...

PR #948, opened 2022-12-12
Switch to Docker BuildKit and build both amd64 and arm64 images

mattcen:
I've [tested this change on my own fork](https://github.com/mattcen/hydrogen-web/actions/runs/3671935542/jobs/6207636268) and confirmed that I can both build and then use the [resulting arm64 and amd64 images](https://github.com/mattcen/hydrogen-web/pkgs/container/hydrogen-web/58282471?tag=master).

Given how long building node stuff takes on single-board computers, this would be a really beneficial change.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -36,9 +36,16 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
       - name: Build and push Docker image
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v3
         with:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -36,9 +36,16 @@ jobs:
...
+        uses: docker/setup-buildx-action@v2
+
       - name: Build and push Docker image
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v3
         with:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          platforms: linux/amd64,linux/arm64
```
hughns:
```suggestion
          platforms: linux/amd64,linux/arm64,linux/arm/v7
```

.github/workflows/docker-publish.yml
```diff
@@ -36,9 +36,16 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
```
sandhose, line 40: In #655 I originally adapted the Dockerfile to properly cross-build instead of relying on (slow) qemu emulation.

This is done by specifying `--platform=${BUILDPLATFORM}` in the `FROM` statement of the builder stage: https://github.com/sandhose/hydrogen-web/blob/6ebb058e59b3dc75fb1c19ca388a59479e1ddfdf/Dockerfile#L1

More on that here: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

mattcen: Yes, that all makes sense. I'll add `--platform=` lines to the builder and final stages in the Dockerfile when I get a chance to test that it behaves as expected. Thanks for the suggestion.

hughns:
Closing as I've gone ahead and implemented in https://github.com/vector-im/hydrogen-web/pull/996.

PR #961, opened 2022-12-15
Make sure fonts are cached by service worker

### LINKED ISSUE #763: Inter font doesn't load when offline (or system hasn't realized it is offline, but network isn't working, or very very slow)
I thought we cached the font in the service worker? perhaps this is failing or we have the wrong caching strategy and it first tries to load from the network rather than first from the cache?

psrpinto:
Fixes #763 

On Chrome, the `v=3.13` query parameter appears to prevent the file from being cached by the service worker. On Firefox it works even with query parameter.

**Before:**

Screenshot 2022-12-15 at 15 23 55

**After:**

Screenshot 2022-12-15 at 15 24 41

### DIFF
--- a/src/platform/web/ui/css/themes/element/inter.css
+++ b/src/platform/web/ui/css/themes/element/inter.css
@@ -3,16 +3,16 @@
   font-style:  normal;
   font-weight: 100;
   font-display: swap;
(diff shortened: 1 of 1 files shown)


bwindels (approved):
Thanks for looking into this. Indeed, we don't need the query param as we already have a content hash :+1:

PR #984, opened 2023-01-12
Buffer toDevice messages in CallHandler

MidhunSureshR:


bwindels:
Sorry to have wasted your time, but actually this is not needed. You would only receive signalling messages when you have sent a state event that you are a member of this call. Your client would only do this when you press the join button on the call, and for this it would already have a call object. So when receiving signalling, you should already have a call object.

PR #987, opened 2023-01-16
Calls: Switch camera lights off when muting

MidhunSureshR:


src/matrix/calls/group/GroupCall.ts
```diff
@@ -125,6 +127,7 @@ export class GroupCall extends EventEmitter<{change: never}> {
     }
 
     get localMedia(): LocalMedia | undefined { return this.joinedData?.localMedia; }
+    get localPreviewMedia(): LocalMedia | undefined { return this.joinedData?.localPreviewMedia; }
```
bwindels, line 130: I think we can remove the `localMedia` getter. It should only be used from previewing in the UI.

src/matrix/calls/group/GroupCall.ts
```diff
@@ -165,10 +168,12 @@ export class GroupCall extends EventEmitter<{change: never}> {
         const membersLogItem = logItem.child("member connections");
         const localMuteSettings = new MuteSettings();
         localMuteSettings.updateTrackInfo(localMedia.userMedia);
+        const localPreviewMedia = localMedia.asPreview();
```
bwindels, line 171: I think we might need to do this as well in `setMedia`?

bwindels (commented):
Thanks, looks like a better approach indeed. Just a few comments.

src/matrix/calls/group/GroupCall.ts
MidhunSureshR: It's used in CallViewModel.toggleCamera() and toggleMicrophone() to check if tracks are available

src/matrix/calls/group/GroupCall.ts
```diff
@@ -200,6 +200,8 @@ export class GroupCall extends EventEmitter<{change: never}> {
         if ((this._state === GroupCallState.Joining || this._state === GroupCallState.Joined) && this.joinedData) {
             const oldMedia = this.joinedData.localMedia;
             this.joinedData.localMedia = localMedia;
+            this.localPreviewMedia?.dispose();
```
bwindels: Slightly confusing to use the public getter here for accessing and the private member for assigning, but I guess that works.

PR #996, opened 2023-01-19
Build docker image for multiple architectures including ARM

hughns:
This is based on https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/ which doesn't use QEMU


spantaleev:
Looking forward to a tagged release, so we can disable building this on the server (for non-`amd64` servers) in https://github.com/spantaleev/matrix-docker-ansible-deploy

PR #1003, opened 2023-01-23
Improve CallTile UI

bwindels:
Mostly done but can be merged anyway so `bwindels/calls` contains the main improvements while I'm out tomorrow. Missing so far:
 - only include 5 or so member avatars?
 - emit updates from view model when member count changes
 - minor style changes like decide on alignment of tile (e.g. left instead of centered)

image
image

### DIFF
--- a/src/matrix/calls/CallHandler.ts
+++ b/src/matrix/calls/CallHandler.ts
@@ -18,7 +18,7 @@ import {ObservableMap} from "../../observable/map";
 import {WebRTC, PeerConnection} from "../../platform/types/WebRTC";
 import {MediaDevices, Track} from "../../platform/types/MediaDevices";
 import {handlesEventType} from "./PeerCall";
-import {EventType, CallIntent} from "./callEventTypes";
+import {EventType, CallIntent, CallType} from "./callEventTypes";
 import {GroupCall} from "./group/GroupCall";
 import {makeId} from "../common";
 import {CALL_LOG_TYPE} from "./common";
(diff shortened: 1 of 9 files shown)


src/matrix/calls/CallHandler.ts
```diff
@@ -130,12 +130,12 @@ export class CallHandler implements RoomStateHandler {
...
     }
 
-    createCall(roomId: string, type: "m.video" | "m.voice", name: string, intent?: CallIntent, log?: ILogItem): Promise<GroupCall> {
+    createCall(roomId: string, type: CallType, name: string, intent?: CallIntent, log?: ILogItem): Promise<GroupCall> {
         return this.options.logger.wrapOrRun(log, "CallHandler.createCall", async log => {
             if (!intent) {
                 intent = CallIntent.Ring;
             }
-            const call = new GroupCall(makeId("conf-"), true, { 
+            const call = new GroupCall(makeId("conf-"), true, undefined, {
```
MidhunSureshR: Why do we pass undefined as start time?

src/platform/web/dom/TimeFormatter.ts
```diff
@@ -75,6 +79,37 @@ export class TimeFormatter implements ITimeFormatter {
...
+
+    formatDuration(milliseconds: number): string {
+        let days = 0;
+        let hours = 0;
+        let minutes = 0;
+        if (milliseconds > DAYS_IN_MS) {
+            days = Math.floor(milliseconds / DAYS_IN_MS);
+            milliseconds -= days * DAYS_IN_MS;
+        }
+        if (milliseconds > HOURS_IN_MS) {
```
MidhunSureshR: Should the `>` be `>=` ?
Currently `formatDuration(25 hours in ms)` gives `1d 0h 60m 0s`

MidhunSureshR (approved):
Looks good, just have one question!

src/matrix/calls/CallHandler.ts
bwindels: Because the call hasn't started yet. It is started by the remote echo of the call event we're about to send.

PR #1010, opened 2023-01-31
Calls: Fixes for toast issues

MidhunSureshR:
- Fix an issue where you would get an error when you were invited to a room and the room had call events:

image

- Fix an issue where toast notifications were incorrectly shown for terminated calls


bwindels (approved):
lgtm, thanks

PR #1039, opened 2023-02-18
Update eslint rules

sleroq:
I have noticed some inconsistencies in the code, which is why I am proposing an update to the ESLint configuration.

In this merge request, I removed `.ts-eslintrc.js` and added new rules to main eslint config.

I picked values are based on what gives fewer warnings and what we had in the past, so this may not be the best for each situation; however, this should provide a good starting point as far as making sure coding practices are consistent across all files.

That being said, I don't want to force anything upon anyone else working on this project - if you feel like some of these rules need changing or removing, then let's discuss it, so we can come up with something everyone agrees upon

I would appreciate any feedback before merging my request into master branch if possible!


sleroq:
I also suggest you to consider adding [object-curly-spacing](https://eslint.org/docs/latest/rules/object-curly-spacing) rule to the ESLint configuration, 

I left it as it, because I'm not sure on the appropriate values for this rule.

.eslintrc.js
```diff
@@ -3,16 +3,31 @@ module.exports = {
...
         "no-empty": "off",
         "no-prototype-builtins": "off",
-        "no-unused-vars": "warn"
+        "no-unused-vars": "warn",
+        "quote-props": ["warn", "consistent"],
+        "no-multi-spaces": ["warn", { "ignoreEOLComments": true }],
+        "@typescript-eslint/no-unused-vars": "warn"
+        // "@typescript-eslint/no-misused-promises": "error",
+        // "@typescript-eslint/no-floating-promises": 2,
+        // "@typescript-eslint/explicit-function-return-type": "error"
```
sleroq, line 30: Rather than taking the to resolve the warnings, I opted to disable this rules for now.

KarriSauce:
no

.eslintrc.js
KarriSauce: time

PR #1040, opened 2023-02-20
Implement SAS Verification for crosssigning

MidhunSureshR:


src/matrix/verification/SAS/stages/WaitForIncomingMessageStage.ts
```diff
@@ -0,0 +1,73 @@
...
+            });
+            this.dispose();
+        });
+    }
+
+    private fetchMessageEventsFromTimeline() {
+        // todo: add timeout after 10 mins
+        return new Promise(resolve => {
+            this.track(
+                this.room._timeline.entries.subscribe({
```
bwindels: Fwiw, `room._timeline` will only exist if the room is open in the UI. Either we should call `room.openTimeline()` here or we need to add some sort of callback for incoming timeline events like we added for state events for calls.

src/matrix/DeviceMessageHandler.js
```diff
@@ -101,6 +105,19 @@ export class DeviceMessageHandler {
             }
         }
     }
+
+    _emitUnencryptedEvents(toDeviceEvents) {
+        const unencryptedEvents = toDeviceEvents.filter(e => e.type !== "m.room.encrypted");
+        for (const event of unencryptedEvents) {
+            this.emit("message", { unencrypted: event });
```
bwindels, line 112:
```suggestion
            this.emit("message", event);
```
why not just work with an optional third `DecryptionResult` argument for encrypted events? Does a bit less allocations and makes it easier to listen for messages if you don't care whether they were encrypted or not.

So below you'd have:
```
            this.emit("message", result.event, result);
```

src/matrix/DeviceMessageHandler.js
```diff
@@ -74,6 +77,7 @@ export class DeviceMessageHandler {
     }
 
     async afterSyncCompleted(decryptionResults, deviceTracker, hsApi, log) {
+        this._emitEncryptedEvents(decryptionResults);
         if (this._callHandler) {
```
bwindels, line 81: ideally we'd convert this code dealing with calls to use the event as well, but perhaps that's for a different PR...

src/matrix/DeviceMessageHandler.js
```diff
@@ -101,6 +105,19 @@ export class DeviceMessageHandler {
...
+
+    _emitUnencryptedEvents(toDeviceEvents) {
+        const unencryptedEvents = toDeviceEvents.filter(e => e.type !== "m.room.encrypted");
+        for (const event of unencryptedEvents) {
+            this.emit("message", { unencrypted: event });
+        }
+    }
+
+    _emitEncryptedEvents(decryptionResults) {
+        for (const result of decryptionResults) {
```
bwindels: so at this stage we haven't checked that the sender key used to send the event is signed by the ed25519 device identity key as published on the homeserver. That's what the code for calls does above. So we haven't verified the identity of the sender. Given that the SAS events are not required to be encrypted, perhaps we just don't emit for now for encrypted events so we don't shoot ourselves in the foot with this once we do use it later on?

src/matrix/verification/SAS/channel/types.ts
```diff
@@ -0,0 +1,25 @@
...
+    Ready = "m.key.verification.ready",
+    Start = "m.key.verification.start",
+    Accept = "m.key.verification.accept",
+    Key = "m.key.verification.key",
+    Cancel = "m.key.verification.cancel",
+    Mac = "m.key.verification.mac",
+    Done = "m.key.verification.done",
+}
+
+export const enum CancelTypes {
```
bwindels: Given the field where this is used is called `reason`, this might be a better name?
```suggestion
export const enum CancelReason {
```

src/matrix/verification/SAS/channel/MockChannel.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    setStartMessage(event: any): void {
+        this.startMessage = event;
+        this.initiatedByUs = event.content.from_device === this.ourUserDeviceId;
+        this.recalculateCommitment();
+    }
+
+    setOurDeviceId(id: string) {
+        this.ourUserDeviceId = id;
```
bwindels: why not pass this through the constructor?

src/matrix/verification/SAS/SASVerification.ts
```diff
@@ -0,0 +1,621 @@
...
+    clock: Clock;
+}
+
+export class SASVerification {
+    private startStage: BaseSASVerificationStage;
+    private olmSas: Olm.SAS;
+    public finished: boolean = false;
+    public readonly channel: IChannel;
+    private timeout: Timeout;
+    public readonly eventEmitter: EventEmitter<SASProgressEvents> = new EventEmitter();
```
bwindels: why not inherit from EventEmitter?

src/matrix/verification/SAS/stages/SelectVerificationMethodStage.ts
```diff
@@ -0,0 +1,97 @@
...
+        if (!this.allowSelection) { return; } 
+        const content = {
+            method: "m.sas.v1",
+            from_device: this.ourUserDeviceId,
+            key_agreement_protocols: KEY_AGREEMENT_LIST,
+            hashes: HASHES_LIST,
+            message_authentication_codes: MAC_LIST,
+            short_authentication_string: SAS_LIST,
+        };
+        await this.channel.send(VerificationEventTypes.Start, content, log);
```
bwindels: Can we add a comment here saying that this should trigger the other side to send an `accept` in reply, which will move us to the next stage?

src/matrix/verification/SAS/channel/Channel.ts
```diff
@@ -0,0 +1,294 @@
...
+export class ToDeviceChannel extends Disposables implements IChannel {
+    private readonly hsApi: HomeServerApi;
+    private readonly deviceTracker: DeviceTracker;
+    private ourDeviceId: string;
+    private readonly otherUserId: string;
+    private readonly clock: Clock;
+    private readonly deviceMessageHandler: DeviceMessageHandler;
+    private readonly sentMessages: Map<VerificationEventTypes, any> = new Map();
+    private readonly receivedMessages: Map<VerificationEventTypes, any> = new Map();
+    private readonly waitMap: Map<string, {resolve: any, reject: any, promise: Promise<any>}> = new Map();
```
bwindels: I've introduced a `Deferred` class in my PR that would fit very well here.

src/matrix/verification/SAS/channel/MockChannel.ts
```diff
@@ -0,0 +1,139 @@
...
+                macContent.mac[keyId] = calculateMac(device.ed25519Key, baseInfo + keyId);
+            }
+            else {
+                const {masterKey} = await this.deviceTracker.getCrossSigningKeysForUser(this.otherUserId);
+                macContent.mac[keyId] = calculateMac(masterKey, baseInfo + keyId);
+            }
+        }
+    }
+
+    setStartMessage(event: any): void {
```
bwindels, line 111: do you need to expose this to the Stage? Since all messages are received through the channel, it could easily do this internally? Or you think it's clearer this way?

src/matrix/verification/SAS/channel/types.ts
```diff
@@ -0,0 +1,25 @@
+export const enum VerificationEventTypes {
```
bwindels: If I'm not mistaken, we usually use singular for enums?
```suggestion
export const enum VerificationEventType {
```

src/matrix/verification/SAS/stages/CalculateSASStage.ts
```diff
@@ -0,0 +1,142 @@
...
+                log.log({l: "Commitment mismatched!", received: receivedCommitment, calculated: hash});
+                await this.channel.cancelVerification(CancelTypes.MismatchedCommitment);
+                return false;
+            }
+            return true;
+        });
+    }
+
+    private get needsToVerifyHashCommitment(): boolean {
+        if (this.channel.initiatedByUs) {
```
bwindels:
```suggestion
        return this.channel.initiatedByUs;
```
or even inline this?

src/matrix/verification/SAS/channel/Channel.ts
```diff
@@ -0,0 +1,294 @@
...
+    [CancelTypes.UnexpectedMessage]: "Unexpected Message.",
+    [CancelTypes.UnknownMethod]: "Unknown method.",
+    [CancelTypes.UnknownTransaction]: "Unknown Transaction.",
+    [CancelTypes.UserMismatch]: "User Mismatch",
+    [CancelTypes.MismatchedCommitment]: "Hash commitment does not match.",
+    [CancelTypes.MismatchedSAS]: "Emoji/decimal does not match.",
+}
+
+export interface IChannel {
+    send(eventType: string, content: any, log: ILogItem): Promise<void>;
```
bwindels:
```suggestion
    send(eventType: VerificationEventTypes, content: any, log: ILogItem): Promise<void>;
```

src/matrix/verification/SAS/channel/MockChannel.ts
```diff
@@ -0,0 +1,139 @@
...
+            return;
+        }
+        const {content} = this.startMessage;
+        const {content: keyMessage} = this.fixtures.get(VerificationEventTypes.Key);
+        const key = keyMessage.key;
+        const commitmentStr = key + anotherjson.stringify(content);
+        const olmUtil = new this.olm.Utility();
+        const commitment = olmUtil.sha256(commitmentStr);
+        olmUtil.free();
+        acceptMessage.commitment = commitment;
```
bwindels, line 79: I'm a bit confused why we modify the sent or received event here and set the commitment on it?

src/matrix/DeviceMessageHandler.js
MidhunSureshR: Created https://github.com/vector-im/hydrogen-web/issues/1061

src/matrix/verification/SAS/channel/MockChannel.ts
MidhunSureshR: Yes, because in the case where you send the start message but also receive a start message, there's a conflict. 
This is resolved in `SelectVerificationMethodStage` which tells the channel which the "true" start message is.

src/matrix/verification/SAS/channel/MockChannel.ts
MidhunSureshR: This is for the unit-tests, where we use hard coded messages. The commitment on them needs to change based on your public key which is generated at random by olm.

src/matrix/verification/SAS/channel/MockChannel.ts
bwindels: aha, that makes sense, thanks!

src/matrix/verification/SAS/channel/MockChannel.ts
bwindels: Oh, I missed that this is in the `MockChannel`, not in the real one!

bwindels (approved):
some typescript failures, but otherwise lgtm

src/matrix/verification/CrossSigning.ts
```diff
@@ -14,19 +14,23 @@ See the License for the specific language governing permissions and
...
 import {verifyEd25519Signature, SignatureVerification} from "../e2ee/common";
 
 import type {SecretStorage} from "../ssss/SecretStorage";
 import type {Storage} from "../storage/idb/Storage";
-import type {Transaction} from "../storage/idb/Transaction";
 import type {Platform} from "../../platform/web/Platform";
 import type {DeviceTracker} from "../e2ee/DeviceTracker";
 import type {HomeServerApi} from "../net/HomeServerApi";
 import type {Account} from "../e2ee/Account";
+import {ILogItem} from "../../lib";
```
bwindels: this wrong import crept back in again

src/matrix/verification/CrossSigning.ts
```diff
@@ -103,7 +112,27 @@ export class CrossSigning {
...
         this.ownUserId = options.ownUserId;
+        this.deviceId = options.deviceId;
         this.e2eeAccount = options.e2eeAccount
+        this.deviceMessageHandler = options.deviceMessageHandler;
+
+        this.deviceMessageHandler.on("message", async ({ unencrypted: unencryptedEvent }) => {
+            if (this.sasVerificationInProgress &&
+                (
+                    !this.sasVerificationInProgress.finished ||
+                    // If the start message is for the previous sasverification, ignore it.
```
bwindels, line 121: This comment doesn't make sense to me, why can we assume here that we're dealing with a start message?

bwindels: Ah because we're only processing start messages below anyway? Perhaps this can be a bit more clear.

bwindels (approved):
lgtm apart from one remaining comment

src/matrix/verification/CrossSigning.ts
MidhunSureshR: I refactored the code in the [other PR](https://github.com/vector-im/hydrogen-web/pull/1057/commits/e198e2039607a572b9bc6cbbddd64057e46d4d1e)

PR #1041, opened 2023-02-23
Proposal: Sync in worker

psrpinto:
Addresses https://github.com/vector-im/hydrogen-web/issues/1045

This is a proposal for running sync in a worker, so that it would be possible to have the same session open in multiple tabs/windows/iframes.

[Rendered](https://github.com/vector-im/hydrogen-web/blob/ea19cf278842cfc870af08ae9d756e0135753867/doc/implementation%20planning/sync-in-worker.md)

## TODO

- [x] Clarify the lifetime and ownership of workers
    - [x] Take in account that service worker must remain optional
    - [x] Explore possibility of using `SharedWorker`
- [x] Clarify whether mutations to the `Session` in the main thread need to be communicated to the worker
- [x] Clarify how messages will be passed between main thread and worker, check browser support for `BroadcastChannel`.
- [x] Clarify whether there's a limit on the number of workers that can run at a given time
- [ ] Clarify the usage of `localStorage` in `Storage`, since `localStorage` is not available in workers
- [ ] Clarify "after sync" step
    - [ ] Some things should only be run once per sync
    - [ ] List of things that would need their updates passed over the message channel between the worker and the page

### DIFF
--- a/doc/implementation planning/sync-in-worker.md
+++ b/doc/implementation planning/sync-in-worker.md
@@ -1,0 +1,119 @@
+# Sync in worker
+Currently, it's not possible to have the same session open in multiple Hydrogen instances (i.e. in multiple tabs, windows or iframes) since this would result in having multiple ongoing syncs for the same session. Having multiple ongoing syncs could corrupt data in `indexedDB` and would also not make sense from a resource utilisation perspective.
+
+To prevent that multiple syncs are ongoing for the same session, a session-closing mechanism is implemented in the service worker: when a session is opened in an instance, the service worker tells all other instances to close that session, if they have it open.
+
+From the user's perspective, this results in Hydrogen going back to the session picker screen when the session that was open in one tab is opened in another tab.
+
+In this document, we propose a solution to address this limitation, making it possible to have the same session simultaneously open in multiple Hydrogen instances, while maintaining a fully-featured experience for the user.
+
+This proposal would also pave the way for having simultaneous ongoing syncs for multiple sessions (i.e. multiple accounts) in the future, though that's not a use case we would focus on at this point.
+
+## Offload sync to worker
+
+The general idea would be to have sync running in a [Web Worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) instead of the main thread, while making it **transparent to UI code** (i.e. code running in the main thread), which should not be aware of whether sync is running in a worker or in the main thread. Making it transparent to UI code is important since we need to fallback to non-worker sync in environments where Web Workers are not available (e.g. IE11).
+
+Since we don't want to tie the lifetime of the sync worker to a specific tab/window (we want the sync to happen for as long as there's at least one instance with that session open), the sync worker would be *owned* by the service worker. Note that this doesn't mean that the service worker would run the sync, but that it would spawn another worker dedicated to the sync.
+
+The UI thread would communicate with the sync worker through messages, possibly leveraging [`BroadcastChannel`](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel).
+
+We would be making the changes behind a feature flag, which would allows us to spread the implementation across multiple PRs, and iterate on the feature until we're confident that it's stable.
+
+## `SyncProxy` (UI side)
+
+`SyncProxy` would be the equivalent of `Sync` but would be a thin layer that *proxies* method calls to the worker, something like the following:
+
+```typescript
+// SyncProxy.ts
+
+// Here we're extending Sync but it would make more sense to extract an ISync
+// interface that both Sync and SyncProxy implement.
+export class SyncProxy extends Sync {
+    // Not a full implementation, just showcasing a method call.
+    // In the final implementation all public methods of Sync would need to
+    // be proxied.
+    async start(args: object): Promise<object> {
+        const result = await sendAndWaitForReply("start", args);
+        if (result.error) {
+            throw result.error;
+        }
+        return result.data;
+    }
+}
+```
+
+In environments where Web Workers are available, and if the feature flag is enabled, we would *swap* `Sync` with `SyncProxy`:
+
+```javascript
+// Client.js
+
+if (window.Worker && syncFeatureFlagEnabled) {
+    this._sync = new SyncProxy({...});
+} else {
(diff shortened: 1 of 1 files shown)


bwindels:
Thanks for this proposal, some early thoughts:
 - being able to build workers with vite would be useful indeed, you might find some notes in https://github.com/vector-im/hydrogen-web/issues/292 thay may or may not be outdated.
 - given the scope of these changes, it would be absolutely critical that we can do this behind a feature flag (apart from any build-system changes perhaps).
 - putting sync in a worker won't help us having to elect a master tab, as a worker's lifetime is tied to the page that started it. A service worker doesn't have this limitation though, and putting sync in the service worker could make sense (and have some benefits like being able to sync when a web push notification comes in) but I'd be a bit hesitant to make Hydrogen not being able to run at all without a service worker. We probably want to be flexible here and allow for different strategies.
 - Something worth mentioning: the `afterSyncCompleted` (on room and session) step should only be run once per sync request. In this sync step, we'll send out network requests that need to happen as a side-effect of the sync, like sharing keys, backfilling, ... so this should only happen once.
 - Some things probably should not be mirrored on both pages, like when you start a call.

bwindels:
I just saw you mentioned the worker being owned by the service worker. That has the same limitation that hydrogen wouldn't be able to run without a worker though.

A [SharedWorker](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker) seems like what we want, but support is lacking on mobile platforms. Perhaps that's good enough though, use a service worker in production, and a SharedWorker during development time. We should of course still be able to run sync within the page and disable cross-tab syncing.

bwindels:
A good first step would be to list things of things that would need their updates passed over the message channel between the worker and the page:
 - room summary updates
 - timeline updates
 - need to think if there's anything else

psrpinto:
Thanks for looking into this proposal @bwindels!

Agree, we would be making the changes behind a feature flag, which would allows us to spread the implementation across multiple PRs, and iterate on the feature until we're confident that it's stable.

When the feature flag is disabled, or the environment doesn't allow for running sync in a worker (e.g. no support for Web Workers API), we would fallback to the non-worker sync, exactly as is today. If necessary, we could also make it possible to disable sync-in-worker through runtime config, or even at the build level.

I agree Hydrogen must continue to be functional without a service worker. I will look further into this topic, and into the "after sync" step and will update the document accordingly.

I've updated the PR description with a TODO section.

psrpinto:
> list things of things that would need their updates passed over the message channel between the worker and the page

@bwindels Is the opposite scenario also needed, i.e. passing updates to the worker when the `Session` in the page is modified? Looking at the [architecture docs](https://github.com/vector-im/hydrogen-web/blob/master/doc/architecture/architecture.md) it's mentioned that:

> `Sync` is the thing (**although not the only thing)** that mutates the `Session`

So if mutations to the `Session` happen in the main page, would the worker also need those mutations for `Sync` to work correctly? I will look further into this but figured you might know something from the top of your head.

bwindels:
> > list things of things that would need their updates passed over the message channel between the worker and the page
> 
> @bwindels Is the opposite scenario also needed, i.e. passing updates to the worker when the `Session` in the page is modified? Looking at the [architecture docs](https://github.com/vector-im/hydrogen-web/blob/master/doc/architecture/architecture.md) it's mentioned that:
> 
> > `Sync` is the thing (**although not the only thing)** that mutates the `Session`
> 
> So if mutations to the `Session` happen in the main page, would the worker also need those mutations for `Sync` to work correctly? I will look further into this but figured you might know something from the top of your head.

hmm, you could probably get away with relying on remote echo, e.g. the /sync api reflecting a change you made like sending a message, creating a room, ... there will be a delay for the server roundtrip but that should be ok?

psrpinto:
Relying on sync to propagate the changes makes sense, server roundtrip should be fine indeed.

psrpinto:
~I've added a couple of sections to the doc ([Worker lifetime and Gracefully closing sessions](https://github.com/vector-im/hydrogen-web/blob/6db66e8e50261ffc7832974aa3f517485d68d8f6/doc/implementation%20planning/sync-in-worker.md#worker-lifetime)) that address the discussions above about service worker vs shared worker, and other topics.~

~In short, we propose to use a "double worker" solution, with one `SharedWorker`, which then spawns a sync `Worker` per `sessionId` (please see the doc for the reasoning behind this). Looking forward to hearing your thoughts on this @bwindels.~

I've also created https://github.com/vector-im/hydrogen-web/issues/1045 to track the feature, which would also serve as the _epic_ for this endeavour. I've started extracting actionables from this proposal into a TODO section on that issue. Let me know if there's anything in that TODO that seems off, or if you have any other feedback.

In the meantime, I'm looking into the other topics in the `TODO` of this PR.

psrpinto:
~I just realised that there are some things that aren't really clear in the [Worker lifetime and Gracefully closing sessions](https://github.com/vector-im/hydrogen-web/blob/6db66e8e50261ffc7832974aa3f517485d68d8f6/doc/implementation%20planning/sync-in-worker.md#worker-lifetime) so I'm giving it another pass. Will update here once done.~

psrpinto:
@bwindels I've clarified the service worker vs shared worker topic in [the doc](https://github.com/vector-im/hydrogen-web/blob/ea19cf278842cfc870af08ae9d756e0135753867/doc/implementation%20planning/sync-in-worker.md#offload-sync-to-worker).

In short, there's one `SharedWorker` per session:

```typescript
const sessionId = "foo";
const worker = new SharedWorker(new URL('./sync-worker.js'), {
    name: `sync-worker-${sessionId}`,
})
```

This code can run in multiple windows/tabs/iframes. The browser spawns a worker the first time one is created with the given `name`. Subsequent attempts to create a worker with the same `name` result in the existing one being returned. The browser will despawn the worker once there are no more references to it.

bwindels:
Just saw that workers of type es module are coming to firefox! This means we may not need a build-step to convert `import` to `importScripts` for the SDK, depending on which browsers need to support the tab sharing feature! it's still behind a flag, but one would hope they enable it soon enough by default. See https://caniuse.com/mdn-api_worker_worker_ecmascript_modules and https://bugzilla.mozilla.org/show_bug.cgi?id=1247687

Issue #954, opened 2022-12-15
Cross signing: implement signing our own devices


bwindels:
Depends on #953 

Once we can trust our MSK with #953, we can now also trust the USK (User Signing Key) and SSK (Self Signing Key). The private keys for these are again stored in 4S, but verification should rather rely on the signatures from the MSK. We should only require to have the private keys when we actually want to sign something.

 - derive USK and SSK trust from parent key signature (MSK). This will not contain the hierarchical caching system for signature validation we eventually want, see #957 for that.
 - sign own device with SSK
 - publish signature
 - sign other users with USK and publish signature

bwindels:
One note on the above, when we have the private MSK, we'll also have the private SSK and USK as they are all unlocked with the 4S key. So verification ... probably still good to check the signatures are ok, but we can rely on the private as well...

bwindels:
In the settings UI, once everything is validated, show a button to sign our own hydrogen device if not signed already. If signed already, show a status message saying so.

bwindels:
This is just adding support in code without exposing it in the UI, this happens in the next issue #955

bwindels:
## Steps for own device signing

 - fetch SSK + signature
 - verify SSK signatures from MSK
 - get device key for our own device (already supported), the ed25519 one?
 - calculate signature with SSK
 - upload signature

bwindels:
For signing other users and other devices apart from the current one:
 - rework way device keys are stored
   - so we can:
     - also store cross-signing keys in same store
     - store signatures
     - store in format as returned by server as that is what we have to calculate signatures on
    - for this, we'll:
       - rename deviceIdentities store to userKeys store
       - rename userIdentities store to userKeyTracking store
       - store device/msk keys in device format as returned by server
       - keep signatures in there, so we can later verify them if needed

bwindels:
Ah but we have an index on the curve25519 sender key for device keys which we don't have for cross-signing keys 🤔

bwindels:
For signing other users and other devices apart from the current one:
 -  add crossSigningKeys store, key by userId|key, we also need to look up by userId|usage (which can contain multiple usages), have an multi index on here? How about IE support?
 - store signatures in deviceIdentites
 - remove crossSigningKeys from userIdentities
 - rename userIdentities to userKeyTracking?

bwindels:
From uhoreg:

[...]

PR #1042, opened 2023-02-24
Cross-signing: sign other users

bwindels:
Implements remaining part of #954 and also signing other users.

Cross-signing keys are now kept in a separate store, 1 key per entry. Device keys are now kept in the format as returned by the server as this makes it easier to sign the json structures. Given that they are now stored in the same format as cross-signing keys, the server format for keys, I renamed deviceIdentity to deviceKey as well everywhere.

This PR includes a storage migration, bringing the storage to version 18.

As you really shouldn't cross-sign anyone without having verified them first (happening in #1040), this puts a massive warning when trying to use this (as well as being behind the feature flag):

[Screencast from 03-03-23 12:04:29.webm](https://user-images.githubusercontent.com/274386/222704777-8e59e17c-04ae-4f00-a6a9-a44ad3a68ad7.webm)

### DIFF
--- a/src/matrix/verification/CrossSigning.ts
+++ b/src/matrix/verification/CrossSigning.ts
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { ILogItem } from "../../lib";
+import {ILogItem} from "../../logging/types";
 import {pkSign} from "./common";
 
 import type {SecretStorage} from "../ssss/SecretStorage";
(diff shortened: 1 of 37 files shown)


src/matrix/verification/CrossSigning.ts
```diff
@@ -14,19 +14,34 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import { ILogItem } from "../../lib";
```
MidhunSureshR: This import is from the wrong file; also can you remove the spaces between braces?

MidhunSureshR (commented):
LGTM :+1:

Issue #957, opened 2022-12-15
Cross signing: implement the shields


bwindels:
Depends on #956 

implementing the shields (doing the signature verification but not signing yet), the signature is included in the section for the signed key in the /keys/query response for that user. We can't do this as the first point because we can't yet trust our own MSK, as we don't have a means of verifying either another client or verifying that the private key in 4S checks out with the published MSK.

we need the keys of all members in a room to be able to show the room shield. we probably at least want to show the shield in the room header when you open the room. Element does not show the shield in the left panel, probably precisely for this reason. Showing the shield implies tracking the room (downloading members, downloading their keys and tracking and member device changes) so does that render the optimizations to not track the room until we send our first message obsolete?

This will also include the signature verification caching framework that goes from the bottom up to the root of the trust tree as laid out in [a comment on the Epic](https://github.com/vector-im/hydrogen-web/issues/827#issuecomment-1277431499)

One crucial implementation note that Valere brought up is that a trusted flag should be a tri-state, not just a boolean: trusted, not trusted and dirty, e.g. need to do work to be able to determine,

 - implement signature validation of signature with parent key
 - implement signature validation of each level:
    - signature of other user device with their SSK
    - signature of other user MSK of their SSK
    - signature of other user MSK with our USK
    - signature of our device with our SSK
    - signature of our SSK with our MSK
 - aggregate trust for devices of user to trust of the user (e.g. any device not trusted, don't trust user)
 - aggregate trust for all users to trust for room (e.g. any user not trusted, don't trust room)
 - cache aggregation at user and room level

bwindels:
Notes from the crypto team: https://hackmd.io/@ktO1wyrVSni2M6FoFrnW7Q/SyJWBMc_o

bwindels:
First PR:
 - no caching yet
 - verify signature of MSK of user with our USK
 - verify signature of all of their device keys with their SSK
 - show shield in member panel

bwindels:
Things that need to get a shield:
 - room events (verify that sender key is verified all the way up to our MSK)
 - room members (verify all their device keys are verified all the way up to our MSK)
 - rooms (check all members are verified)

MidhunSureshR:
Closed in favor of #1052, #1053, #1054, #1055

PR #1047, opened 2023-03-03
Cross-signing: user shield

bwindels:
Without caching for now. Part of #957, based on #1042

### DIFF
--- a/src/domain/session/rightpanel/MemberDetailsViewModel.js
+++ b/src/domain/session/rightpanel/MemberDetailsViewModel.js
@@ -17,6 +17,7 @@ limitations under the License.
 import {ViewModel} from "../../ViewModel";
 import {RoomType} from "../../../matrix/room/common";
 import {avatarInitials, getIdentifierColorNumber, getAvatarHttpUrl} from "../../avatar";
+import {UserTrust} from "../../../matrix/verification/CrossSigning";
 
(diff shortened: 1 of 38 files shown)


bwindels:
Remaining:
 - [x] consider doing the verification in reverse order (e.g. from our MSK to their device keys) as that will incur less signature verifications for users we haven't cross-signed. (assuming the most common case for non-green shields is users we haven't verified yet but that have verified all their devices)
 - [ ] The UI is shown after session.load but session.crossSigning is only available during session.start (run after each reconnect). We should move the creation of everything depending on secret storage (key backup, cross-signing) to session.load but defer any network requests that these need to make to session.start. Then, session.crossSigning will already exist by the time the UI is shown. (as this PR stands, if you have the member detail panel open and refresh, you'll get an exception that session.crossSigning is undefined.
 - [ ] We need to implement a better UI with an actual shield icon and a descriptive tooltip of the trust status
 - [ ] We need to implement a grey shield status (need to find docs for shield rules)
 - [ ] We should probably return an ObservableValue for the user trust so updates can be sent, e.g. when we cross-sign the user. Or should we put this on a member object that can emit changes and make it more a core part of the model? Ideally, we'd make cross-signing optional though...

bwindels:
Replaced by https://github.com/vector-im/hydrogen-web/pull/1062

PR #1058, opened 2023-03-22
Refactor calls toast code to support other types of toasts 

MidhunSureshR:
ToastCollectionViewModel now exposes  a ConcatList of all toast viewmodels.


bwindels (approved):
LGTM

PR #1068, opened 2023-03-30
Cross-signing: implement observing user trust so UI can update when signing

bwindels:


src/matrix/verification/CrossSigning.ts
```diff
@@ -382,6 +410,11 @@ export class CrossSigning {
         };
         const request = this.hsApi.uploadSignatures(payload, {log});
         await request.response();
+        // we don't write the signatures to storage, as we don't want to have too many special
+        // cases in the trust algorithm, so instead we just clear the cross signing keys
```
MidhunSureshR: > instead we just clear the cross signing keys

Is it the device keys here?

MidhunSureshR (approved):

PR #1085, opened 2023-05-05
Fix broken sdk when used without encryption

### LINKED ISSUE #1081: SDK is broken when used without encryption
`afterSyncCompleted` uses `e2eeAccount` without checking if it exists.

MidhunSureshR:


MidhunSureshR:
Fixes #1081

PR #1150, opened 2023-11-07
Add option to allow disabling read receipts

Half-Shot:
To allow upstream apps the ability to stop sending them.


weeman1337 (approved):
Looks good :+1:

MidhunSureshR:
I'm a bit concerned about having code in Hydrogen that solves a very specific use case of a downstream project. In this case, ideally we'd have a `ReadOnlyRoomViewModel` that disables sending messages, read receipts and calls. I'm going to merge this now but I've also created https://github.com/vector-im/hydrogen-web/issues/1151

Half-Shot:
Yeah, I agree. I think in this case the problem was that the read receipt code was somewhat burred in the Room component. Perhaps we can fix that next time.

PR #1153, opened 2023-11-09
Add ability to adjust the token on the homeserver API.

Half-Shot:
This is needed in the case of expiring access tokens. I haven't implemented the full token refresh system yet, but I need this as a stepping stone for an upstream project.

### DIFF
--- a/src/matrix/Client.js
+++ b/src/matrix/Client.js
@@ -343,7 +343,7 @@ export class Client {
      * @param {string} token A Matrix Access Token 
      */
     async updateAccessToken(token) {
-        if (!_this.session) {
+        if (!this._session) {
             throw Error("No session loaded, cannot update access token");
         }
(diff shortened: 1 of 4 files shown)


src/matrix/Client.js
```diff
@@ -337,6 +337,16 @@ export class Client {
         }
     }
 
+    /**
+     * Update the access token in use by the client.
+     * Will also update the token in session storage.
+     * @param {string} token A Matrix Access Token 
+     */
+    async updateAccessToken(token) {
+        this._session.updateAccessToken(token);
```
justinbot, line 349: I think `_session` could be `null`?

Half-Shot: I think the check is sufficient

MidhunSureshR (approved):

PR #1161, opened 2024-01-28
uploadRoomKeysToBackup: Use V3 route

### LINKED ISSUE #741: `PUT /_matrix/client/r0/room_keys/keys` is used, while it was never specced.
Related: https://github.com/ruma/ruma/issues/1125

Came up in https://gitlab.com/famedly/conduit/-/issues/262

Some context, the `room_keys` had a bumpy road towards specification, and only landed in v1.1 when `v3` versions were mandatory.

However, during that time, only a *subset* of the `room_keys` APIs were specified under `r0`, *specifically* `PUT .../r0/room_keys/keys` was never specced pre-v1.1.

Synapse has a more lenient catch-all for r0 paths to v3 paths, but conduit follows ruma, which follows the spec rigorously, where there currently exist no released version that specifies `r0` for this endpoint.

See the relevant ruma code;

https://github.com/ruma/ruma/blob/74423b7a4e12254ca4fc40b3a35bc1c83bb216dc/crates/ruma-client-api/src/backup/add_backup_keys.rs#L18-L24

And relevant spec area;

https://spec.matrix.org/v1.2/client-server-api/#put_matrixclientv3room_keyskeys

NicolaiSoeborg:
Fixes https://github.com/element-hq/hydrogen-web/issues/741

"Tested" locally by rewriting `r0` to `v3` in reverse proxy (Caddy) and seeing `All keys are backed up.` instead of repeated failing PUT requests towards Conduit (home server)

```
matrix.example.com {
    # this will make Hydrogen and Conduit happy together,
    # but it shouldn't be the responsibility of the reverse proxy:
    rewrite /_matrix/client/r0/room_keys/keys /_matrix/client/v3/room_keys/keys?{query}

    reverse_proxy /_matrix/* localhost:8448
}
```

### DIFF
--- a/src/matrix/net/HomeServerApi.ts
+++ b/src/matrix/net/HomeServerApi.ts
@@ -251,7 +251,8 @@ export class HomeServerApi {
         return this._get(`/room_keys/keys/${encodeURIComponent(roomId)}/${encodeURIComponent(sessionId)}`, {version}, undefined, options);
     }
 
-    uploadRoomKeysToBackup(version: string, payload: Record<string, any>, options?: BaseRequestOptions): IHomeServerRequest {
+    uploadRoomKeysToBackup(version: string, payload: Record<string, any>, options: BaseRequestOptions = {}): IHomeServerRequest {
+        options.prefix = CS_V3_PREFIX;
         return this._put(`/room_keys/keys`, {version}, payload, options);
     }
 


NicolaiSoeborg:
@MidhunSureshR : I see you are active on this project, can I get you to review this PR? :)

MidhunSureshR (approved):

MidhunSureshR:
Reverting since there isn't a sign-off.

fwiw we should really deprecate the `r0` endpoint everywhere.

NicolaiSoeborg:
Can you re-commit this if I add the signoff message here? Otherwise I will make a new PR soon :)

Signed-off-by: Nicolai Søborg <git@xn--sb-lka.org>

PR #1174, opened 2024-08-16
Support service worker in all environments

MidhunSureshR:
For https://github.com/element-hq/hydrogen-web/issues/1173


MidhunSureshR:
There's a lot of formatting changes here but I can't be bothered to fix it since we don't have a prettier config.
I'll fix the formatting across the project in another PR.

PR #1201, opened 2024-12-02
Add support for `webpush` pusher as defined in MSC4174

MatMaul:
https://github.com/matrix-org/matrix-spec-proposals/pull/4174

### DIFF
--- a/src/matrix/Session.js
+++ b/src/matrix/Session.js
@@ -946,7 +946,7 @@ export class Session {
     async _enablePush() {
         return this._platform.logger.run("enablePush", async log => {
             const defaultPayload = Pusher.createDefaultPayload(this._sessionInfo.id);
-            const pusher = await this._platform.notificationService.enablePush(Pusher, defaultPayload);
+            const pusher = await this._platform.notificationService.enablePush(Pusher, defaultPayload, this._hsApi);
             if (!pusher) {
                 log.set("no_pusher", true);
                 return false;
--- a/src/platform/web/dom/NotificationService.js
+++ b/src/platform/web/dom/NotificationService.js
@@ -37,8 +37,7 @@ export class NotificationService {
                 events_only: true,
                 default_payload: defaultPayload
             };
-            return pusherFactory.httpPusher(
-                this._pushConfig.gatewayUrl,
+            return pusherFactory.webpushPusher(
                 this._pushConfig.appId,
                 pushkey,
                 data
--- a/src/platform/web/dom/NotificationService.js
+++ b/src/platform/web/dom/NotificationService.js
@@ -20,12 +20,24 @@ export class NotificationService {
         this._pushConfig = pushConfig;
     }
 
-    async enablePush(pusherFactory, defaultPayload) {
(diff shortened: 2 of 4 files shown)


src/platform/web/dom/NotificationService.js
anoadragon453, line 45: This has been replaced with a `url` field in the latest version of the MSC.

src/platform/web/dom/NotificationService.js
```diff
@@ -20,12 +20,24 @@ export class NotificationService {
         this._pushConfig = pushConfig;
     }
 
-    async enablePush(pusherFactory, defaultPayload) {
+    async enablePush(pusherFactory, defaultPayload, hsApi) {
```
anoadragon453, line 23: Does this code handle the `201` response code that the MSC asks for?

> If the request creates a new pusher or modifies values under `pushkey` , `PusherData.url`, or `PusherData.auth`, then the server MUST respond with 201, ...

as well as handling the subsequent validation push containing `app_id` and `ack_token`?

anoadragon453: We seem to also be missing the bit where the client calls `POST /_matrix/client/v3/pushers/ack` after receiving the validation push.

The client needs to handle the new `M_EXPIRED_ACTIVATION_TOKEN` and `M_UNKNOWN_ACTIVATION_TOKEN` error types.

src/matrix/Session.js
```diff
@@ -946,7 +946,7 @@ export class Session {
     async _enablePush() {
```
anoadragon453, line 946: It seems that nowhere are we checking for `PusherData.activated` before attempting to use this pusher.

anoadragon453 (changes requested):
Needs a few updates with respect to the latest draft of MSC4174 (`f59cb97551216812d1ca492fafba05d057cc1298`)

src/matrix/Session.js
p1gp1g: The client may use that property only for debugging/troubleshooting purpose, so that's normal to not check it here.

But we don't have the token activation yet (todo here: https://github.com/element-hq/hydrogen-web/blob/master/src/platform/web/sw.js#L336)

src/platform/web/dom/NotificationService.js
p1gp1g: ~~The new endpoint has to go there: https://github.com/element-hq/hydrogen-web/pull/1201/changes#r2923213109~~

This `endpoint` field is used for the current *http* pusher. That's OK to keep it with the new pusher, it will simply be ignored. To make things a little more evident, we can do `supportDirectWebPush ?  null : subscriptionData.endpoint`

src/platform/web/dom/NotificationService.js
p1gp1g: To be done on the SW 👍

anoadragon453: SW?

p1gp1g: service worker

PR #1205, opened 2025-01-11
fix: Spaces can be sent as message

### LINKED ISSUE #1203: Spaces can be sent as message
**Describe the bug**
Messages containing only whitespace are sendable.

**To Reproduce**
1. Open the chat interface
2. Type only spaces in the message input field
3. Send the message
4. Message with only whitespace gets sent successfully

**Expected behavior**
Messages containing only whitespace should not be sendable.

**Screenshots**
Screenshot 2025-01-12 020416

**Desktop**

 - Browser : Chrome

**Additional context**

KhushamBansal:
This change prevents sending messages that are either empty or contain only whitespace. When a user tries to send such a message, the method will simply return without sending anything. 
Closes #1203


KhushamBansal:
Hi maintainers. can you review this PR whenever you get time?

PR #1217, opened 2025-11-06
Use dedicated subdomain for bug reporting

benbz:
There is a dedicated subdomain for Rageshakes, it should be used to avoid issues with element.io path routing


Half-Shot (approved):
Can I approve this?

benbz:
It looks approved. CI is broken but that's not related to this PR, so I'm going to merge

